PR #2, opened 2025-06-29
Refactors

MeirGavish:
I made a few refactors in separate commits. One was to remove some magic numbers and replace them with meaningful defines, the other was to subdivide functions into smaller functions. Both should make the code more readable, maintainable and flexible.
There are still long functions that can be further subdivided, and I bet some more numbers can be replaced with meaningful names but these are the things that first came to mind when going over the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /.vscode
 /build
+/.vs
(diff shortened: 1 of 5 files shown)


cellos51:
lgtm

PR #7, opened 2025-06-30
Added pixel to small background text height to make it more readable

MeirGavish:
I added a pixel to the height of some small background text so it looks better and is more readable. 
I also added a pixel to the spaces containing the real-time printed number values so they are centered.
Also note that the "Score Aim" change adds a pixel to the overall height of the column, so the "Small Blind" text area is now located one pixel higher in case that matters.


Cryhme (approved):
Good.

cellos51:
Yeah this looks better, but the E could definitely be improved with something more like this: 
image

TheBreadPerson (approved):

TheBreadPerson (changes requested):
The original text was "Score Min" not "Score Aim". Unless this was an intentional change, that should be changed

TheBreadPerson (changes requested):
I think you misread the text next to score, it was originally "Min", not "Aim"

MeirGavish:
I guess I did misread it. I suppose this supports my point of making it more readable. I fixed it, as well as the E, and I also made the C ins "Discard" match the one in "Score".

HOWEVER NOTE I now see that after merging the latest changes in the main branch to this one there is a strange bug when displaying the end round screen.
image

I am not sure why this happens, I edited the images with LibreSprite, I can try a different software if you have a recommendation for a free one.
I pushed for now in case you want to take a look but I think for now this branch shouldn't be merged before this is fixed.

cellos51:
It's just a quirk of how the GBA handles graphics. I can fix it later.

PR #8, opened 2025-06-30
Changed the default game state to GAME_PLAYING so the game can be played

MeirGavish:
It seems the default game state was changed to GAME_ROUND_END, I don't know if this was intentional but since the game can't be played and tested, I changed it back for now.


cellos51:
Already merged something that changed it back.

PR #9, opened 2025-06-30
GAME_LOSE state

expected-ingot:
When at 0 hands, the game changes into a `GAME_LOSE` state. There is no code to actually handle it, it simply exists to help with adding a lose screen in the future

This also sets the initial game state to GAME_PLAYING, instead of GAME_ROUND_END


cellos51:
Looks G to M. I accidentally pushed with the game state on round end.

PR #11, opened 2025-07-03
Replace vid_sync() with VBL interrupt

MeirGavish:
According to [this issue](https://github.com/gbadev-org/tonc/issues/112) in Tonc, vid_vsync() shouldn't be used and VBlankIntrWait() should be used instead. Tonc's examples were also updated in [this commit](https://github.com/gbadev-org/libtonc-examples/commit/db70fa29a0baae12c5c7603426d8535ebb5cc6ed).

### DIFF
--- a/source/main.c
+++ b/source/main.c
@@ -141,7 +141,7 @@ int main()
 
 	while(true)
     {
-        vid_vsync();
+        VBlankIntrWait();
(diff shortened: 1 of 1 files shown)


cellos51:
damn i thought i pushed a change that added this already. guess not.

PR #14, opened 2025-07-06
Card graphic change

MeirGavish:
This is a proposed graphics change for the cards to a more simplistic design to fit the low resolution, the current design feels a little cramped at 32x24 sprites.
For the face cards I used the... faces from [these assets](https://drawsgood.itch.io/8bit-deck-card-assets) ([which were actually used in an early build of Balatro](https://x.com/LocalThunk/status/1795698573662982480)), and changed the color palette to be the same as the current design and the original Balatro.


cellos51:
So I'm biased because I made the original sprites but I do see how this improves the clarity. I'm going to let some people look at this first and have the popular opinion be what decides if this gets merged.

cellos51:
Here's a preview for both options. 
image
image

TheBreadPerson (commented):
I personally feel like although the new cards are simpler to read and provide more clarity, they take some of the personality out of the game. Going off of the screenshot sent by cellos51 in the discussion, the original cards fit better and more closley resemble the actual game.

KevDoy:
I'm a big fan of the original ones visually, but also agree that the more readable ones are the way to go. Maybe there's a middle ground between the two styles that could be explored?

MeirGavish:
I believe in the vision of making this port similar to the original game, maintaining the animations etc. I think the limitations of the hardware make it a little problematic in this instance. 
Here is another comparison screenshot that better illustrates the problems this redesign is trying to fix.

example_screenshot
example_screenshot_original

In the 9 and 10 of Clubs, there's not enough room on the cards and the club icons end up touching which doesn't look very good. Moreover the downscaled face cards are in great need of a makeover.
I'd say this redesign gives the game a different personality which is more simplified and may be more fitting for the GBA but I agree it does lose a bit of the "real cards" feel and if you feel this strays too much from the vision, we can try exploring other solutions to these problems.

GreenNinja2525:
One thing I noted was that the numbers are not flipped on the bottom of the new design like the old style does. I think the old style is more faithful and like someone mentioned earlier, a middle ground between both styles could be worth looking into. Perhaps just making the symbols smaller to fit them on the card could work or maybe something else.

MeirGavish:
Yes, I opted to forego flipping the bottom suit icons because if you look at the screenshot, when they are stacked in the hand you would only see the bottom left suit icon and if it were flipped it would be too unreadable.

MeirGavish:
I will say that while finding another solution may be possible, it will not be simple. The suit icons are 5x5 pixels which is already quite small, I don't know how well and readable smaller icons would look. The face cards are a bigger issue. The original Balatro sprites are 95x71 which is about 3 times bigger than these sprites, downscaling pixel art by a factor of 3 loses a lot. I'm not a pixel artist, all I can do is use a downscaling algorithm and try to do some fixes by hand but maybe a good pixel artist could make it look good.
I did try using pixel art in the assets whose faces I used to maybe maintain full-body rotationally symmetric face cards. They are 45x33 which is still bigger than our card sprites but much closer and they are already designed for a lower resolution than the original Balatro ones. After algorithmic downscaling and some fixes by hand I ended up with something like this which looks pretty good but unfortunately is still too big by a few pixels, there's no room for suit icons and letters.
Queen of Hearts downscaled 2 enlarged

This approach would take quite a bit of work to get to something that looks good because further downscaling is required and this needs to be done for each face card.

cellos51:
We could just make the card style a toggle able option in-game.

pumpedrowdy:
I think!

GreenNinja2525:
Edit: I forgot to properly utilize rotational symmetry, ignore this one, I am gonna rework it a little.
deck_gfx idea

MeirGavish:
I don't know if it's the best course of action to leave the decision to the user, it's also more implementation for something other than gameplay when we are still missing many gameplay parts. I suppose I can just as well keep the designs in my fork. But if I continue the approach from my previous comment and focus on the face cards by downscaling the assets I used and fixing them by hand to try to make something that both fits the size, looks good, and is still full body with rotational symmetry, would that be more acceptable?

Here is the same image with the Balatro Heart palette.
Queen of Hearts downscaled 2 enlarged

MeirGavish:
@GreenNinja2525 , I think this could be a good approach but I would try to maintain rotational symmetry and resemblance to Balatro and real cards, especially for the 10 cards, so for the 10 of clubs for example maybe something like this
10_of_clubs_edited_enlarged

GreenNinja2525:
> @GreenNinja2525 , I think this could be a good approach but I would try to maintain rotational symmetry and resemblance to Balatro and real cards, especially for the 10 cards, so for the 10 of clubs for example maybe something like this 10_of_clubs_edited_enlarged

That makes sense, not only does that look better, it is a much better approach. If we apply that to the others, it might actually look okay. Now we just gotta do something about the face cards...

GreenNinja2525:
@MeirGavish I redid the 9's and 10's to not only look like how you showed me, but I also think they look much better. Here is a screenshot of my current progress: Edit: the 9 of hearts is slightly off, the two hearts on the left side on the bottom of the card need shifted one pixel to the left. I'll fix this later
newdeckideav2
I also tried to fix the "Q" for the queens because it was driving me crazy because it looked like a "9" lol:
newqueen
I can upload the full card sheet if you want as well.
I would really appreciate some feedback. :)

GreenNinja2525:
New9
Sorry for the 9 being so big, I fixed the error I made before with it. (Ignore the pixels looking weird, I used a pixel editor with my phone to make the quick fix, I'll properly fix it on the full sheet later. I only posted this to show how the 9 was supposed to look from my previous post) Thanks, sorry for the constant posting, just wanted to gather some feedback! Thanks!

MeirGavish:
Looks good in my opinion, just in the 9 of clubs push the 2 clubs in the group of 4 like 1 pixel away from the group of 5. I currently don't have access to my PC so you can upload the full sprite sheet and I'll make a pull request with it later. I will close this one and we can see if @cellos51 approves the other.

GreenNinja2525:
deck_gfx_new
I believe I made the correct changes. I still think the simplified cards look good and would be a great option to either toggle to or use. I only really fixed the 9's and 10's, the face cards could still use some work but I think the cards I worked on are better than the default ones in this project. The simplified cards would look much better on an actual Game Boy Advance and the more detailed ones I helped to work on would look better in an Emulator.

GreenNinja2525:
Completely unrelated and I'm sorry that I'm posting this in the wrong place, but while I was fixing the cards I messed around with making a Joker card just to see how it could look on such a small card and I think I accidentally made something that almost looks like Flashcard Joker lol. I wanted to post this here just to show what else I was doing while editing. (I also left the cursor on screen too lol) Again sorry for posting a bunch and sorry for posting this Joker card in the wrong place, I just thought it looked funny. Maybe this could be a placeholder until someone makes actual Jokers in the future.
Jonkler

MeirGavish:
Looks good, I bet we can use it once Jokers are implemented, maybe you can already add the png in another pull request

PR #17, opened 2025-07-08
Five of a kind bug fix

MeirGavish:
Fixed a bug where five of a kind did not work


cellos51:
Lesbian Gay Transgender Men

PR #20, opened 2025-07-09
TTE Rect Refactor and More

MeirGavish:
So these are a few refactors from last commit to first
1. This is a big one, I replaced (almost) all the TTE rect magic numbers with named constants and added a wrapper struct and function for that. Of course I checked and everything seems to work the same, hopefully I didn't miss anything.
2. I created graphic_utils.h for these wrappers and moved other definitions there.
3. I extracted the erasing of the round end menu hack to a function. I added a TODO to refactor the round end animation code like you asked to reuse the code from the shop animation where it's properly. I may try it, but it seems like a lot of work and a little scary.

### DIFF
--- a/include/graphic_utils.h
+++ b/include/graphic_utils.h
@@ -1,6 +1,11 @@
 #ifndef GRAPHIC_UTILS_H
 #define GRAPHIC_UTILS_H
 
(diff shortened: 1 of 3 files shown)


cellos51 (approved):
Actually doing God's work my man

PR #21, opened 2025-07-10
Refactors - Added screen rect vertical copying function

MeirGavish:
I added a generic function that copies a screen rectangle vertically by one tile for the menu pop up animations etc.

I also looked into refactoring game_round_end() to use it but as I commented there are a bunch of other manual hard-coded tilemap animations there that are very hard to understand, and if I change the background image for the pop up it will change the tile charblock layout and screw them up so it's all or nothing...
I don't think I'll change it but it might be best to comment what each of the states does if you still remember...

By the way, I suppose using magic numbers for the states is not as bad as some of the others (since they're not reused) as long as you comment what each of them does. It's very important to comment and document your code but many times a comment can be replaced with code that's written in a clear self-documenting way. Another option there is to extract the contents of each case into a function with a self-documenting name.


cellos51:
If it wasn't made obvious, I didn't intend on anyone contributing to this which is why the code is like that. I'll try to remake the game round end to be better structured but primarily I just want this to be finished.

MeirGavish:
Yes, of course, I understand. Don't refactor game_round_end() unless you have to, it's a bit much and I don't mind so much, but if you do remember what the states do, just add it in comments for future reference at least

cellos51:
Yeah don't worry, I do remember. I'll push the comments in the next commit.

PR #31, opened 2025-07-13
Address potential Copyright and Legal Action through Disclaimer and Copyright Notice

GreenNinja2525:
I think the people that commented before about the copyright infringement and legal action are on to something. I feel this project is potentially at risk do to Playstack taking down the Commodore 64 port of Balatro and this project may not be entirely out of the water. After researching the legal reprecussions, I found that fan projects are generally not covered under fair use or through transformative works. After discovering and revisiting the comments from KevDoy and Cryhme, I see how this project could easily be taken down next by Playstack. What I am advising you do is to understand that after this project nears completion or if it does reach it, it may not last very long and could be taken down and issued a cease and desist. Until then, I am putting in the effort to help this project see the light of day. It may come down to renaming the project to make it harder for Playstack to discover this. I thought this commit is a step in the right direction to help convey this project is not meant to infringe or hurt the sales of Balatro in any way. I also added a steam link to encourage people to purchase the full game to demonstrate this.


MeirGavish:
I think this is a very good point. Also notice that the Steam demo was removed, according to [this discussion](https://steamcommunity.com/app/2379780/discussions/0/7434950195648667161/), because it had too much of the full game in it and they didn't want it detracting sales.
I think we may even want to aim for a smaller scope than that.

GreenNinja2525:
> I think this is a very good point. Also notice that the Steam demo was removed, according to [this discussion](https://steamcommunity.com/app/2379780/discussions/0/7434950195648667161/), because it had too much of the full game in it and they didn't want it detracting sales. I think we may even want to aim for a smaller scope than that.

That was exactly what I was thinking @MeirGavish .The January Demo has roughly like 120 jokers or so. We could definitely cut out some. Cellos51 could also change the project name. What if it was called Joker Poker? That was the name of the old prototype of Balatro before it was renamed. That would eliminate the copyright of the name of the game. The only remaining issue is the derivative works copyright but realistically the sprites are like 3x smaller than the original game so maybe it's fine, idk.

cellos51:
Yeah I do think that this should cover less content than the steam demo release. As for the name, I don't _entirely_ think a change is necessary. It honestly probably would help in avoiding a takedown, but it also feels a bit weird to like "keep a low profile" from PlayStack.

cellos51:
If we really wanted to _ensure_ it could stay up, we'd probably make the repo private, and do what KevDoy said in regards to using the full game as a key to unlock the rom or something.

GreenNinja2525:
That would help, but I think this project is still in it's infancy and is not even considered fully playable. But this is still a potential future issue that may need some attention.

GreenNinja2525:
Hey @cellos51, I messed up with my commits and it says 4 need merged when I only meant the one commit. If you merge this request, I can undo the disclaimer with another commit if you want, I only say this because I made the same change that @MeirGavish did and removed the one include line so I could compile the project. If you merge this, that would fix it on my end when it says my repo is 4 commits ahead, which is not correct. Sorry for this fumble on my part and I will be more careful next time I push a commit. Sorry! 😬

cellos51 (approved):

PR #38, opened 2025-07-19
Fixed screenblock overflow bug and once again merged up/down version of POP_MENU_ANIM_RECT

MeirGavish:
I think the screenblock bleeding issue you may have seen earlier was due to a bug I had with screenblock clipping when performing the vertical copy. There shouldn't be any overflows now.


cellos51:
thank you

PR #47, opened 2025-07-23
Moved the round end blind sprite and score a bit to the left so they look better for 4 digits

MeirGavish:
Also added TODO and warning note for CardObject and JokerObject


MeirGavish:
These are small changes so I'm merging them without a review but it's good that they're documented and you see them in a merge request

PR #48, opened 2025-07-23
Sprite vram optimizations

cellos51:
Previously I was loading the entire deck sprites statically into vram and this decreased the future scope of the game. Cards now only load their tiles into memory when needed, this makes it so only a maximum of 16 cards will be held in vram at a time.

### DIFF
--- a/source/card.c
+++ b/source/card.c
@@ -19,7 +19,6 @@ const static u16 card_sprite_lut[NUM_SUITS][NUM_RANKS] = {
 
 void card_init()
 {
-    //GRIT_CPY(&tile_mem[4], deck_gfxTiles);
     GRIT_CPY(&pal_obj_mem[CARD_PB], deck_gfxPal);
 }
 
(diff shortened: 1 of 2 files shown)


source/card.c
MeirGavish, line 1: Looks good here, the only thing is the commented out code, I think it's better to delete it altogether so no one wonders why it's there and why it's commented out, it seems like the intention here is to permanently move it and never uncomment it. The git history keeps track of what was there in case we need it.

source/game.c
MeirGavish, line 1: Just revert the initial game state back to GAME_BLIND_SELECT and the initial money amount.
I assume both were changed just for testing.

source/joker.c
MeirGavish, line 1: Same as before for the commented out code.

source/card.c
cellos51: Alright

MeirGavish (approved):

cellos51:
I think this should help a lot because now we have functionally unlimited sprite memory if i'm correct.

PR #50, opened 2025-07-23
Temporarily Revert "Sprite vram optimizations"

MeirGavish:
Reverts cellos51/balatro-gba#48

It seems there is a bug here that makes the card sprites override the blind sprites so I think the merge to main should be temporarily reverted until it is fixed on sprite-vram-optimizations and it can be merged back.
balatro-gba-0


cellos51:
oh this was my bad i did * instead of +

PR #53, opened 2025-07-24
Fixed sprite vram optimizations

cellos51:
I'm bad at git but i think this accounts for everything now


MeirGavish:
Yes, it seems correct in terms of git changes. Since Github's PR revert created a revert commit further in the tree, you needed another commit to revert the revert so it's a little messy...

So this time I ran the branch and tested it and I now see there's a small sprite flickering on the cards when discarding.
[discard_sprite_flicker_recording.webm](https://github.com/user-attachments/assets/9d4f2a48-fa7b-4abd-8ea0-c152123f2045)
It seems to happen on the frame the new card is drawn.
discard_sprite_flicker

cellos51:
Yeah that doesn't seem like something I'd really be able to fix. I think it's because it's writing to memory between scanlines or something and it hasn't written to the full tile data by then.

cellos51:
Yeah that doesn't seem like something I'd really be able to fix. I think it's because it's writing to memory between scanlines or something and it hasn't written to the full tile data by then.

It could definitely still be used on the jokers though since we would only have like 9 jokers maximum able to store statically, if all deck cards are loaded too.

MeirGavish:
Alright, I'll merge it and we can check if it can be fixed later.

cellos51:
I have an idea to fix the sprite flicker by only copying memory the first time the card object is created instead of each time the layer is changed. We'd just need to keep track of what it's tile id in memory is and also which tile slots are free to write to since they'll be using a different indexing system than the card layer.

PR #58, opened 2025-07-26
Created a CI Build Test using Github Actions

MeirGavish:
I created a test workflow using Github Actions that will trigger for any push action or for any pull request to the main branch. It will attempt to build running make on a clean Linux machine with the GBA dev environment installed. If make fails, the job will fail, this will automatically make sure any changes made by contributors don't break the build.


MeirGavish:
I'm merging this, I tested it on my fork and it should be good. It might temporarily fail if the devkitpro servers blocked us, but it will only be temporary.

PR #63, opened 2025-07-26
Extracted JokerObject and CardObject common functionalities to SpriteObject, removing duplicate code and dangerous casts.

MeirGavish:


MeirGavish:
Since this is a big refactor, can we request a review from copilot? I don't seem to have that option...

cellos51:
The copilot reviews might require GitHub pro

PR #68, opened 2025-07-28
Jokers 6-16, Refactor hand-type detection

theturtlemafia:
So this is a pretty big change, I also haven't tested out every single joker addition for bugs.

This changes the way hand-type is detected in hand_get_type() and extracts the logic into the following functions:
hand_contains_two_pair
hand_contains_full_house
hand_contains_straight
hand_contains_flush
hand_contains_n_of_a_kind (returns the highest of-a-kind in a hand, so a full-house hand would return 3)

I think it makes hand_get_type() a lot cleaner, but at the cost of scanning through the hand several times instead of once. It runs fine in mGBA, but not sure if this would affect performance on a real GBA or a more accurate emulator.

There's some additions that I think are a little ugly, but making it cleaner would mean a way bigger refactor to the hand and played cards systems, which might still be worth considering. This relies on the scored_card being NULL for the jokers that trigger after the full hand has scored.

Let me know what you think, and feel free to reject this PR or change as much as you like if you guys have a better idea for how to set up the code for these and future jokers.


MeirGavish:
Requesting myself to review, will review when I have time

samdiefenbacher:
I really want to try this branch, is there any way I check check the branch out? Can't do `git checkout` on it.

Never mind, I just checked out the PR. all good.

samdiefenbacher:
I've been testing this on my GBA-SP with an EZFLASH Omega and the performance feels fine. I know you mentioned performance could be a concern with scanning hand types multiple times, but I didn't notice any impact.

theturtlemafia:
Thanks @samdiefenbacher , appreciate it. I don't have real GBA hardware to test on so I really don't want to push code that works in an emulator but not the real thing. Thank you for checking

MeirGavish:
Honestly, I don't have a real GBA too. 😅
Sorry about the delay with this PR, I wanted to make sure the jokers palette system is figured out before moving to anything else, I will try to review it tomorrow.

theturtlemafia:
No problem, I probably won't be able to work on this again until the weekend

graphics/joker_gfx.png
MeirGavish, line 1: Okay, this will conflict with the joker palette system in my other pull request #70, that will need to be consolidated. I will add the new joker sprites to my PR and merge it in first.

source/hand_analysis.c
```diff
@@ -0,0 +1,83 @@
...
+    for (int i = 0; i < NUM_RANKS; i++) {
+        if (ranks[i] >= 2) {
+            if (contains_other_pair)
+                return true;
+            contains_other_pair = true;
+        }
+    }
+    return false;
+}
+
```
MeirGavish, line 46: Copilot pointed out hand_contains_full_house() fails for hands with 6 or more cards. I don't think this should ever become relevant but I'm quoting him here so at least it's documented. Not sure we should take his fix though, it's a little less readable and would need to be tested... Maybe just add a comment in the code linking to this review for in case it does become relevant?

Fails for hands with two trips (e.g., 7♠ 7♥ 7♣ 8♠ 8♥ 8♣).
To correctly detect a full house, especially in hands with more than 5 cards (such as if the game allows 6-card hands, or if jokers/multiples are present), hand_contains_full_house() should do the following:

Look for at least one rank with 3 or more cards (three-of-a-kind).
Look for a separate (different) rank with at least 2 cards (pair).
If there are two or more three-of-a-kinds, one serves as the "three," and the other as the "pair" (as in standard poker, e.g., [7,7,7,8,8,8] is a full house: 7s over 8s or vice versa).
Here’s a corrected version:

```
bool hand_contains_full_house(u8 *ranks) {
    int count_three = 0;
    int count_pair = 0;
    for (int i = 0; i < NUM_RANKS; i++) {
        if (ranks[i] >= 3) {
            count_three++;
        } else if (ranks[i] >= 2) {
            count_pair++;
        }
    }
    // Full house if there is:
    // - at least one three-of-a-kind and at least one other pair,
    // - OR at least two three-of-a-kinds (second "three" acts as pair).
    return (count_three >= 1 && (count_pair >= 1 || count_three >= 2));
}
```
Explanation:

If you have one or more three-of-a-kinds and at least one additional pair, it's a full house.
If you have two (or more) three-of-a-kinds, standard poker uses one as the triple and the other as the pair.
This covers all edge cases, including "double trips" and hands larger than five cards.

MeirGavish (commented):
Looks good, also ran it, seems to be running correctly for common basic gameplay cases.
I'll add the new joker sprites to my PR and merge that one in, then this would need to be updated so there are no conflicts.

graphics/joker_gfx.png
theturtlemafia: Sure thing, those are just placeholder graphics I made so the jokers don't look like missingno from pokemon. Feel free to delete/overwrite any graphics I add to a PR lol

source/hand_analysis.c
theturtlemafia: I think that's good feedback from Copilot. I'm not sure if it's possible for that to happen even in the real version of Balatro, but I doubt it understands the rules of Balatro that well. The final return there can be simplified to this which I think is more readable

```
return (count_three >= 2 || (count_three && count_pair));
```

But we can leave that out for now and commit it later if we want / after testing it.

MeirGavish: Nice simplification of that boolean expression! Now I think we should take it actually.

include/game.h
```diff
@@ -72,4 +72,11 @@ enum HandType
 void game_init();
 void game_update();
 
+// Utility functions for other files
+typedef struct CardObject CardObject; // forward declaration, actually declared in card.h
+CardObject **get_hand_array(void);
+CardObject **get_played_array(void);
+int get_hand_top(void);
+int get_played_top(void);
```
MeirGavish, line 80: By the way, is this what you meant when you said ugly additions? Yeah, the stacks need to be refactored into a generalized stack code.

theturtlemafia: Yeah exactly. These functions exist just for joker_effect.c to use. It's not terrible but also not very clean. But I think with these and a few more getters we could implement most or all of the joker's mentioned in the Joker Scope thread without refactoring all of the stacks.

MeirGavish:
I pushed a few commits to the branch to make it up to date but now GitHub is telling me I can't merge due to conflicts in the image file even though it's up to date, I'll try to resolve this issue tomorrow. If all else fails I will just bypass GitHub by merging locally and pushing to main

theturtlemafia:
I guess the github website doesn't like resolving merges on binary files. Merging locally might be easiest. I can also try force push a removal of my branch's joker_gfx.png

MeirGavish:
Merged locally and pushed, apparently it also marked this pull request as merged. Cool.

PR #73, opened 2025-08-01
Adds XMult, Joker Stencil and Misprint

theturtlemafia:
I added the additional Jokers that @GreenNinja2525 made art for, Stencil and Misprint. That includes implementing XMult for Stencil. I did not commit a changed spritesheet this time, since they would probably need to be changed anyways haha.

I added 2 more get functions to game.c, get_jokers() and get_jokers_top(). And moved the get functions closer to the top of the file. Stencil's function could be simpler if we know for sure 2 of the same joker will not happen, otherwise the current method will cover the case of multiple Stencils being owned.


source/joker_effects.c
```diff
@@ -1,7 +1,9 @@
 #include "joker.h"
 #include "util.h"
 #include "hand_analysis.h"
+#include <stdlib.h>
 
+static const int JOKER_STENCIL_ID = 16;
```
MeirGavish: This made me realize DEFAULT_JOKER_ID and GREEDY_JOKER_ID are the only ones defined and not actually used... I guess we shouldn't add ID definitions that are not being used but all these definitions should be in joker.h either way. Also I'm very tempted to turn the joker ID's into an enum, might be a bit of a refactor though.

source/joker_effects.c
```diff
@@ -184,6 +186,36 @@ static JokerEffect half_joker_effect(Joker *joker, Card *scored_card) {
...
+
+    return effect;
+}
+
+static JokerEffect misprint_joker_effect(Joker *joker, Card *scored_card) {
+    JokerEffect effect = {0};
+    if (scored_card != NULL)
+        return effect; // if card != null, we are not at the end-phase of scoring yet
+
+    effect.mult = random() % 24;
```
MeirGavish: I think just for readability this magic number should be a define, even right before this function. Also I wonder if there is a rand_range() function we could use instead of the modulo operator.

include/joker.h
```diff
@@ -39,7 +39,7 @@ typedef struct
     bool processed;
 } Joker;
 
-typedef struct // copy of CardObject in card.h
+typedef struct JokerObject // copy of CardObject in card.h
```
MeirGavish: Just noticed this comment, it's no longer true and should be removed. I should have done that in my refactor but since you're already making changes here it would be nice if you could.

source/game.c
```diff
@@ -214,6 +214,31 @@ static const Rect SHOP_REROLL_RECT          = {88,      96,    UNDEFINED, UNDEFI
 #define PITCH_STEP_UNDISCARD_SFX    2*PITCH_STEP_DRAW_SFX    
 // Naming the stage where cards return from the discard pile to the deck "undiscard"
 
+// get-functions, for other files to view game state (mainly for jokers)
```
MeirGavish: Since these were moved, maybe they should be even further up, next to the stack definitions and their push/pop functions.

MeirGavish (commented):
So, since you didn't add the sprites in, does that mean the jokers appear as garbled "MissingNo" sprites now? If so, I think they should be added in. Also I don't think they would be changed themselves, though they may need to be regrouped together with other sprites.
Did you understand the changes that need to be made to jokers.c to and and the graphics files to add another spritesheet?

Also, I'm all for Stencil taking other stencils into account even though that shouldn't happen in the current version of the game because that can happen in the original game and that is how it works in that case.

theturtlemafia:
Sure thing, I made most of the suggested changes, except the "random() % 24" one. C doesn't provide a stand rand_range() sort of function, they recommend the `random() % (ceiling+1)` method if you don't need something cryptographically secure, and dedicated libraries if you need a secure random. We could always turn that into a function or macro though in utils.

We can move it out to a define, but I'm not sure what a good name for it would be. Any suggestions? And then I'll commit that.

theturtlemafia:
About the changes to add another spritesheet, I understand them code-wise, but do the jokers need to be grouped in a way that they share the most similar colors or anything like that? I just don't want to break the palettes again 😂

MeirGavish:
Well, they should share up to 16 colors, I think you're fine if you're grouping Misprint and Joker Stencil, they don't use many colors as they are.
Let me know if you run into trouble there.

source/joker_effects.c
MeirGavish: I'll make this edit and commit it.

theturtlemafia: Thanks!

theturtlemafia:
@MeirGavish I committed the changes, I think it's correct. But when I build it, my colors are incorrect. However I also just noticed that when I build the main branch my colors are incorrect too. I wonder if it's something with my setup (I'm using the MacOS version of devkitpro/pacman), would you mind testing it out?

MeirGavish:
Tested it, it seems good! Merging.

theturtlemafia:
Thanks! I wonder if the incorrect colors from my first PR were because of this too

MeirGavish:
That might be, it is odd, maybe grit does something incorrectly in MacOS

MeirGavish:
image

These are the palettes I'm seeing when starting the game.

In mGBA Tools->Game state views->View palette

theturtlemafia:
image

These are mine. Weird how some are completely different and some are correct.

PR #75, opened 2025-08-02
Switching to a boss blind now changes the background palette

cellos51:
This is just a minor change that would make it so during a boss blind (and in the future anything else that changes BG color), the background will change to match the token's color theme like in the retail version.

image
image

### DIFF
--- a/include/graphic_utils.h
+++ b/include/graphic_utils.h
@@ -129,4 +129,13 @@ void update_text_rect_to_right_align_num(Rect* rect, int num, int overflow_direc
  */
 void memcpy16_tile8_with_palette_offset(u16* dst, const u16* src, uint hwcount, u8 palette_offset);
 
+/*Copies 32 bit data from src to dst, applying a palette offset to the data.
+ * This is intended solely for use with tile8/8bpp data for dst and src.
+ * The palette offset allows the tiles to use a different location in the palette
+ * memory
+ * This is useful because  grit always loads the palette to the beginning of 
+ * pal_bg_mem[]
+ */
+void memcpy32_tile8_with_palette_offset(u32* dst, const u32* src, uint hwcount, u8 palette_offset);
+
 #endif //GRAPHIC_UTILS_H
--- a/source/affine_background.c
+++ b/source/affine_background.c
@@ -8,7 +8,7 @@ AFF_SRC_EX asx = {32<<8, 64<<8, 120, 80, 0x0100, 0x0100, 0};
 
 void affine_background_init()
 {
-    memcpy16_tile8_with_palette_offset((u16*)tile8_mem[AFFINE_BG_CBB], (const u16*)affine_background_gfxTiles, affine_background_gfxTilesLen/2, AFFINE_BG_PB);
+    memcpy32_tile8_with_palette_offset((u32*)&tile8_mem[AFFINE_BG_CBB], (const u32*)affine_background_gfxTiles, affine_background_gfxTilesLen/4, AFFINE_BG_PB);
     GRIT_CPY(&se_mem[AFFINE_BG_SBB], affine_background_gfxMap);
     affine_background_load_palette(affine_background_gfxPal);
 
--- a/source/graphic_utils.c
+++ b/source/graphic_utils.c
@@ -170,9 +170,20 @@ void update_text_rect_to_right_align_num(Rect* rect, int num, int overflow_direc
 
 void memcpy16_tile8_with_palette_offset(u16* dst, const u16* src, uint hwcount, u8 palette_offset)
 {
+    const u16 offset = (((palette_offset) << 8) | (palette_offset));
     for (int i = 0; i < hwcount; i++)
     {
         // Copying u8 data twice across u16 data
-        dst[i] = src[i] + (((palette_offset) << 8) | (palette_offset));
+        dst[i] = src[i] + offset;
+    }
+}
+
+void memcpy32_tile8_with_palette_offset(u32* dst, const u32* src, uint hwcount, u8 palette_offset)
+{
+    const u32 offset = (palette_offset << 24) | (palette_offset << 16) | (palette_offset << 8) | palette_offset;
+    for (int i = 0; i < hwcount; i++)
+    {
+        // Copying u8 data 4 times across u32 data
+        dst[i] = src[i] + offset;
     }
 }


source/game.c
```diff
@@ -419,6 +420,13 @@ void change_background(int id)
             else if (current_blind == BOSS_BLIND)
             {
                 main_bg_se_copy_rect(BOSS_BLIND_TITLE_SRC_RECT, TOP_LEFT_BLIND_TITLE_POINT);
+                u16 palette[5] = {};
+                palette[0] = blind_get_color(current_blind, BLIND_BACKGROUND_SECONDARY_COLOR_INDEX);
+                palette[1] = blind_get_color(current_blind, BLIND_SHADOW_COLOR_INDEX);
+                palette[2] = blind_get_color(current_blind, BLIND_BACKGROUND_SHADOW_COLOR_INDEX);
+                palette[3] = blind_get_color(current_blind, BLIND_MAIN_COLOR_INDEX);
+                palette[4] = blind_get_color(current_blind, BLIND_TEXT_COLOR_INDEX);
+                memcpy16(&pal_bg_mem[16 * 15], palette, 5);
```
MeirGavish: Magic numbers - define AFFINE_BG_PB 15, and use PAL_ROW_LEN for the 16 or alternatively the pal_bg_bank array. Also I think you should define AFFINE_BG_PAL_LEN 5 or at least use NUM_ELEM_IN_ARR() for the length.

source/main.c
```diff
@@ -33,7 +33,7 @@ void init_map() // Temp function derived from libtonc example
         // Actually for future reference, each number seems to equate to an XY on the palette, so 0xF0F0F0F0 would just push everything down 15 colors, since the second number isn't set.
     }
 
-	memcpy(&tile8_mem[AFFINE_BG_CBB], correctedTiles, affine_background_gfxTilesLen);
+	memcpy16(&tile8_mem[AFFINE_BG_CBB], correctedTiles, affine_background_gfxTilesLen);
```
MeirGavish: Shouldn't it be affine_background_gfxTilesLen/2 like in the GRIT_CPY macro?

source/main.c
```diff
@@ -33,7 +33,7 @@ void init_map() // Temp function derived from libtonc example
...
     GRIT_CPY(&se_mem[AFFINE_BG_SBB], affine_background_gfxMap);
     memcpy16(&pal_bg_mem[16 * 15], affine_background_gfxPal, 5);
```
MeirGavish: Don't forget to replace the magic numbers here too

MeirGavish (commented):
Hey! Great to have another feature in this project from you! Looks cool! 
When more boss blinds are added, they will need a system to swap palettes (and sprites/tiles if you didn't add that for them) between RAM and VRAM. I thought of using grit for that to define the palettes instead of hardcoding them like they are now and I wonder if this would still work then.
I also thought of making the small blind and large blind use the same palette bank to optimize the sprite palette bank space and leave more for jokers etc.

cellos51:
Thanks! I've been on and still am on vacation right now which is why I haven't been active. I think all of your optimizations are objective improvements and having the blind palettes not be hard coded would be good. We could change the 'blind_get_color' function to fetch from sprite vram instead of reading from an array and that would make it more dynamic.

source/game.c
cellos51: I will define all the magic numbers created before I request merging

source/affine_background.c
```diff
@@ -0,0 +1,63 @@
...
+    memcpy16(&pal_bg_mem[AFFINE_BG_PB], affine_background_gfxPal, AFFINE_BG_PAL_LEN);
+    for (int i = 0; i < AFFINE_BG_PAL_LEN; i++)
+    {
+        COLOR old_color;
+        memcpy16(&old_color, &pal_bg_mem[AFFINE_BG_PB] + i, 1); // Copy the color from the background palette
+
+        int r, g, b;
+        r = (old_color >> 10) & 0b11111; // Extract red component
+        g = (old_color >> 5) & 0b11111; // Extract green component
+        b = old_color & 0b11111; // Extract blue component
```
MeirGavish: Use RED_MASK, RED_SHIFT, GREEN_MASK, GREEN_SHIFT, etc. (first mask then shift) from tonc_video.h. Also I think the color order is flipped here but seems like it didn't matter for your use here.

include/affine_background.h
```diff
@@ -0,0 +1,16 @@
+#ifndef AFFINE_BACKGROUND_H
+#define AFFINE_BACKGROUND_H
+
+#include <tonc.h>
+
+#include "graphic_utils.h"
+
+#define AFFINE_BG_PAL_LEN 5
+#define AFFINE_BG_PB (PAL_ROW_LEN * 15) // This isn't really a palette bank, just the starting index of the palette
+#define AFFINE_BG_DEFAULT_COLOR 0x3E4B
```
MeirGavish: Maybe using the RGB15 macro from tonc_video.h would make this a little clearer by showing the RGB values...
Maybe instead of changing the background to grayscale and hardcoding this value, it would be better to extract it from the background on startup and store it for when you need to change back?
I think it would allow easier editing for the background and the default color, up to you.

source/affine_background.c
```diff
@@ -0,0 +1,63 @@
...
+
+        int brightness = (r + g + b) / 3; // Calculate brightness
+
+        const float brightness_divergence = 0.03f; // The difference in brightness between each color in the palette
+        const float brightness_modifier = -0.4f; // The overall brightness modifier for the palette
+
+        clr_adj_brightness(&pal_bg_mem[AFFINE_BG_PB] + i, &color, 1, float2fx(brightness_modifier + (brightness * brightness_divergence))); // Adjust brightness for the palette colors
```
MeirGavish: Note that this operation includes floating point arithmetic so it will be relatively slow. Since it's done only 5 times and only on background change I don't think this should be an issue but keep that in mind. 
I didn't see any performance issue when testing on emulators, did you get a chance to test this on real hardware?

source/affine_background.c
```diff
@@ -0,0 +1,63 @@
...
+AFF_SRC_EX asx = {32<<8, 64<<8, 120, 80, 0x0100, 0x0100, 0};
+
+void affine_background_init()
+{
+    unsigned int correctedTiles[affine_background_gfxTilesLen / 4];
+
+    for (int i = 0; i < affine_background_gfxTilesLen / 4; i++)
+    {   
+        correctedTiles[i] = affine_background_gfxTiles[i] | 0xF0F0F0F0;
+    }
```
MeirGavish: In the previous file you had comments explaining this, don't remove them, they were helpful for understanding it.
Also now looking at it I think the "+" operator would be more correct than the "|" operator here.

source/affine_background.c
```diff
@@ -0,0 +1,63 @@
...
+        g = (old_color >> 5) & 0b11111; // Extract green component
+        b = old_color & 0b11111; // Extract blue component
+
+        int brightness = (r + g + b) / 3; // Calculate brightness
+
+        const float brightness_divergence = 0.03f; // The difference in brightness between each color in the palette
+        const float brightness_modifier = -0.4f; // The overall brightness modifier for the palette
+
+        clr_adj_brightness(&pal_bg_mem[AFFINE_BG_PB] + i, &color, 1, float2fx(brightness_modifier + (brightness * brightness_divergence))); // Adjust brightness for the palette colors
+    }
```
MeirGavish, line 40: Also, I think this entire color mapping operation could have been done with [clr_rgbscale() from tonc](https://github.com/devkitPro/libtonc/blob/ccc03fa321e56f51aed5e2ee1d6e3df3d1cbc803/src/tonc_color.c#L188), so maybe consider using it at least in case this doesn't work for other colors or we run into performance issues...
I think it might look a little different though.

source/affine_background.c
```diff
@@ -0,0 +1,63 @@
...
+    {
+        COLOR old_color;
+        memcpy16(&old_color, &pal_bg_mem[AFFINE_BG_PB] + i, 1); // Copy the color from the background palette
+
+        int r, g, b;
+        r = (old_color >> 10) & 0b11111; // Extract red component
+        g = (old_color >> 5) & 0b11111; // Extract green component
+        b = old_color & 0b11111; // Extract blue component
+
+        int brightness = (r + g + b) / 3; // Calculate brightness
```
MeirGavish: Also maybe this operation here can be extracted to a generic "extract_brightness(COLOR)" in graphic_utils

MeirGavish: Not very important, resolving

source/affine_background.c
MeirGavish: Also, are you sure you need floats here? Maybe using FIXED and fxmul() would do the trick?

source/affine_background.c
cellos51: Oh that's literally what i was doing already. Yeah, I just wasn't aware of that.

cellos51:
Premature review whoops

cellos51:
I think this covers most of your comments now. I reverted the grayscale of the background but it still uses a hard-coded value, to me this makes more sense and you could always just bypass it by just loading the palette directly from the image at any point. Of course, if you want the palette to just be taken from the image that's fine too, I just couldn't personally think of a consistent way to do it.

cellos51:
Ok, disregard my previous message. I changed it to load from the image instead of a color.

source/affine_background.c
```diff
@@ -0,0 +1,46 @@
...
+void affine_background_set_color(COLOR color)
+{
+    memcpy16(&pal_bg_mem[AFFINE_BG_PB], affine_background_gfxPal, AFFINE_BG_PAL_LEN);
+    for (int i = 0; i < AFFINE_BG_PAL_LEN; i++)
+    {
+        clr_rgbscale(&pal_bg_mem[AFFINE_BG_PB] + i, &pal_bg_mem[AFFINE_BG_PB] + i, 1, color);
+    }
+}
+
+void affine_background_load_palette(const void *src)
```
MeirGavish: This should be u16 * instead of void*. Use void* only if you really need a generic type and you absolutely have to because using void* loses type safety and can become dangerous.

source/affine_background.c
MeirGavish: Cool!

include/graphic_utils.h
```diff
@@ -120,4 +120,13 @@ void tte_erase_rect_wrapper(Rect rect);
...
 void update_text_rect_to_right_align_num(Rect* rect, int num, int overflow_direction);
 
-#endif //GRAPHIC_UTILS_H
\ No newline at end of file
+/*Copies 16 bit data from src to dst, applying a palette offset to the data.
+ * This is intended solely for use with tile8/8bpp data for dst and src.
+ * The palette offset allows the tiles to use a different location in the palette
+ * memory
+ * This is useful because  grit always loads the palette to the beginning of 
+ * pal_bg_mem[]
```
MeirGavish, line 128: Note I added some documentation here

source/graphic_utils.c
```diff
@@ -168,3 +168,14 @@ void update_text_rect_to_right_align_num(Rect* rect, int num, int overflow_direc
     }
 }
 
+void memcpy16_tile8_with_palette_offset(void* dst, const void* src, uint hwcount, u8 palette_offset)
```
MeirGavish: This should also be changed from void* (and cast when calling the function if needed) but honestly I'm a little confused as for the use of 16-bit here... Supposedly 8BPP tiles use 32-bit values but I still don't fully understand why and how exactly they map into the palette memory, and thus what's going on in this function and why it works.
I guess I need to go further in depth in the tiled backgrounds tutorial to understand this.
For me, this can stay as it is in the meantime and only change the void* to u16*, I guess there's no point of changing it without understanding.

include/affine_background.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include "graphic_utils.h"
+
+#define AFFINE_BG_PAL_LEN 5
+#define AFFINE_BG_PB (PAL_ROW_LEN * 15) // This isn't really a palette bank, just the starting index of the palette
+
+void affine_background_init();
+void affine_background_update();
+void affine_background_set_color(COLOR color);
+// Must be called with an array of size at least  AFFINE_BG_PAL_LEN
```
MeirGavish, line 14: Note I added some documentation here too

source/graphic_utils.c
cellos51: Honestly, I don't know why but I could only get it to work properly with u16. If it was u8 it would functionally half the horizontal resolution and draw every even vertical line twice. I brute forced the creation of this function because I couldn't find any real documentation on the formatting of tiles beyond a few random blog posts.

MeirGavish: I'll look into it and try to figure it out. I think it makes sense that u8 wouldn't work because VRAM doesn't allow 8-bit writes, they have to be in at least 16-bits, that's why memcpy etc. are problematic and you need memcpy16 etc.
But 32-bit should be okay, and 8BPP tiles are defined as 32-bit.

source/affine_background.c
```diff
@@ -8,7 +8,7 @@ AFF_SRC_EX asx = {32<<8, 64<<8, 120, 80, 0x0100, 0x0100, 0};
 
 void affine_background_init()
 {
-    memcpy16_tile8_with_palette_offset(&tile8_mem[AFFINE_BG_CBB], affine_background_gfxTiles, affine_background_gfxTilesLen/2, AFFINE_BG_PB);
+    memcpy16_tile8_with_palette_offset((u16*)tile8_mem[AFFINE_BG_CBB], (const u16*)affine_background_gfxTiles, affine_background_gfxTilesLen/2, AFFINE_BG_PB);
```
MeirGavish: It should still be (u16*)&tile8_mem[AFFINE_BG_CBB] so you send the address and not the value...

MeirGavish (commented):
Just one last thing

source/graphic_utils.c
MeirGavish: I hope the [tonc chapter on backgrounds](https://gbadev.net/tonc/regbg.html) would hold the answer

cellos51: I mean I could do this:

```c
void memcpy32_tile8_with_palette_offset(u32* dst, const u32* src, uint hwcount, u8 palette_offset)
{
    for (int i = 0; i < hwcount; i++)
    {
        // Copying u8 data 4 times across u32 data
        dst[i] = src[i] + (((palette_offset) << 24) | ((palette_offset) << 16) | ((palette_offset) << 8) | (palette_offset));
    }
}
```

cellos51:
Sorry for requesting so many reviews

source/graphic_utils.c
MeirGavish: It does feel like the safer way to do this

MeirGavish:
No problem, thanks for taking all my comments into account.

PR #82, opened 2025-08-06
Changed Clubs corner symbol to improve card readability

MeirGavish:
I made a small change to the small corner Clubs symbols on the cards.
The main goal is to differentiate them from the Spades symbol and make them more readable, especially on the face cards where this icon is the main way to recognize the suit.
My only concern is that they look a little different from the 5x5 Clubs symbol on the numbered cards that has a different stem but it does look similar to the big symbol on the Ace card.
I would like to know your thoughts.
@GreenNinja2525


GreenNinja2525:
I think it looks good, it definitely helps change up the symbols because before it just was a smaller spades symbol. I think this is a nice change that makes it easier to quickly see which cards you have. Thanks for asking my opinion, I think this change is very much needed. Thanks! 👍

PR #84, opened 2025-08-08
Added Joker art updates by @GreenNinja2525 with some minor tweaks

MeirGavish:


MeirGavish:
@GreenNinja2525 
I thought I left a message here but don't see it now. I reverted this commit because the phone app your were using seems to have messed up the color palette and create visual bugs. I couldn't fix it with LibreSprite or usenti.

PR #89, opened 2025-08-09
Card corner fixes

MeirGavish:
There was an issue with some of the jokers where their corners are sharp for some reason, so I rounded them


GreenNinja2525:
I think that was a error on my fault when I exported them, I copy and pasted over each joker when I was drawing them and I might have forgot to fix the corners before uploading them, sorry about that (if it was me because I think it is) @MeirGavish

PR #92, opened 2025-08-09
Joker pooling + Shop code refactor

MeirGavish:
This changes the behavior so Jokers already owned can't appear in the shop (and it can't have duplicates)

There are a lot of changes but that's just because it includes a refactor.

1. I added  a list (plus list API) that tracks which jokers are available to shop and they are always chosen from there.
2. One refactor was extracting functions from game_shop() to clean the code up a bit.
3. Another was removing the first_time parameter from game_shop_create_items() and extracting it's functionality into game_shop_reroll().

@cellos51 I'm requesting a review, I think this code should be okay, if you have time just take a look at my changes and see that I didn't break anything in the existing flow.


MeirGavish:
Cool, thanks!

PR #91, opened 2025-08-09
HBlank based visual effects for affine background

cellos51:
This changes the affine background to use more visually interesting animations which I think is more in line with how the vibe of Balatro feels.

https://github.com/user-attachments/assets/33cf4b28-9997-45e2-9c24-c10f1564e10b

PR #93, opened 2025-08-10
Main menu and splash screen

cellos51:
This change adds a main menu and disclaimer splash screen {which I'm still working on).
The splash screen will describe how this is an open source project and will show a link to the github before the main menu appears.

This change also functions as game over logic, when you lose you will be taken back to the main menu.

image

### DIFF
--- a/include/graphic_utils.h
+++ b/include/graphic_utils.h
@@ -24,7 +24,6 @@
 #define AFFINE_BG_SBB 2
 #define AFFINE_BG_CBB 2
 #define PAL_ROW_LEN 16
-#define PAL_BG_MAX_SIZE (PAL_ROW_LEN * 9)
 #define NUM_PALETTES 16
 
 #define TTE_BIT_UNPACK_OFFSET 14


cellos51:
Oh boy this is gonna be a fun one ain't it

cellos51:
Yeah I might need some help with this one

MeirGavish:
Looks cool! Yeah, it does seem like a big one, restarting will be hard because there are so many things in the global scope that need to be restarted. I honestly thought of just making a game over screen and letting the user just restart the console.
I want to take a look but unfortunately it seems like I won't have time to do the reviews today, I will try tomorrow.

cellos51:
Alright then. I might just leave out the game over functionality for now and add that in as a separate PR later.

cellos51:
So this is the final design as of now. I can't do anymore until #91 is merged. I think #91 should just have the HBlank effect in code but not used for the game and only the menu since there isn't anything going on during it that would cause lag.
image

source/affine_background.c
```diff
@@ -20,20 +21,37 @@ void affine_background_update()
...
+    }
+
     bg_rotscale_ex(&bgaff, &asx);
     REG_BG_AFFINE[2] = bgaff;
 }
 
 void affine_background_set_color(COLOR color)
 {
-    memcpy16(&pal_bg_mem[AFFINE_BG_PB], affine_background_gfxPal, AFFINE_BG_PAL_LEN);
+    memcpy16(&pal_bg_mem[AFFINE_BG_PB], affine_main_menu_background_gfxPal, AFFINE_BG_PAL_LEN);
```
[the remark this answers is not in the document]
cellos51: WRONG

source/game.c
```diff
@@ -1028,6 +1056,24 @@ void game_init()
...
+
+void game_start()
+{
+    set_seed(rng_seed);
+
+    affine_background_change_background(AFFINE_BG_GAME);
+
+    // Normally I would just cache these and hide/unhide but I didn't feel like dealing with defining a layer for it
+    card_destroy(&main_menu_ace->card);
+    card_object_destroy(&main_menu_ace);
```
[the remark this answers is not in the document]
cellos51: WRONG

source/game.c
```diff
@@ -1028,6 +1056,24 @@ void game_init()
...
+}
+
+void game_start()
+{
+    set_seed(rng_seed);
+
+    affine_background_change_background(AFFINE_BG_GAME);
+
+    // Normally I would just cache these and hide/unhide but I didn't feel like dealing with defining a layer for it
+    card_destroy(&main_menu_ace->card);
```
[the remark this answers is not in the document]
cellos51: WRONG

cellos51:
@GreenNinja2525 Since you wrote a majority of the README, I'd appreciate it if you gave your opinion on this:

image

cellos51:
Made a branch called 'experimental' to merge this and #91 and I think it came out perfect.

https://github.com/user-attachments/assets/ec8974b0-4ede-4b93-b508-09a3f4fc63e8

GreenNinja2525:
> @GreenNinja2525 Since you wrote a majority of the README, I'd appreciate it if you gave your opinion on this:

Thanks for asking me! I think it is great except for the line with the GitHub link, in my opinion it just looks a bit funky. Otherwise I think it looks good and gets the message across very well. Great job @cellos51 and I think credits should be included as a button on the main menu instead of adding a "collection" button because I think that we talked before about not dealing with unlocking cards/jokers and just having that feature omitted. That credits thing can come later of course, just a small suggestion I had. Anyways, I think the disclaimer is very good, as the full disclaimer is on the GitHub page

MeirGavish:
Yeah it looks absolutely great! Now I really want to use same background image for the rounds and just set its color with affine_background_set_color(). 
Maybe just use the same affine function without the HBlank functionality if it looks okay...

cellos51:
this is what it looks like with setting the color 
image

source/affine_background.c
```diff
@@ -20,20 +22,37 @@ void affine_background_update()
...
+    switch (background)
+    {
+    case AFFINE_BG_MAIN_MENU:
+        asx.scr_x = SCREEN_WIDTH / 2;
+        asx.scr_y = SCREEN_HEIGHT / 2;
+        asx.tex_x = 16560;
+        asx.tex_y = 14400;
+        asx.sx = 128;
+        asx.sy = 128;
+        asx.alpha = 1000; // This slightly rotates it which increases the perceived resolution of the background
```
MeirGavish, line 33: These are all constants so the entire asx struct can be defined as a const for this case and just used in bg_rotscale_ex() instead of copying the values every frame, this should improve performance, unless the compiler optimizes this, not sure.

Also, if you want to keep the switch case here, break it up into functions for each case so the logic is separated and this function doesn't grow too big.

source/game.c
```diff
@@ -241,6 +243,8 @@ static const Rect CASHOUT_RECT              = {88,      72,     UNDEFINED, UNDEF
 static const Rect SHOP_REROLL_RECT          = {88,      96,    UNDEFINED, UNDEFINED };
 
 //TODO: Properly define and use
+#define MAIN_MENU_BUTTONS 4
```
MeirGavish: The comment needs to go to the next define where it was before... a long awaited TODO...

source/game.c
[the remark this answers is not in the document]
MeirGavish: Interesting that it was so wrong about this, it's quite obvious from the function signature.

source/game.c
```diff
@@ -1028,6 +1056,24 @@ void game_init()
     obj_hide(blind_select_tokens[BIG_BLIND]->obj);
     obj_hide(blind_select_tokens[BOSS_BLIND]->obj);
 
+    game_set_state(game_state);
+}
+
+void game_start()
+{
+    set_seed(rng_seed);
```
MeirGavish, line 1112: I didn't notice that set_seed() is also called in game_round_init(), it should be removed from there, the seed shouldn't be set for every round.

source/game.c
```diff
@@ -1028,6 +1056,24 @@ void game_init()
...
+
+    affine_background_change_background(AFFINE_BG_GAME);
+
+    // Normally I would just cache these and hide/unhide but I didn't feel like dealing with defining a layer for it
```
MeirGavish, line 1116: I actually think the current approach is the better one, both in terms of saving memory - the card shouldn't occupy memory outside the main menu where it isn't used, and in terms for logical game state - it doesn't make sense for the ace card to be valid outside the main menu, otherwise it could open the door for side effects and bugs.

source/game.c
```diff
@@ -257,12 +261,8 @@ int get_num_discards_remaining(void) {
...
 void set_seed(int seed)
 {
-    if (!seeded)
-    {
-        seeded = true;
-        rng_seed = seed;
-        srand(rng_seed);
-    }
+    rng_seed = seed;
+    srand(rng_seed);
```
MeirGavish, line 277: Notice that removing this check, you now need to remove set_seed() from game_round_init() like I said in my other comment so it's actually seeded every run and not per round (ruining randomization)

source/game.c
```diff
@@ -2639,23 +2683,70 @@ void game_blind_select()
...
+        if (selection_x > 0)
         {
-            rng_seed *= 2;
+            selection_x--;
         }
     }
+    else if (key_hit(KEY_RIGHT))
+    {
+        // TODO: Don't allow cursor on sold items
+        if (selection_x < MAIN_MENU_BUTTONS)
```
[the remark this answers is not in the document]
MeirGavish: This is also not true

source/main.c
```diff
@@ -71,7 +70,26 @@ void init()
...
+    {
+        VBlankIntrWait();
+        key_poll();
+        if (key_hit(KEY_A | KEY_B | KEY_START | KEY_SELECT)) // Skip intro
+        {
+            break;
+        }
+        tte_erase_rect_wrapper((Rect){208, 144, 240, 152});
+        tte_printf("#{P:208,144; cx:0xF000}%d", 1 + (SPLASH_DURATION - i) / FPS);
+    }
```
MeirGavish: The splash screen should definitely be extracted to a different function, it's not part of the init logic and doesn't quite belong in main.c. 
In fact I think it should be extracted to another game state rather than have its own duplicate game loop.

source/main.c
```diff
@@ -18,15 +18,14 @@
 #include "soundbank.h"
 #include "soundbank_bin.h"
 
+#define FPS (60)
+#define SPLASH_DURATION (FPS * 10) // FPS * SECONDS
```
MeirGavish: I think it's better to change to SPLASH_DURATION_FRAMES to specify the units. And maybe also define SPLASH_DURATION_SECONDS 10 and then you can remove the comment.

source/main.c
```diff
@@ -18,15 +18,14 @@
 #include "soundbank.h"
 #include "soundbank_bin.h"
 
+#define FPS (60)
```
MeirGavish: I think this should be renamed SPLASH_FPS, just to signify this is only used for the splash screen and not imply that the game will always run at the same speed.

MeirGavish (commented):
I added my review comments.

For the splash screen, I think the issue with the URL is that it's not aligned, maybe just align it to the rest of the lines like this?
balatro-gba-0

Also just keep in mind that if the repository is to be made private, the rom might still go around with a link to a repository that can no longer be reached, not that I'm saying we should not include it, just pointing it out.

Also this is just my thoughts, but maybe it's better to go for a more neutral tone and mention scamming? Maybe "This is a non-profit project and this rom should not be sold"?
@GreenNinja2525

cellos51:
> Also this is just my thoughts, but maybe it's better to go for a more neutral/formal tone and not mention scamming? Maybe "This is a non-profit project and this rom is not to be sold"?

The reason I put scamming there is because of the off chance someone makes a bootleg cartridge with our rom. I don't realistically think it would happen to us but I have seen it happen to others. To be fair though, if someone was tech-savvy enough to get it on a cart in the first place, they'd probably also know enough to just remove the splash screen. But I'm fine with changing that.

source/main.c
cellos51: I don't really want it to be a part of game.c because it has nothing to do with the game and it should only be shown once.

source/game.c
cellos51: Noted.

source/affine_background.c
cellos51: I'll extract it into a function but is there a specific reason to do it like that or is it just preference?

source/main.c
MeirGavish: Don't put it in game.c then (it's quite an overly large file as it is), extract it to splash_screen.c. I think it could still have a game state and I think you can expose game_set_state() for it and add a function to do all the things you do just once when moving from the splash screen to the game. It could check a static variable to make sure it's done just once.
In general I think game.c should probably be separated into a file for each state, but that would be a bigger refactor...

source/affine_background.c
MeirGavish: It's cleaner, separating the logic for each case and avoiding clogging the function with a lot of logic that would make it grow too long. I'm not saying separate every switch case to functions but once each case has more than a few simple lines of code you should so it doesn't become too long.  
Long functions (especially with many levels of nested logic) are difficult to read and maintain, I think a good guideline to go by is every function should perform only a single task and fit within the screen. If there are several tasks they should probably be separated into functions (if there are several loops they are good candidates for separation). Of course this isn't a hard rule you should always follow, just a guideline for more maintainable code. 
This function is only on the verge of becoming too long (unless another case is added) but it also feels like since the background types behave quite differently, their logic should be separated.

source/game.c
```diff
@@ -2639,23 +2683,65 @@ void game_blind_select()
...
+    {
+        if (selection_x > 0)
         {
-            rng_seed *= 2;
+            selection_x--;
         }
     }
+    else if (key_hit(KEY_RIGHT))
+    {
+        if (selection_x < MAIN_MENU_BUTTONS)
```
MeirGavish: By the way, I forgot something here, I think we should not allow the cursor to reach buttons that are not implemented, it's a little weird that you can go right but there are no buttons there. Maybe  change the number to 1 for now or add a define for MAIN_MENU_IMPLEMENTED_BUTTONS

MeirGavish:
> The reason I put scamming there is because of the off chance someone makes a bootleg cartridge with our rom. I don't realistically think it would happen to us but I have seen it happen to others. To be fair though, if someone was tech-savvy enough to get it on a cart in the first place, they'd probably also know enough to just remove the splash screen. But I'm fine with changing that.

I'm also fine with it either way, you can keep the warning, I'm not sure if removing the splash screen is on the same level of tech-savviness as just putting it on a cart.

source/affine_background.c
MeirGavish: I also want to point out that this case is not critical, this function isn't that long and complex, I just wanted to fully explain my reasons

cellos51: Thank you for explaining. I'd just rather know why I'm doing something instead of blindly following instructions.

MeirGavish: Of course, you should! I try to always give a brief reason for my comments and they are always open for discussion, some of them are more suggestions rather than things I would insist on.

MeirGavish:
Also, maybe the "experimental" branch can be merged in? It looks better than the static background and it doesn't cause any performance issues if it's just in the main menu, right?

cellos51:
Yeah it would be safe to merge but it doesn't have the proposed changes applied to it yet

source/main.c
```diff
@@ -71,7 +71,26 @@ void init()
...
 
+    // Splash screen
+    tte_printf("#{P:72,8; cx:0xF000}DISCLAIMER");
+    tte_printf("#{P:8,24; cx:0xF000}This project is NOT endorsed \n by or affiliated with \n Playstack or LocalThunk.\n\n If you have paid for this, \n you have been scammed \n and should request a refund \n IMMEDIATELY. \n\n The only official place \n to obtain this is from: \n\n 'github.com/\n  cellos51/balatro-gba'");
+    tte_printf("#{P:8,144; cx:0xF000}(Press any key to skip)");
+    for (int i = 0; i < SPLASH_DURATION_FRAMES; i++)
+    {
+        VBlankIntrWait();
+        key_poll();
+        if (key_hit(KEY_A | KEY_B | KEY_START | KEY_SELECT)) // Skip intro
```
MeirGavish: Something else I forgot here, the splash screen says "Press any key" but you didn't include the shoulder buttons or directional keys, is that intentional? Otherwise you can use KEY_ANY

cellos51: Yeah, that was a copilot addition and I just forgot to change it

cellos51:
I think I've covered the listed changes

source/game.c
```diff
@@ -2824,6 +2876,54 @@ void game_blind_select()
...
+    if (key_hit(KEY_LEFT))
+    {
+        if (selection_x > 0)
+        {
+            selection_x--;
+        }
+    }
+    else if (key_hit(KEY_RIGHT))
+    {
+        if (selection_x < MAIN_MENU_IMPLEMENTED_BUTTONS)
```
```suggestion
        if (selection_x < MAIN_MENU_IMPLEMENTED_BUTTONS - 1)
```

Because it's 0-based

cellos51:
@MeirGavish 🥺

source/splash_screen.c
```diff
@@ -0,0 +1,30 @@
...
+    tte_printf("#{P:72,8; cx:0xF000}DISCLAIMER");
+    tte_printf("#{P:8,24; cx:0xF000}This project is NOT endorsed \n by or affiliated with \n Playstack or LocalThunk.\n\n If you have paid for this, \n you have been scammed \n and should request a refund \n IMMEDIATELY. \n\n The only official place \n to obtain this is from: \n\n 'github.com/\n  cellos51/balatro-gba'");
+    tte_printf("#{P:8,144; cx:0xF000}(Press any key to skip)");
+}
+
+void splash_screen_update(uint timer)
+{
+    if (timer < SPLASH_DURATION_FRAMES)
+    {
+        tte_erase_rect_wrapper((Rect){208, 144, 240, 152});
```
MeirGavish: Give this rect a name to make it readable, I guess it's the countdown timer digits? Use it in the tte_printf too

source/game.c
```diff
@@ -627,6 +631,19 @@ void change_background(int id)
...
+        toggle_windows(false, false);
+
+        tte_erase_screen();
+        memcpy16(pal_bg_mem, background_main_menu_gfxPal, PAL_ROW_LEN * 9);
+        GRIT_CPY(&tile_mem[MAIN_BG_CBB], background_main_menu_gfxTiles);
+        GRIT_CPY(&se_mem[MAIN_BG_SBB], background_main_menu_gfxMap);
+
+        // Disable the button highlight colors
+        // Select button PID is 5 and the outline is 3
+        memcpy16(&pal_bg_mem[3], &pal_bg_mem[5], 1);
```
MeirGavish: Name the palette indices in defines

source/main.c
```diff
@@ -44,7 +40,7 @@ void init()
     // BG1 is the main background layer
     REG_BG1CNT = BG_PRIO(1) | BG_CBB(MAIN_BG_CBB) | BG_SBB(MAIN_BG_SBB) | BG_8BPP;
 	// BG2 is the affine background layer
-    REG_BG2CNT = BG_PRIO(2) | BG_CBB(AFFINE_BG_CBB) | BG_SBB(AFFINE_BG_SBB) | BG_8BPP | BG_WRAP | BG_AFF_32x32;
+    REG_BG2CNT = BG_PRIO(2) | BG_CBB(AFFINE_BG_CBB) | BG_SBB(AFFINE_BG_CBB) | BG_8BPP | BG_WRAP;
```
MeirGavish:
```suggestion
    REG_BG2CNT = BG_PRIO(2) | BG_CBB(AFFINE_BG_CBB) | BG_SBB(AFFINE_BG_SBB) | BG_8BPP | BG_WRAP;
```

I think this was missed in the merge

cellos51:
> Also, I noticed that the mobile and console versions of Balatro have only 3 buttons in the main menu because "Quit" doesn't make much sense for them and I think it doesn't make much sense for us either. I guess we won't ever make a "Collection" button either...

So should there be just no buttons then?

GreenNinja2525:
> So should there be just no buttons then?

Unless continuing a run gets implemented, then a continue button could be added. I don't think autosaving or saving will get added, so I don't think that is a viable option either. Maybe just center the play button and remove the other slots, that is just my suggestion. Or if you want to add a credits page that lists the main contributors

cellos51:
@GreenNinja2525 ok 2 buttons it is then

source/game.c
```diff
@@ -492,16 +498,16 @@ void change_background(int id)
...
         main_bg_se_clear_rect(ROUND_END_MENU_RECT);
         tte_erase_rect_wrapper(HAND_SIZE_RECT);
     }
     else if (id == BG_ID_SHOP)
     {
-        REG_DISPCNT &= ~DCNT_WIN0;
+        toggle_windows(false, true);
 
-        memcpy(pal_bg_mem, background_shop_gfxPal, 64);
+        memcpy16(pal_bg_mem, background_shop_gfxPal, 64);
```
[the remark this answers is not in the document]
cellos51: That is a preexisting constant from tonc that is 512 😔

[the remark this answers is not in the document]
MeirGavish: Actually, I think the correct solution here would be to add the -pn 32 flag in grit (which should limit the size of the buffer it creates for the palette) and then use GRIT_CPY.
I think 32 would be the correct value because it's 32 colors which are 64 bytes.
I haven't tried it but I think that would work.

MeirGavish:
Maybe we would also make "Options" at some point but I don't know...

GreenNinja2525:
> Maybe we would also make "Options" at some point but I don't know...

An options page sounds good, many options can be brought over except for the crt filter stuff and stake stickers. There can be audio settings and game settings, but for the graphics tab in the settings,  there is not a lot of stuff left because of the crt thing. Having a three button layout on the title screen would match the console ports but instead of the collection button, it would be something else. I imagine the buttons like this: (Play) (Options) (Credits) unless you want to put the credits in the settings button and the main menu looks like this: (Play) (Options)

MeirGavish (approved):
Looks good, hopefully didn't miss anything, just note my reply to copilot about the 64 buffer size magic number.

source/game.c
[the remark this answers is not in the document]
cellos51: yeah i tried and i don't think those flags work

cellos51:
@MeirGavish is this ok to be merged now? the only change was replacing the magic numbers.

MeirGavish:
So you're using up all the palette space up until the affine background instead? Did you run it to make sure it doesn't do anything funky? If so yes. I still think the grit solution is better but we can leave it as a separate task. Also I think we should extract a vram_map.h header and have all these defines concentrated there.

cellos51:
I agree that the grit solution is better too but it doesn't seem like those flags work in my testing anyway. Making the palette that size for the background doesn't cause any issues.

cellos51:
actually no, the flags do work. maybe i messed something up.

cellos51:
Ok so I think this PR is finished?

source/game.c
```diff
@@ -2824,6 +2876,54 @@ void game_blind_select()
...
+    main_menu_ace->sprite_object->trotation = lu_sin((timer << 8) / 2) / 3;
+    main_menu_ace->sprite_object->rotation = main_menu_ace->sprite_object->trotation;
+
+    // Seed randomization
+    rng_seed++;
+    if (key_curr_state() != key_prev_state()) // If the keys have changed, make it more pseudo-random
+    {
+        rng_seed *= 2;
+    }
+
```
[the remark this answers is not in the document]
MeirGavish: He may have a point here but that's something to look at on another day.

MeirGavish:
Yes, I think, if everything seems okay on your testing you can merge.

PR #96, opened 2025-08-11
Optimized affine background HBLANK effect 

MeirGavish:
I made some optimizations and there is an improvement but it's still not good enough, there is still text flickering and slowing and artifacts in the audio.

I moved the code to IWRAM to speed it up and moved the calculations to a VBLANK function that would pre-compute the values so the HBLANK function can just fetch them.

I also don't encounter the 3 first pixels issue anymore, not sure what caused it.

I tried to do some more fiddling for  optimization but to no avail. Maybe we can use some performance profiling tools to see which functions take the most time and try to optimize them.

### DIFF
--- a/include/affine_background.h
+++ b/include/affine_background.h
@@ -7,7 +7,7 @@
 
 #define AFFINE_BG_PAL_LEN 5
 #define AFFINE_BG_PB (PAL_ROW_LEN * 15) // This isn't really a palette bank, just the starting index of the palette
-#define AFFINE_BG_IDX 2 // The index of BGCNT register etc.
+#define AFFINE_BG_IDX 2 // The index of the affine background BGCNT register etc.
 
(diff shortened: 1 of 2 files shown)


cellos51 (approved):

cellos51:
Yeah there are definitely other things that can be optimized. I'm pretty sure the cards use floating point math 💀

cellos51:
I don't know if it would work but you could also try updating the background every other frame or interlacing the background effect

MeirGavish:
Hm... I thought they were only converting constants and assumed it was fine but now that I look at it, maybe that's a good place to optimize

MeirGavish:
Made separate branch PR for that #97, it didn't seem to make a enough difference but I thought it might be good to have regardless of whether we use the HBLANK effect.

PR #97, opened 2025-08-12
Removed all uses of float for optimization

MeirGavish:
Removed all uses of float for optimization.

Marking this as a draft since I still want to carefully check that there are no visual differences as some of the numbers are approximations of the previous ones.

### DIFF
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -163,7 +163,7 @@ void sprite_object_update(SpriteObject* sprite_object)
     sprite_object->vrotation += (sprite_object->trotation - sprite_object->rotation) / 8; // Rotate the card when it's played
 
     // set velocity to 0 if it's close enough to the target
-    const FIXED epsilon = FIX_ONE>>7; // 1/(2^7) = 1/128 = 0.0078125
+    const FIXED epsilon = float2fx(0.01f);
     if (sprite_object->vx < epsilon && sprite_object->vx > -epsilon && sprite_object->vy < epsilon && sprite_object->vy > -epsilon)
     {
         sprite_object->vx = 0;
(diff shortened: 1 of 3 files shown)


source/sprite.c
```diff
@@ -163,7 +163,7 @@ void sprite_object_update(SpriteObject* sprite_object)
     sprite_object->vrotation += (sprite_object->trotation - sprite_object->rotation) / 8; // Rotate the card when it's played
 
     // set velocity to 0 if it's close enough to the target
-    const float epsilon = float2fx(0.01f);
+    const FIXED epsilon = FIX_ONE>>7; // 1/(2^7) = 1/128 = 0.0078125
```
cellos51: I would assume that the compiler optimizes this and makes it unnecessary and I know I'm the one who originally wrote this, but wouldn't this be better as a define? genuine question.

MeirGavish: You mean for performance precomputing the value and placing it here instead of doing the shift? Yeah, the compiler optimizes this. I've been struggling to find a way to write fixed point constants in a readable way and the shift is the best I could.
As for magic numbers and readability, the constant is put directly into a named variable so it's readable, and is not reused as a constant so it's maintainable as a single source.
But now that you mention it, I checked by looking at the disassembly, and the compiler also optimizes float2fx so I'll revert it, it's both more readable, and functionally unchanged from the original.
What IS important is changing the type of epsilon to FIXED from float, it does save a ton of floating point operations on the comparisons and improve performance.
This is the disassembly on the main branch, you can see all the **bl <____aeabi_fcmp...>** instructions that jump to floating point routines.

image

And they're gone with the optimization.
image

cellos51: Oh cool. Are you viewing the assembly in vscode or something? Because that would be useful for me.

MeirGavish: Yes! In vscode using [a debug configuration](https://github.com/cellos51/balatro-gba/discussions/80#discussion-8676631)  set up a breakpoint and once it hits, right click and there's an "Open Disassembly View" option down below.
image

Pretty cool

Issue #99, opened 2025-08-14
Joker Selling Mechanic


MeirGavish:
Selling jokers is an important mechanic in Balatro and I think we should have it. In the PS5 version you sell jokers with L1, so I think we can allow selling with the left shoulder button too and leave the A button for rearranging in the future.
I thought of making the same focus UX as the shop and cards in hand, raising the card by 10 pixels when it is focused.
For displaying the price, I thought of just showing it below the focused joker, either directly below the now raised joker, or another tile down below where it was before raising.
For the selling animation, I think it can be the same as the discard animation, just make it fly off screen.

In the original Balatro you can sell jokers at the shop and during rounds (maybe in other states?) but for implementation simplicity I think we can just allow selling jokers in the shop. 
It means we probably won't implement the Luchador joker, but boss blind effects are also quite far off.

MeirGavish:
Assigned myself and started working on it

PR #102, opened 2025-08-16
Joker Selling

MeirGavish:
This implements joker selling, including animation as defined in issue #99.
I used the left shoulder button to sell but I wonder if it's a little unintuitive.
To allow removing arbitrary jokers I changed the jokers stack to be a list.


GreenNinja2525:
@MeirGavish The Left shoulder button is pretty well suited for Joker selling. Perhaps I can add controls to the Readme page for those that possibly get confused, if that would help. I don't think deck peeking will ever be added, unless the select button gets bound to it instead so the left shoulder button should be good. I was the one that suggested that sort rank/suit gets bound to the right shoulder button but I feel like that is a little unintuitive. I'll see what I can do to make this easier for those who get confused.

cellos51 (approved):
Everything looks good. Could you also add another selection row for selling during gameplay?

MeirGavish:
> Everything looks good. Could you also add another selection row for selling during gameplay?

Theoretically yes, but it would take some work since the round selection would have to be refactored to use the selection grid system.

PR #105, opened 2025-08-21
Game over screen

### LINKED ISSUE #104: Game Over Screen
There were already several issues opened where people ran out of hands and thought it's a bug they couldn't do anything then.
We need a game over screen to indicate to the player that he lost.
The first basic version of this would show a menu message with a "Game Over" text. It's important to use the same round background and overlay it on that so the player sees the state the game was in and how he lost.
The original Balatro game over screen has many stats for the run, there's no need to include them currently. The rest of the game is under a transparent red filter behind the game over menu. This will be difficult to do, instead we can use affine_background_set_color() to change the color of the affine background (in the same way it changes for the boss blinds) to a red color.
In the future this menu could have a "New Run" and/or "Main Menu" buttons but this will not be part of this task as restarting the run will take some additional work.

MeirGavish:
Closes #104
Added both a losing "Game Over" screen when the player runs out of hands but doesn't beat the blind and a winning "You Win" screen when the player beats the last ante.
I'm not sure it's possible to beat ante 8 with the current implemented features but might as well.
When reaching a game over state, the round end hand discard animation plays and then the screen pops up from the bottom.
I added the discarding animations for 2 reasons,
1. So I can disable the window and have the game over window visible and the hand background will not be visibly opaque.
2. So the popping dialog will not pass under the card sprites and look weird.
I didn't create a dedicated background or used change_background() because I wanted to keep the same background that was present when the winning/losing event occurred.

game_over_large
you_win_large

This also includes several refactors in separate commits.
1. I added defines for the text colors.
2. In graphic_utils.c/.h, renamed a few functions and optimized  main_bg_se_copy_rect()
3. Extracted card_in_hand_loop_handle_discard_and_shuffling() from cards_in_hand_update_loop() to reduce function size and depth and make it more readable.
4. Moved a few functions in game.c to a better position.


source/main.c
```diff
@@ -28,15 +28,15 @@ void init()
...
+    tte_init_se(0, BG_CBB(TTE_CBB) | BG_SBB(TTE_SBB), 0, CLR_WHITE, TTE_BIT_UNPACK_OFFSET, NULL, NULL);
     tte_erase_screen();
     tte_init_con();
 
     // TTE palette setup
-    pal_bg_bank[12][15] = 0x029F; // Yellow. honestly fuck libtonc because i cannot figure out how you're supposed to select a color from the palette index so i'm doing it like this
-    pal_bg_bank[13][15] = 0x7E40; // Blue
-    pal_bg_bank[14][15] = 0x213F; // Red
-    pal_bg_bank[15][15] = CLR_WHITE;
+    pal_bg_bank[TTE_YELLOW_PB][TTE_BIT_ON_CLR_IDX] = TEXT_CLR_YELLOW; // honestly fuck libtonc because i cannot figure out how you're supposed to select a color from the palette index so i'm doing it like this
```
cellos51: You probably don't need to have that comment there anymore 😭

cellos51 (approved):
Looks pretty good! Now people are gonna start making issues asking how to restart the game lol.

source/main.c
MeirGavish: lol I'll remove it

MeirGavish:
Nice! We can open an issue for adding a restart button lol

PR #107, opened 2025-08-22
Additional change for Blue Joker

MeirGavish:
@GreenNinja2525 
I made a small tweak to the neckpiece, what do you think?
image


GreenNinja2525:
@MeirGavish That looks much better, I couldn't figure it out (the blue spot things) and you seemingly managed to do it. That looks so much better! I also worked on Smiley Face and Scary face in the Joker art discussion. I think Smiley Face could use some more work. Thanks for fixing Blue Joker! You are epic 👍

GreenNinja2525:
Actually, hold on a second @MeirGavish, the card shine is a little off next to his neck piece as it cuts off and is not connected to him anymore. The bottom left of his neck piece and the bottom right have a disconnect from the card shine, I just noticed it after looking at it again. Just a small little nitpick sorry lol. I would fix it real quick but I am not at my computer right now

MeirGavish:
Thanks! Good point, I'll extend the shine so it reaches the joker

MeirGavish:
Blue_joker_large_fixed

There

GreenNinja2525:
That's great! Now Blue Joker can be marked as complete now. Thanks!

PR #117, opened 2025-09-11
Add docker support

ricfehr3:
Add docker support via a docker compose file with included instructions.

I've struggled with devkitpro setups and I've found it easier to just use their [official container](https://hub.docker.com/r/devkitpro/devkitarm). This compose file is pretty straight forward. 

I unfortunately cannot test this on windows right now. In the past, similar setups worked fine for me on windows, but if someone has docker desktop installed on windows already, mind confirming for me if I can't sooner :) .

### DIFF
--- a/.github/workflows/build_ci_workflow.yml
+++ b/.github/workflows/build_ci_workflow.yml
@@ -8,14 +8,43 @@ on:
     branches:
       - main
   workflow_dispatch:
+    inputs:
+      clear_cache:
+        description: 'Clear docker cache before running'
+        type: boolean
+        default: false
 
 jobs:
   build:
     runs-on: ubuntu-latest
-
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+
+      - name: Get docker image cache
+        uses: actions/cache@v4
+        with:
+          path: /tmp/docker-cache
+          key: docker-images-${{ hashFiles('docker-compose.yml') }}
+
+      - name: Clear docker cache (if requested)
+        if: github.event.inputs.clear_cache == 'true'
+        run: |
+          echo "Clearing Docker cache..."
+          rm -rf /tmp/docker-cache
+          echo "Docker cache cleared"
+
+      - name: Load or pull images
+        run: |
+          if [ -f /tmp/docker-cache/balatro.tar ]; then
+            echo "Loading Image"
+            docker load -i /tmp/docker-cache/balatro.tar
+          else
+            echo "Pulling Image"
+            docker compose pull balatro
+            mkdir -p /tmp/docker-cache
+            docker save $(docker compose config | grep image: | awk '{print $2}') -o /tmp/docker-cache/balatro.tar
+          fi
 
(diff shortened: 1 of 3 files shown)


ricfehr3:
I realize now this project has a CI pipeline, so I'll just leave this here but no worries if you don't want repeat work on the docker part. I do find this easier, regardless. 

You may prefer this file over the github actions workflow and I wouldn't mind converting the `build_ci_workflow.yml` file to use docker compose in the runner. As long as you're okay trusting the container source from the docker repository, I think it would be a cleaner approach.

MeirGavish:
Cool! I don't have much experience with Docker but I'm assigning myself to review. I will probably only be able to look at this once I'm back from vacation after October 5th.

MeirGavish:
@ricfehr3 The workflow is actually failing now because it receives a 403 when trying to get devkitpro, it seems like GitHub runners are being blocked due to excessive requests, we do cache it for that reason but I suppose the requests are coming from many forks now... Do you think using the docker image there would solve the problem? If you could check I'll appreciate it, if not, I'll probably do it once I get back.

ricfehr3:
@MeirGavish Hey, it was no problem to convert it to use docker compose in my fork.

It will be pulling from the docker container repository. There are some set limits per IP. I'm not sure how that works across github runners (_I figure this repos runner will be separate from forks, forks disable actions by default_). 

This solution has a cache to store the docker image. Each build it will try to load the cached image. Otherwise if it's not found, `github` actions will pull and cache it. **Note, it will not automatically pull the latest image**. That may be a feature for later if it's a problem, but devkitpro doesn't get many updates so I figure it will be fine.

If you do want to manually delete the cached image (_maybe cronjob this daily/weekly or whatever_). You can run the action manualy and check the box `Clear docker cache before running`. 

Hope this helps! 

**Note**: I still haven't tested on windows. Just want to let you know I'd like to hold off on merging until I can verify it.

ricfehr3:
@MeirGavish Just confirmed it works on windows following my instructions (download docker desktop and run `docker compose up`). 

Let me know if you want some changes, but I think it's good to merge here.

Also! I see I have completely polluted the actions workflow runs. Really sorry about that, I did quite a few iterations, lmao. 

I'm not sure why it's still there, but I had introduced a temporary new workflow `clear-docker-cache.yml`. That file is gone, I think you may have to delete the logs for those runs? Not sure, just wanted to bring that to your attention.

MeirGavish:
Awesome! I wish I had more time to give this an in-depth review and learn some Docker on the way, but this seems to be the "urgent fix" for our CI pipeline issues so I'm gonna merge it, thanks a bunch!

ricfehr3:
> Awesome! I wish I had more time to give this an in-depth review and learn some Docker on the way, but this seems to be the "urgent fix" for our CI pipeline issues so I'm gonna merge it, thanks a bunch!

Sweet! Just let me know if something is wrong or if you wanted some explanation later. I'm not going anywhere, lmao.

PR #119, opened 2025-09-13
Add blueprint and brainstorm jokers

Nanrech:
Implements blueprint and brainstorm to the game, but leaves the implementations commented out as they are still missing assets.


Nanrech:
I notice now that there are assets made for both jokers. I don't know how to go about adding them so I've just left them commented out as I mentioned in the commit message.

MeirGavish:
It seems like the build run failed on setting up devkitpro for getting 403, apparently it's not cached. This happens because of excessive requests to devkitpro, might be from other forks, we'll have to wait until they unblock the runners and run it again.

MeirGavish:
I merged PR #117 which seems to have fixed the CI test issue, can you merge main into your branch? Then I think you will get the updated workflow and it will pass.

Nanrech:
There you go

MeirGavish:
Great! The build is passing now which is good, I will give the code a review when I can, probably only after I return on October 5th. Just one question, since you couldn't add the art, does that mean you didn't test this?

Nanrech:
No, I did test this. I replaced scary face and smiley face's functionality temporarily with my code while I was working on it so I could see them in action.

MeirGavish:
Okay, great, thanks!

include/joker.h
```diff
@@ -36,6 +36,8 @@
 #define DEFAULT_JOKER_ID 0
 #define GREEDY_JOKER_ID 1 // This is just an example to show the patern of making joker IDs
 #define JOKER_STENCIL_ID 16
+#define JOKER_BLUEPRINT_ID 99 // This needs to be changed
+#define JOKER_BRAINSTORM_ID 99 // This needs to be changed
```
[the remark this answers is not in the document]
MeirGavish: This is true even if the IDs are temporary, it will avoid potential issues.

source/joker_effects.c
```diff
@@ -464,6 +464,47 @@ static JokerEffect odd_todd_joker_effect(Joker *joker, Card *scored_card) {
     return effect;
 }
 
+__attribute__((unused))
+static JokerEffect blueprint_joker_effect(Joker *joker, Card *scored_card) {
+    JokerEffect effect = {0};
+    List* jokers = get_jokers();
+    bool trigger_next_joker = false;
+
+    for (int i = 0; i < list_get_size(jokers); i++ ) {
```
MeirGavish: A bit awkward doing this with a search for loop but since the effect function has no way of knowing the position of the joker, I guess it's the most sensible thing...

source/joker_effects.c
```diff
@@ -464,6 +464,47 @@ static JokerEffect odd_todd_joker_effect(Joker *joker, Card *scored_card) {
...
+        JokerObject* joker_object = list_get(jokers, i);
```
```suggestion
        JokerObject* curr_joker_object = list_get(jokers, i);
```

Rename the variable to make it more readable - which is the iterated joker and which is the effect joker

source/joker_effects.c
```diff
@@ -464,6 +464,47 @@ static JokerEffect odd_todd_joker_effect(Joker *joker, Card *scored_card) {
...
+        if (trigger_next_joker) {
+            effect = joker_get_score_effect(joker_object->joker, scored_card);
+            break;
+        }
+
+        // JOKER_BLUEPRINT_ID (joker.h) will need to be updated
+        if (joker_object->joker->id == JOKER_BLUEPRINT_ID)
```
```suggestion
        if (curr_joker_object->joke == joker)
```

Using the ID is problematic, in case there are multiple blueprints, since they all have the same ID, this will only activate the first Blueprint so I suggest using pointer comparison to uniquely identify this joker.

source/joker_effects.c
```diff
@@ -464,6 +464,47 @@ static JokerEffect odd_todd_joker_effect(Joker *joker, Card *scored_card) {
...
+static JokerEffect brainstorm_joker_effect(Joker *joker, Card *scored_card) {
+    JokerEffect effect = {0};
+    if (joker->processed)
+        return effect;
+
+    List* jokers = get_jokers();
+    JokerObject* first_joker = list_get(jokers, 0);
+
+    // JOKER_BRAINSTORM_ID (joker.h) will need to be updated
+    if (first_joker->joker->id != JOKER_BRAINSTORM_ID) {
```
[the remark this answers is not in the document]
MeirGavish: It's always good to have a NULL-check but I think in practice the list should never be empty when this is called since it will always contain Brainstorm at least.

source/joker_effects.c
```diff
@@ -464,6 +464,47 @@ static JokerEffect odd_todd_joker_effect(Joker *joker, Card *scored_card) {
...
 
+__attribute__((unused))
+static JokerEffect blueprint_joker_effect(Joker *joker, Card *scored_card) {
+    JokerEffect effect = {0};
+    List* jokers = get_jokers();
+    bool trigger_next_joker = false;
+
+    for (int i = 0; i < list_get_size(jokers); i++ ) {
+        JokerObject* joker_object = list_get(jokers, i);
+        if (trigger_next_joker) {
```
MeirGavish: Why not just activate the i+1 joker while checking you don't go over the list size rather than postpone it to the next iteration?
It's cleaner and more readable.

source/joker_effects.c
```diff
@@ -464,6 +464,47 @@ static JokerEffect odd_todd_joker_effect(Joker *joker, Card *scored_card) {
...
+
+    List* jokers = get_jokers();
+    JokerObject* first_joker = list_get(jokers, 0);
+
+    // JOKER_BRAINSTORM_ID (joker.h) will need to be updated
+    if (first_joker->joker->id != JOKER_BRAINSTORM_ID) {
+        // manually flip this to avoid infinite blueprint + brainstorm loops
+        joker->processed = true;
+        effect = joker_get_score_effect(first_joker->joker, scored_card);
+        joker->processed = false;
```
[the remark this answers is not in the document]
MeirGavish: This is redundant with the previous comment, resolving

MeirGavish: This is a bit of a hacky way to solve this problem, this variable is actually meant for the game code to keep track of which jokers it already activated and I really don't think the jokers themselves should touch it, there has got to be a better way to do this.

Nanrech: There are definitely better ways to handle this. I'm not a fan of touching that variable either but I didn't really know if adding an entirely new variable or messing with memory in any other way were good ideas.

MeirGavish: Adding a new variable would be better

Nanrech:
Made a bit of a mess on my side trying to fix the merge conflicts. Could you check if I missed/deleted anything important? @MeirGavish

source/joker_effects.c
Nanrech: I have changed brainstorm's logic so it uses a separate, static variable instead of joker->processed.

source/joker_effects.c
```diff
@@ -637,6 +678,14 @@ const JokerInfo joker_registry[] = {
     // The following jokers don't have sprites yet, 
     // uncomment them when their sprites are added.
 #if 0
+    { COMMON_JOKER, 5, raised_fist_joker_effect },
+    { COMMON_JOKER, 6, reserved_parking_joker_effect },
+
+    { COMMON_JOKER, 4, abstract_joker_effect },
+    { UNCOMMON_JOKER, 6, bull_joker_effect},
```
MeirGavish: Yes, there's a merge error here, Raised Fist, Reserved Parking, Abstract Joker and Bull are all active now but they are duplicated in the inactive zone.

Nanrech: Got it, fixed.

source/joker_effects.c
MeirGavish: Actually using a static variable here turns out to be quite nice.

source/joker_effects.c
MeirGavish: @Nanrech Just this left

Nanrech: I don't think I understand what you're saying here

source/joker_effects.c
```diff
@@ -582,6 +582,47 @@ static JokerEffect triboulet_joker_effect(Joker *joker, Card *scored_card) {
...
+        JokerObject* curr_joker_object = list_get(jokers, i);
+        if (trigger_next_joker) {
+            effect = joker_get_score_effect(curr_joker_object->joker, scored_card);
+            break;
+        }
+
+        // JOKER_BLUEPRINT_ID (joker.h) will need to be updated
+        if (curr_joker_object->joker == joker)
+            trigger_next_joker = true;
+    }
```
```suggestion
    int list_size = list_get_size(jokers);
    
    for (int i = 0; i < list_size - 1; i++ ) {
        JokerObject* curr_joker_object = list_get(jokers, i);
        if (curr_joker_object->joker == joker) {
            JokerObject* next_joker_object = list_get(jokers, i + 1);
            effect = joker_get_score_effect(next_joker_object->joker, scored_card);
            break;
        }
    }
```

I meant doing it like this.
It does need to be tested though, I intend to test this PR before merging.

source/joker_effects.c
MeirGavish: I made a suggestion with my intention

source/joker_effects.c
Nanrech: I see. Let me test this really quickly.

Nanrech: LGTM

MeirGavish:
Cool, I'll integrate the art straight into this PR, test it, and merge if everything is okay.

MeirGavish:
@Nanrech 
Really cool to have Blueprint and Brainstorm, thanks!

Nanrech:
Thank you for reviewing my PR!

PR #121, opened 2025-09-14
Removed redundant copy of shadow buffer into OAM.

drewjohnson2:
This may make me a shameless one-line-change-merchant, but since obj_mem and oam_mem point to the same place I figured the extra copy was redundant.

### DIFF
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -98,7 +98,6 @@ void sprite_init()
 
 void sprite_draw()
 {
-    obj_copy(obj_mem, obj_buffer, MAX_SPRITES);
     obj_aff_copy(obj_aff_mem, obj_aff_buffer, MAX_AFFINES);
     oam_copy(oam_mem, obj_buffer, MAX_SPRITES);
 }


cellos51:
Why are they named different things if they do the same thing 😭

source/sprite.c
```diff
@@ -98,7 +98,6 @@ void sprite_init()
 
 void sprite_draw()
 {
-    obj_copy(obj_mem, obj_buffer, MAX_SPRITES);
     obj_aff_copy(obj_aff_mem, obj_aff_buffer, MAX_AFFINES);
     oam_copy(oam_mem, obj_buffer, MAX_SPRITES);
```
cellos51, line 102: ig it would look more consistent to comment out the oam function imo if they're identical, but it doesn't matter

cellos51 (approved):

source/sprite.c
drewjohnson2: @cellos51 The way I understand it is that `oam_copy` performs better than `obj_copy`, but I think that comes with the caveat that you only see that performance boost for affine and non-affine objects?

Idk, it's weird. But honestly I don't mind that, what really bothers me is that `oam_mem` and `obj_mem` both point to the exact same place. There has to be a reason for it, right?

This thing is your baby, man. If you'd like me to go back and keep `obj_copy` and remove `oam_copy` I can 100% do that for you.

cellos51: literally i have no idea what i'm doing. I just read tonc and was like "yeah i can probably do this"

if you think it performs better, then it probably does.

drewjohnson2: @cellos51 Yeah. Same here.

Does the build usually fail on a 403? It doesn't matter if I'm working or programming for fun, I can never escape failing builds on PRs.

cellos51: @drewjohnson2 I think it's just not working right now. 

I tested it myself, and obviously it does work, so it can probably be merged. But I want to see what @MeirGavish  has to say first, because he didn't merge a previous pull request in relation to the CI tests I think.

MeirGavish: It seems to have been failing for all PRs, seems like devkitpro blocked GitHub runners for excessive requests.
Either way I merged PR #117 which switches to use the docker image instead and it seems to have fixed it. Can you merge main into your branch? Then I think  you will get the updated test and it will pass.

drewjohnson2: @MeirGavish Have you experienced any weird permission issues with anyone attempting to merge? Initially I tried pushing, but it failed because I didn't have a workflow permission. So I tried generating a token with that permission and now it gives me a 403 when I try to push to my fork.

MeirGavish: Odd, haven't experienced anything like that. You can't push to your own fork? Have you tried syncing it? 
Either way I will merge since @cellos51 tested it and approved and I don't want to send you into a rabbit hole just to get that green tick mark. Just make sure you can still work with your fork I guess.

PR #123, opened 2025-09-16
Adding an x macro to cut down on some boilerplate.

drewjohnson2:
This should make adding additional jokers require a little less typing. Additional jokers will need an entry in `/include/joker_table.h`.

I'm not aware of any downsides of using x macros in this context. I've seen them used in other games, most notably Ocarina of Time, but never in a GBA game.


source/joker.c
```diff
@@ -116,39 +116,12 @@ void joker_init()
...
-    joker_gfxPal[11] = joker_gfx11Pal;
-    joker_gfxPal[12] = joker_gfx12Pal;
-    joker_gfxPal[13] = joker_gfx13Pal;
-    joker_gfxPal[14] = joker_gfx14Pal;
+
+#define DEF_JOKER_GFX(idx) 							\
+	joker_gfxTiles[idx] = joker_gfx##idx##Tiles;	\
+	joker_gfxPal[idx] = joker_gfx##idx##Pal;
+#include "../include/joker_table.h"
+#undef DEF_JOKER_GFX
```
ricfehr3: I don't see any problem with using an X macro, it's far cleaner. There's an opportunity here to refactor out the `malloc` calls above I've seen from my investigation: https://github.com/cellos51/balatro-gba/issues/124 .

On lines 116-117 you can remove the `malloc`'s in favor of allocating the arrays statically:

```c
#define MAX_JOKERS 15
static const unsigned int* joker_gfxTiles[MAX_JOKERS];
static const unsigned short* joker_gfxPal[MAX_JOKERS];
```

When you remove those `malloc` calls, you can then use your X macro at declaration, cleaning up this function call further. 

Not sure how to tackle the `joker_spritesheet_pb_map`, but that could be considered outside of the scope of this PR.

drewjohnson2: @ricfehr3 Got it taken care of.

source/joker.c
```diff
@@ -13,9 +13,20 @@
 
 #define JOKER_SCORE_TEXT_Y 48
 #define NUM_JOKERS_PER_SPRITESHEET 2
+#define MAX_JOKERS 15
 
-const unsigned int **joker_gfxTiles;
-const unsigned short **joker_gfxPal;
+const unsigned int *joker_gfxTiles[MAX_JOKERS] =
```
ricfehr3: Nitpick: These should be declared static, consistent with declarations below.

source/joker.c
ricfehr3: Thanks @drewjohnson2 !

source/joker.c
drewjohnson2: Can't believe I missed that. Idk why I assumed they already were.

drewjohnson2: @ricfehr3 all done

source/joker.c
```diff
@@ -13,9 +13,20 @@
 
 #define JOKER_SCORE_TEXT_Y 48
 #define NUM_JOKERS_PER_SPRITESHEET 2
+#define MAX_JOKERS 15
```
MeirGavish: Since it seems like it's only used to define a static array, I think you can remove this define and leave the brackets in the definition empty, it would be much better because this define is one more thing to update.

If I missed something and you must have this define, its name is too generic and misleading, it should be renamed NUM_JOKER_SPRITESHEETS.

include/def_joker_gfx_table.h
MeirGavish, line 1: The name of this header is too generic, it does something very, very specific, it should be renamed def_joker_gfx_calls.h

MeirGavish (commented):
This does improve the process of adding joker graphics. However, I've never seen a header used like this and it does make me a little uneasy that it can be included from other places and call an undefined macro...
Maybe a comment within the header explaining its use at least can help...

Otherwise see my comments about naming.

drewjohnson2:
> This does improve the process of adding joker graphics. However, I've never seen a header used like this and it does make me a little uneasy that it can be included from other places and call an undefined macro... Maybe a comment within the header explaining its use at least can help...
> 
> Otherwise see my comments about naming.

@MeirGavish This is a pretty common meta-programming trick. 

As you know, when you `#include` a header you're essentially telling the preprocessor "Cut and paste the contents of `foo.h` here" ("here" being the location of the `#include`). This is true for all `#include`s. 

What you see here is the definition of a macro, the we `#include "../joker_table.h`, which contains calls to the macro you just defined. After those calls are finished we `#undef` the macro, so we don't pollute the namespace and so we can reuse `joker_table.h` if a situation allows. It's not any different from calling a macro manually in the file (other than the ability to reuse the macro declaration).

You are correct though that some more descriptive names are needed here. I'll also add a comment explaining use.

MeirGavish:
Yeah, I understand. Still, what makes me uneasy is making it possible to include the header from other places and having it call an undefined macro...

ricfehr3:
I second what @drewjohnson2 Says. "X macros" are used everywhere. It's a great way to cut down on boiler plate. For example, I'm using it for the memory pool pr:

X Macro declaration:
https://github.com/cellos51/balatro-gba/blob/0a202557c3324a304a0c65f0fd95b73182d3128d/include/pools.def

Used here to declare a type:
https://github.com/cellos51/balatro-gba/blob/0a202557c3324a304a0c65f0fd95b73182d3128d/include/pool.h

Then define the multiple types:
https://github.com/cellos51/balatro-gba/blob/0a202557c3324a304a0c65f0fd95b73182d3128d/source/pool.c

A huge benefit of this approach is extensibility. You only need to add to that X macro template. 

Had to check the wikipedia page: https://en.wikipedia.org/wiki/X_macro "remains relatively unknown". That's a bit funny to note.

drewjohnson2:
You'll get build error if you try and include the header and haven't defined the macro.

It would be the same result as trying to call a function from another file you haven't included.

drewjohnson2:
Unlike many vague and difficult to debug macro errors, in this case if you include the header without defining the macro you'll see this error when building.

```
/home/drew/source/balatro-gba/source/../include/joker_table.h:1:1: error: implicit declaration of function 'DEF_JOKER_GFX' [-Wimplicit-function-declaration]
    1 | DEF_JOKER_GFX(0)
      | ^~~~~~~~~~~~~
/home/drew/source/balatro-gba/source/../include/joker_table.h:1:1: error: initialization of 'const unsigned int *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
/home/drew/source/balatro-gba/source/../include/joker_table.h:1:1: note: (near initialization for 'joker_gfxTiles[0]')
/home/drew/source/balatro-gba/source/../include/joker_table.h:1:1: error: initializer element is not constant
/home/drew/source/balatro-gba/source/../include/joker_table.h:1:1: note: (near initialization for 'joker_gfxTiles[0]')
/home/drew/source/balatro-gba/source/../include/joker_table.h:2:1: error: expected '}' before 'DEF_JOKER_GFX'
    2 | DEF_JOKER_GFX(1)
      | ^~~~~~~~~~~~~
/home/drew/source/balatro-gba/source/joker.c:19:1: note: to match this '{'
   19 | {
      | ^
make[1]: *** [/opt/devkitpro/devkitARM/base_rules:40: joker.o] Error 1
make: *** [Makefile:129: build] Error 2

```

This is a pretty explicit error, especially for the world of macros.

MeirGavish:
Yeah, I suppose it's okay and I'm just fretting because it's a macro with things defined in separate files, this can be merged once my other comments are addressed

drewjohnson2:
It certainly is wise to be wary of macros.

drewjohnson2:
@MeirGavish I think I've addressed your concerns. Let me know if you have anymore feedback.

MeirGavish:
Looks good, merging

PR #130, opened 2025-09-18
misc: add docs to build the game with Apple Container

rnjohn:
This PR aims to add:
- Documentation to build the game using Apple Container

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 services:
   balatro:
-    image: devkitpro/devkitarm:latest
-    working_dir: /balatro
+    build:
+      dockerfile: ./Dockerfile
     volumes:
       - ./:/balatro
     user: "${UID-1000}:${GID-1000}"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM devkitpro/devkitarm:latest
+WORKDIR /balatro
+
+CMD ["sh", "-c", "make -j$(nproc)"]
--- a/README.md
+++ b/README.md
@@ -50,6 +50,17 @@ A docker compose file is provided to build this project.
 
 Docker will build the project and the ROM will be in the same location as step 7 describes below.
 
+## **-Apple Container-**
+If for some reason you want to use Apple Container to build this project, you can also do so!
+
+1.) Install `container` from the releases in their [GitHub page](https://github.com/apple/container/releases)
+2.) Start the container system with `container system start`
+3.) Run `container build --tag balatro-builder --file Dockerfile .`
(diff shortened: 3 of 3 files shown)


rnjohn:
I have a question @ricfehr3 , in the commit where you added the docker compose file, I can see the line where you set the user with ${UID-1000}:${GID-1000}. Excuse me for my ignorance, but could you explain to me why is this set?

ricfehr3:
Hey @rnjohn ! Glad to see mac support. 

When building using docker locally, the project directory is mounted in the container. Windows doesn't care about the UID/GID values, so both use the default `1000` value (which would typically be the first non-root user created on a unix system).

On linux though, `1000` may map to another users GID/UID. So you need to set them before building. Side note, many users will probably not even need this since a large number of single user linux desktops use 1000:1000 for the one user. 

Not sure how this works on MacOS, but since it's unix based I figure you would want the same rules for linux?

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 services:
   balatro:
-    image: devkitpro/devkitarm:latest
-    working_dir: /balatro
+    build:
+      dockerfile: ./Dockerfile
```
ricfehr3, line 4: I don't know much about apple containers. So correct me if I'm wrong here, but I don't think there should be separate Dockerfile outside of this compose file.

Mainly I say that because the new Dockerfile doesn't add any functionality to the compose file and just moves the declaration of the `image`/`workdir`/`cmd` variables to itself. It seems like a messy solution to target a niche audience of Apple Container users. 

I suggest just sticking to regular Docker compose for now, unless you can avoid modifying the compose file and adding a new Dockerfile. 

I'm all ears though, I have been and will continually be wrong in many things, haha.

rnjohn: I'm sorry for looking at this so late. 

I mainly added that Dockerfile to use it with `container` since it doesn't support docker-compose.yml files. And I modified the docker-compose.yml file to use the Dockerfile because as you mentioned it only moved the declaration of the `image`, `workdir` and `cmd`, but it can also be left as-is.

ricfehr3: It's good work! I'm not a maintainer, so I don't get the final say. If it works and they like it, sweet!

My piece is, apple containers are suited more for capturing apple specific environments (containers with MacOS dev environments natively running on their Arm silicon, for example), so maybe we shouldn't introduce it as it has to add that dockerfile and it's use case doesn't fit here. 

However, nitpicking has it's limits! I'm not a mac dev, so maybe apple containers are badass and I'm just a grump.

rnjohn: Ok, I'll close this PR in the meantime then, until any maintainer says otherwise 👍🏼

PR #132, opened 2025-09-20
🔴WIP: DO NOT MERGE:🔴 Feature/replace malloc with pools

ricfehr3:
Making a PR here to share while I work on this. 

The interface is meant to be simple, split into 4 macros:

```c
DECLARE_POOL_TYPE(type)
DEFINE_POOL_TYPE(type, capacity)
POOL_GET()
POOL_FREE()
```

Each pool can hold a maximum of 128 objects, I'm calling the '**slots**'. This is can be adjusted, but over 128 seems to be overkill. Each slot is tracked as a flag in a bitmap, which is an array of 4 32 bit words. 

Using bitwise operations and some compiler specific calls navigating the pool's bitmap tracker is very fast. (sorry, only works on GCC now, can you even use another compiler? haha). 

Things to do:
- [ ] Consistent naming for macros. 'POOL_' should prefix.
- [ ] Ideally, I think it would be nice to describe all the memory pools in one file to give more visibility.
- [ ] Fix whitespaces/unnecessary/accidental changes


include/pool.h
```diff
@@ -0,0 +1,54 @@
+#ifndef POOL_H
+#define POOL_H
+
+#include <tonc.h>
+
+#define POOL_BITS_PER_WORD 32
+#define POOL_WORD_T u32
```
drewjohnson2: Curious about the inclusion of this `define`.

ricfehr3: I should remove it. It was to be able to set different sizes, but I didn't even use it correctly below. The GBA will always have 32-bit words, so why bother. Thanks!

source/joker.c
```diff
@@ -11,11 +11,21 @@
...
+#include "pool.h"
+
 #define JOKER_SCORE_TEXT_Y 48
 #define NUM_JOKERS_PER_SPRITESHEET 2
 
-const unsigned int **joker_gfxTiles;
-const unsigned short **joker_gfxPal;
+// Jokers in deck
+DECLARE_POOL_TYPE(Joker)
+DEFINE_POOL_TYPE(Joker, 8); // 8 should be fine, we only have a max of 2
```
ricfehr3: comment is incorrect, fix it.

source/pool.c
```diff
@@ -0,0 +1,30 @@
+#include "pool.h"
+
+void pool_clear_idx(PoolBitmap *bm, int idx) {
+    u32 i = idx / POOL_BITS_PER_WORD; // word offset
+    u32 b = idx % POOL_BITS_PER_WORD; // bit offset
+    bm->w[i] &= ~((POOL_WORD_T)1 << b);
+}
+
+int pool_get_free_idx(PoolBitmap *bm) {
```
drewjohnson2: This is really clean.

source/pool.c
```diff
@@ -0,0 +1,30 @@
+#include "pool.h"
+
+void pool_clear_idx(PoolBitmap *bm, int idx) {
```
drewjohnson2: This is probably really nitpicky, and I know this is WIP so you may have not gotten around to it yet, but since we're dividing by the word size we could replace the division and modulo operations with

`u32 i = idx >> 5`

and

`u32 b = idx & (32 - 1)`

I think that would save several cycles. I don't know how much it matters, though.

You'll probably want to double check me on that though, I'm pretty tired.

ricfehr3: Oh that's very true, good catch! Yes, definitely should just bit shift. As for finding the byte offset... I'm curious that trick is indeed faster in terms of cycles than a modulo operation. 

Best to test it :) 

Here's my test functions:
```c
void bar_div(int idx) {
    volatile u32 i = idx / POOL_BITS_PER_WORD; // word offset
    (void)i;
}

void bar_bitshift(int idx) {
    volatile u32 i = idx << 5;
    (void)i;
}

void foo_modulo(int idx) {
    volatile u32 b = idx % POOL_BITS_PER_WORD; // bit offset
    (void)b;
}

void foo_bit_and(int idx) {
    volatile u32 b = idx & 31; // bit offset
    (void)b;
}
```

Then disassemble with:

```sh
/opt/devkitpro/devkitARM/bin/arm-none-eabi-objdump -d -C --no-show-raw-insn  pool.o
```

and get output:

```
00000020 <bar_div>:
  20:   movs    r3, #31
  22:   asrs    r2, r0, #31
  24:   ands    r3, r2
  26:   adds    r0, r3, r0
  28:   sub     sp, #8
  2a:   asrs    r1, r0, #5
  2c:   str     r1, [sp, #4]
  2e:   ldr     r3, [sp, #4]
  30:   add     sp, #8
  32:   bx      lr

00000034 <bar_bitshift>:
  34:   sub     sp, #8
  36:   lsls    r0, r0, #5
  38:   str     r0, [sp, #4]
  3a:   ldr     r3, [sp, #4]
  3c:   add     sp, #8
  3e:   bx      lr

00000040 <foo_modulo>:
  40:   asrs    r3, r0, #31
  42:   lsrs    r1, r3, #27
  44:   adds    r2, r0, r1
  46:   movs    r0, #31
  48:   ands    r0, r2
  4a:   sub     sp, #8
  4c:   subs    r3, r0, r1
  4e:   str     r3, [sp, #4]
  50:   ldr     r3, [sp, #4]
  52:   add     sp, #8
  54:   bx      lr
  56:   nop                     @ (mov r8, r8)

00000058 <foo_bit_and>:
  58:   movs    r3, #31
  5a:   sub     sp, #8
  5c:   ands    r3, r0
  5e:   str     r3, [sp, #4]
  60:   ldr     r3, [sp, #4]
  62:   add     sp, #8
  64:   bx      lr
  66:   nop                     @ (mov r8, r8)
```

Before digging too deep, there's obvious cycles saved for both cases since there is simply less assembly and these calls can all assumed to be 1-2 cycles since they are basic load/math operations. 

I didn't know about that modulo trick. That's pretty neat.

I'll update it, thanks!

drewjohnson2: My life would be a lot easier if I had the sense to use the tools made available to me the way you do.

Nice test!

ricfehr3: Just fixed it. And thanks!

scripts/get_memory_map.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
```
ricfehr3, line 1: @drewjohnson2 You may find this interesting. The output looks like this right now:

```sh
Object          | location   | pool size  | func size  | bitmap size
Sprite          | 0x03002ba8 | 2048       | 96         | 16
SpriteObject    | 0x03002828 | 896        | 104        | 16
Joker           | 0x030026a8 | 40         | 108        | 16
JokerObject     | 0x030026e0 | 64         | 40         | 16
Card            | 0x03002730 | 104        | 40         | 16
CardObject      | 0x030027a8 | 128        | 40         | 16
Total bytes used: 3804
```

ricfehr3: The locations also are the same in gdb via mgba:

```sh
[gbagdb]>p &Sprite_storage
$5 = (Sprite (*)[128]) 0x3002ba8 <Sprite_storage>
[gbagdb]>p &SpriteObject_storage
$6 = (SpriteObject (*)[16]) 0x3002828 <SpriteObject_storage>
[gbagdb]>p &Joker_storage
$7 = (Joker (*)[8]) 0x30026a8 <Joker_storage>
[gbagdb]>p &JokerObject_storage
$8 = (JokerObject (*)[8]) 0x30026e0 <JokerObject_storage>
[gbagdb]>p &Card_storage
$9 = (Card (*)[52]) 0x3002730 <Card_storage>
[gbagdb]>p &CardObject_storage
$10 = (CardObject (*)[16]) 0x30027a8 <CardObject_storage>
```

source/joker.c
```diff
@@ -11,11 +11,13 @@
...
 #include <string.h>
 
+#include "pool.h"
+
 #define JOKER_SCORE_TEXT_Y 48
 #define NUM_JOKERS_PER_SPRITESHEET 2
 
-const unsigned int **joker_gfxTiles;
-const unsigned short **joker_gfxPal;
+static const unsigned int* joker_gfxTiles[15];
```
ricfehr3: Fix magic numbers

ricfehr3: D'oh, I remember, this is fixed here: https://github.com/cellos51/balatro-gba/pull/123

include/card.h
```diff
@@ -69,4 +72,4 @@ void card_object_set_selected(CardObject* card_object, bool selected);
 bool card_object_is_selected(CardObject* card_object);
 Sprite* card_object_get_sprite(CardObject* card_object);
 
-#endif // CARD_H
\ No newline at end of file
+#endif // CARD_H
```
ricfehr3: remove this chunk

scripts/get_memory_map.sh
```diff
@@ -0,0 +1,48 @@
...
+
+POOL_DEF_FILE='./include/pools.def'
+ELF_FILE='./build/balatro.elf'
```
ricfehr3: change to an optional argument.

source/pool.c
```diff
@@ -0,0 +1,43 @@
...
+    bm->w[i] &= ~((u32)1 << b);
+}
+
+int pool_bm_get_free_idx(PoolBitmap *bm)
+{
+    for (u32 i = 0; i < bm->nwords; i++)
+    {
+        u32 inv = ~bm->w[i];
+
+        // guard so we don't cal `ctz` with 0, since __builtin_ctz(0) is undefined
```
ricfehr3: typo

scripts/get_memory_map.sh
```diff
@@ -0,0 +1,48 @@
...
+READELF='/opt/devkitpro/devkitARM/bin/arm-none-eabi-readelf'
+TOTAL_BYTES=0
+
+get_pool_names() {
```
ricfehr3, line 8: Remove, not used anymore

source/pool.c
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    return -1;
+}
+
+
+#define POOL_ENTRY(name, capacity) \
+POOL_DEFINE_TYPE(name, capacity);
+#include "sprite.h"
```
ricfehr3: remove

source/joker.c
```diff
@@ -38,7 +40,7 @@ static bool used_layers[MAX_JOKER_OBJECTS] = {false}; // Track used layers for j
 
 // Maps the spritesheet index to the palette bank index allocated to it.
 // Spritesheets that were not allocated are
-static int* joker_spritesheet_pb_map;
+static int joker_spritesheet_pb_map[128];
```
ricfehr3: fix this magic?

include/pool.h
```diff
@@ -0,0 +1,61 @@
+#ifndef POOL_H
+#define POOL_H
+
+#include <tonc.h>
+
+#define POOL_BITS_PER_WORD 32
+#define POOL_MAX_CAPACITY 128
```
ricfehr3: Remove? It's unused but it's a good note. I don't think it gets compiled.

source/pool.c
```diff
@@ -0,0 +1,43 @@
+#include "pool.h"
+
+void pool_bm_clear_idx(PoolBitmap *bm, int idx)
+{
+    // Divide by 32 to get the word index
+    u32 i = idx >> 5;
+    // Get last 5-bits, same as a modulo (% 32) operation on positive numbers
+    u32 b = idx & 31;
```
ricfehr3: Change to `0x1F` for clarity

scripts/get_memory_map.sh
drewjohnson2: You were 100% right, this is very interesting to me.

I like what's going on here a lot.

ricfehr3: After looking at this 100 times, I realize "location" is a terrible name for "address" haha.

ricfehr3: Fixed

ricfehr3:
@MeirGavish I'm happy with this one now. It's a big one so no rush. 

Also, there's a change that is also in this PR to have it work in the CI: https://github.com/cellos51/balatro-gba/pull/143

ricfehr3:
@MeirGavish I have to say, copilot kinda rocks for PRs. Never used it before. I'll fix what it found.

MeirGavish:
It does, though it also does sometimes make mistakes or suggests unnecessary changes.

source/pool.c
```diff
@@ -0,0 +1,42 @@
...
+        // where the first free slot is. This operation prevents looping through every bit of filled flags, and
+        // will instead operate only on the first word with free slots.
+        if (inv)
+        {
+            int bit = __builtin_ctz(inv);
+            bm->w[i] |= ((u32)1 << bit);
+            int idx = i * POOL_BITS_PER_WORD + bit;
+            return idx;
+        }
+    }
```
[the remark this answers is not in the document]
ricfehr3: This is a fair point. I think the easier solution may be to just put the capacity of the pool in the pool bitmap struct. Then, change the return to:

`return (idx < bm->cap) ? idx : -1`

source/sprite.c
```diff
@@ -2,39 +2,28 @@
 #include "util.h"
 #include "audio_utils.h"
 #include "soundbank.h"
+#include "pool.h"
```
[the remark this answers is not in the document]
ricfehr3: Ok computer

source/sprite.c
```diff
@@ -2,39 +2,28 @@
...
 static bool free_affines[MAX_AFFINES] = {false};
 
 // Sprite methods
 Sprite *sprite_new(u16 a0, u16 a1, u32 tid, u32 pb, int sprite_index)
 {
-    Sprite *sprite = malloc(sizeof(Sprite));
+    Sprite* sprite = POOL_GET(Sprite);
 
     sprite->obj = NULL;
     sprite->aff = NULL;
```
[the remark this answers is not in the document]
ricfehr3: True, need to re-add the tracker.

source/sprite.c
```diff
@@ -51,7 +40,6 @@ Sprite *sprite_new(u16 a0, u16 a1, u32 tid, u32 pb, int sprite_index)
 
         if (aff_index == MAX_AFFINES)
         {
```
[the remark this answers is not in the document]
ricfehr3: true, need to fix

scripts/get_memory_map.sh
```diff
@@ -0,0 +1,57 @@
...
+            grep -E "${name}$"                     | \
+            sed -E 's@ +@ @g; s@^ @@'              | \
+            tr -d '\n'                               \
+        )"
+
+
+    address="$(cut -d ' ' -f 2 <<< $output_pool)"
+    pool_size="$(cut -d ' ' -f 3 <<< $output_pool)"
+    func_size="$(cut -d ' ' -f 3 <<< $output_func)"
+    bm_size=16 #always gonna be 16, 4 * sizeof(u32)
```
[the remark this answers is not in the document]
ricfehr3: This does keep it consistent with how the other values are found. I'll add it this way, but fit it better to the style above.

scripts/get_memory_map.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+POOL_DEF_FILE='./include/pools.def'
+ELF_FILE="${1-./build/balatro-gba.elf}"
+READELF='/opt/devkitpro/devkitARM/bin/arm-none-eabi-readelf'
```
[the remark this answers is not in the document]
ricfehr3: I'll change the variable to default to the specified path, but let it be overridden if someone went a lil custom with their install.

source/sprite.c
```diff
@@ -2,39 +2,40 @@
...
 
 OBJ_ATTR obj_buffer[MAX_SPRITES];
 OBJ_AFFINE *obj_aff_buffer = (OBJ_AFFINE*)obj_buffer;
 
-static Sprite *free_sprites[MAX_SPRITES] = {NULL};
+typedef struct
+{
+    bool used;
+    Sprite* p_sprite;
+} _UsedSprite;
```
ricfehr3: So, this is a bit of an interesting change. I did need to readd this tracker, but mapping it directly to the array of sprite structs just isn't possible with the pool anymore. So, at the cost of space, there's this tracker.

I put a lot of thought into this, I may add more to this comment later.

But the TLDR;

- I don't want this to be a forced change to work with the pools, but I think there's value to this:
- We have control over the number of entries in the pool, as it's now decoupled from tracking it's OAM memory offset.
- Adding the OAM memory offset to the sprite struct is appropriate since it is strongly typed to the sprite itself.

Scrutiny is allowed.

source/pool.c
MeirGavish: @ricfehr3 
@drewjohnson2 

I do think the compiler optimizes this in some way.
I noticed that in your example you used a **shift left** while a division is a **shift right**

```
void bar_div(int idx) {
    volatile u32 i = idx / POOL_BITS_PER_WORD; // word offset
    (void)i;
}

void bar_bitshift(int idx) {
    volatile u32 i = idx << 5;
    (void)i;
}
```

I didn't get into the assembly code and I don't yet understand why the modulo and division have more instructions but the division does use the shift operator
`  2a:   asrs    r1, r0, #5`

and for the modulo I think this is the core operation
`  48:   ands    r0, r2`

and they do avoid calling a division subroutine that would probably be needed for a non-power of 2 (as there's no division hardware unit) and would be costly.

ricfehr3: I didn't notice I did the wrong shift direction. Good catch, that was my bad. I honestly just saw that there were less instructions with the bitshift and kept moving along. However, from our conversation with @pansythoughts in #145 , I've been learning a lot about this division business, haha. 

This video explains a method used for division with the recipricol on arm hardware that has no division unit: https://www.youtube.com/watch?v=ssDBqQ5f5_0 . What we see in this disassembly is different because the compiler can do a (to your point) a better optimized version of division since it is divided by a power of 2.

Taking it step-by-step: 

- The assembly for `x / 32` looks like this in C
**`(x + ((x >> 31) & 31)) >> 5`**
- That last bitshift `>> 5` is self-explanatory at this point, a rudimentary division by 32 with bitshifts. Let's remove it and simplify the syntax:

**`x + ((x >> 31) & 0x1F)`**

- Great, so let's work from inner to outer parenthesis to figure out what we are bitshifting by 5, exactly:

- `x >> 31` : This makes a `signed mask`. Very importantly, it's operation uses [`asrs`](https://developer.arm.com/documentation/ddi0597/2025-09/Base-Instructions/ASRS--immediate---Arithmetic-Shift-Right--setting-flags--immediate---an-alias-of-MOV--MOVS--register--). Which is a bitshift that preserves the sign bit as it shifts. For us, that means that any **signed** value looks like `0xFFFF` and any **unsigned** value looks like `0x0000` after this operation.

- Then, taking that signed mask value, we end up with 0x000 for unsigned values, and 0x1f for signed values. Lets just write that down for clarity:

**`x + ((x >> 31) & 0x1F)`** =
    Signed: `x + 0x1f`
    Unsigned: `x`
    
- Plug that back into the original equation:
**`x + ((x >> 31) & 0x1F) >> 5`** =
    Signed: `(x + 0x1F) >> 5`
    Unsigned: `x >> 5`
    
- Now, that should look familiar. `x >> 5` is what I end up using here, but what's with `(x + 0x1F) >> 5` for signed numbers? That is for truncation toward 0. A simple example is with `x = -33`.

Expectation (what c expects): -33 / 32 = -1 :heavy_check_mark: 
Bad (just bitshift no offset): -33 >> 5 = -2 :heavy_multiplication_x: 
Good (add the offset of 31): (-33 + 31) >> 5 = (-2) >> 5 = -1 :heavy_check_mark: 

So, since we know there will not be signed numbers, `x >> 5` is the cleanest form. 

Also holy shit division is some crazy stuff. Hats off to the giants whose shoulders I get to stand on who figured all that out.

MeirGavish: @ricfehr3
Thanks for the detailed explanation! So the difference in the number of instructions is essentially down to signed value handling, I wonder if it would be the same if we passed an unsigned int instead of int though I do prefer to avoid unsigned variables since 0 is such a common value and you're just a -1 away from overflow.
Anyway, I think I still prefer using the division and modulo operations since you use the constant POOL_BITS_PER_WORD rather than magic numbers making it more readable and maintainable. To me it doesn't seem like shaving off a few instructions is worth it here.

ricfehr3:
@MeirGavish @drewjohnson2 

Well, I think this one is good officially now. I couldn't rest easy without putting some type of test in there. Take a look at the new `tests` directory. I added a test to fill, empty, fill, empty, fill, empty partially, refill, empty, fill, and empty again. The test was added to the CI and it is setup to welcome other tests via the script `run_tests.sh` (this script is now ran in the CI too). 

I found a bug with the number of bytes for the bitmap array. Turns out I was using `sizeof(uint32_t)` which didn't make sense. I upped that value to 8, to allow for 256 maximum entries. Which may be needed in the future. Not a big deal to add it now, all pool capacities can be limited anyway.

Anyway, I'm happy with this. I won't merge my own request, it's bad luck. But, it's all rebased.

ricfehr3:
@drewjohnson2 I added some timing metrics to the test scripts. It looks like though a single call to malloc vs pool is slower, many subsequent calls to POOL_GET is faster. However, this data is moot. On these modern machines it's likely caching that storage array and shredding the time. I would say that the data gives a heuristic that it is acceptable (though slower) in terms of it's speed in comparison to malloc. 

That being said, I could definitely see the pool shining without a doubt on machines that have a cache.  

```sh
Run cd tests && ./run_tests.sh
  cd tests && ./run_tests.sh
  shell: /usr/bin/bash -e {0}
  env:
    UID: 1001
    GID: 1001
rm -f build/pool_test
mkdir -p build
gcc -I../../include -I. -g -O3 -Wall -Werror -DPOOLS_DEF_FILE=\"def_test_mempool.h\"  -o build/pool_test pool_test.c ../../source/pool.c 
Testing Pool Fill and Empty 1x.
Testing Pool Fill and Empty 2x.
Testing Pool Fill, Partial Empty, Refill, Empty.
Testing Pool Fill and Empty.
---------------------------------------------------------
Pool Tests Passed
---------------------------------------------------------
Testing execution time for fun :)

Pool Get One:
	Elapsed: 80 ns

Pool Free One:
	Elapsed: 30 ns

Malloc One:
	Elapsed: 30 ns

Free One:
	Elapsed: 29 ns

Fill and Empty Pool 240 times:
	Elapsed: 1583 ns

Fill and Empty Malloc 240 times:
	Elapsed: 8016 ns
```

.github/workflows/build_ci_workflow.yml
```diff
@@ -53,8 +53,11 @@ jobs:
           echo "UID=$(id -u)" >> "$GITHUB_ENV"
           echo "GID=$(id -g)" >> "$GITHUB_ENV"
 
+      - name: Run Tests
+        run: cd tests && ./run_tests.sh
+
       - name: Build the project
-        run: docker compose -f docker-compose.yml run --rm balatro sh -c "make -j$(nproc)"
+        run: docker compose -f docker-compose.yml run --rm balatro sh -c "make -j$(nproc) && ./scripts/get_memory_map.sh"
```
MeirGavish, line 60: Why include the get_memory_maps.sh script in the workflow? How does it affect the workflow?

Makefile
```diff
@@ -25,13 +25,14 @@ LIBTONC := $(DEVKITPRO)/libtonc
...
-MUSIC		:= audio
-GRAPHICS	:= graphics
+TARGET         := $(notdir $(CURDIR))
+BUILD          := build
+SOURCES	       := source
+INCLUDES       := include
+DATA           :=
+MUSIC          := audio
+GRAPHICS       := graphics
+POOLS_DEF_FILE := \"def_balatro_mempool.h\"
```
MeirGavish, line 35: Why make this an external define?

include/card.h
```diff
@@ -6,6 +6,9 @@
 
 #include "sprite.h"
 
+#define MAX_CARDS (NUM_SUITS * NUM_RANKS)
```
MeirGavish, line 9: I guess this is okay for now but will need to be increased if/when Tarot/Joker cards are implemented that allow the deck to grow past 52 cards.

include/joker.h
```diff
@@ -8,6 +8,10 @@
 #include "game.h"
 #include "graphic_utils.h"
 
+// This won't be more than the number of jokers in your current deck
+// plus the amount that can fit in the shop, 8 should be fine. For now...
+#define MAX_ACTIVE_JOKERS 8
```
MeirGavish, line 13: Also okay for now, once booster packs are implemented, this will need to be at least 9 to account for up 4 jokers from the pack. It doesn't have to be a power of 2, right?

source/sprite.c
```diff
@@ -3,39 +3,40 @@
...
+static _UsedSprite free_sprites[MAX_SPRITES] = {{0}};
 
 // Sprite methods
 Sprite *sprite_new(u16 a0, u16 a1, u32 tid, u32 pb, int sprite_index)
 {
-    Sprite *sprite = malloc(sizeof(Sprite));
+    if(free_sprites[sprite_index].used)
+    {
+        return free_sprites[sprite_index].p_sprite;
+    }
```
MeirGavish: Previously it returned NULL when an allocation of a sprite in a used index. I found one NULL-check that this might affect in `sprite_object_set_sprite()`. I think it might be better to preserve that behavior to avoid issues, though maybe we should add NULL-checks to a few other uses of `sprite_new()`

source/sprite.c
MeirGavish: I'm okay with this, but why is the previous tracker not possible with the pools now?

source/pool.c
```diff
@@ -0,0 +1,42 @@
+#include "pool.h"
+
+void pool_bm_clear_idx(PoolBitmap *bm, int idx)
+{
+    // Divide by 32 to get the word index
+    uint32_t i = idx >> 5;
+    // Get last 5-bits, same as a modulo (% 32) operation on positive numbers
+    uint32_t b = idx & 0x1F;
+    bm->w[i] &= ~((uint32_t)1 << b);
```
MeirGavish: Just copying my reply to [a previous thread](https://github.com/cellos51/balatro-gba/pull/132#discussion_r2365966071) to make sure you see (with an addition after some more thought),

Thanks for the detailed explanation! So the difference in the number of instructions is essentially down to signed value handling, I wonder if it would be the same if we passed an unsigned int instead of int for the index though in general I do prefer to avoid unsigned variables since 0 is such a common value and you're just a -1 away from overflow. Also, I think passing an unsigned variable here would generate a warning because of the pointer subtraction passed in the call.
Anyway, I think I still prefer using the division and modulo operations since you use the constant POOL_BITS_PER_WORD rather than magic numbers making it more readable and maintainable. To me it doesn't seem like shaving off a few instructions is worth it here.

include/pool.h
```diff
@@ -0,0 +1,62 @@
...
+#include <stdint.h>
+
+#define POOL_BITS_PER_WORD 32
+#define POOL_BITMAP_BYTES   8
+
+typedef struct PoolBitmap {
+    uint32_t *w;
+    uint32_t nbits;
+    uint32_t nwords;
+    uint32_t cap;
```
MeirGavish, line 13: Why did you switch tonc's u32 for uint32_t? Much of the existing code uses those types.

source/pool.c
```diff
@@ -0,0 +1,42 @@
...
+        if (inv)
+        {
+            int bit = __builtin_ctz(inv);
+            bm->w[i] |= ((uint32_t)1 << bit);
+            int idx = i * POOL_BITS_PER_WORD + bit;
+            return (idx < bm->cap) ? idx : -1;
+        }
+    }
+
+    return -1;
```
MeirGavish: I think using UNDEFINED here is cleaner than the explicit -1

source/joker.c
```diff
@@ -48,7 +50,7 @@ static bool used_layers[MAX_JOKER_OBJECTS] = {false}; // Track used layers for j
 
 // Maps the spritesheet index to the palette bank index allocated to it.
 // Spritesheets that were not allocated are
-static int* joker_spritesheet_pb_map;
+static int joker_spritesheet_pb_map[MAX_SPRITES];
```
```suggestion
static int joker_spritesheet_pb_map[(MAX_DEFINABLE_JOKERS + 1)/NUM_JOKERS_PER_SPRITESHEET];
```

The size of this array is not really tied to the number of sprites that can be displayed, theoretically if we implement all 150 jokers and decide to decrease `NUM_JOKERS_PER_SPRITESHEET` to 1, it would be larger. I suggest defining MAX_DEFINABLE_JOKERS as 150 to get a better cap for the size of this array and avoid the malloc.

MeirGavish:
Another thing I want to note, is that in the main branch there is currently a sprite tear/flickering issue when drawing cards introduced in the dynamic sprite swapping in VRAM PR #53 (mentioned in the discussion there). It seems better now but I still see it occasionally, see images below.
I suspect that this is because allocating a new card sprite takes too long and the sprite VRAM ends up being updated during VDRAW so this PR might fix it by optimizing that.

sprite_flicker_tear_1
sprite_flicker_tear_2

.github/workflows/build_ci_workflow.yml
ricfehr3: There is no reason. Initially I thought maybe it would be useful somewhere, but I can remove it.

include/card.h
ricfehr3: Should be no problem to add or remove here.

include/joker.h
ricfehr3: It does not. Right now every POOL has a 256 max capacity, limited by the 8 x 32 bits from the bitmap tracker. Any number in-between will work.

include/pool.h
ricfehr3: I'm going to respond to this related comment here too:
https://github.com/cellos51/balatro-gba/pull/132#discussion_r2443270548

Summary:

**These changes allow for testing the memory pool on host machines.** As such they need to be independent from _libtonc_ and _devkitpro_ entirely, and instead use the host toolchain. I wouldn't normally try to justify these (rightfully called out) inconsistencies. However, here I think allowing the code to be testable on the host is **extremely** important for any type of memory subsystem, here included.


**Why did you switch tonc's u32 for uint32_t? Much of the existing code uses those types.** 
- This is to remove the `tonc.h` dependency. Switching it to `uint32_t` is chosen because. 1. It reads better than "unsigned int" to signify it is 32 bits in size. 2. It is the correct type that `tonc`'s u32 ends up being: I had a related issue with a tonc update back in May. You can check this ticket out, but basically `u32` == `uint32_t` ->https://github.com/devkitPro/grit/pull/20#issue-3088805745 . 

**Why make this an external define?**
- This let's different pools be used, again allowing testing on the host. I use the same pool implementation but with a different definition file for testing. You can see there's another Makefile for the host tests with a simpler pool:
https://github.com/ricfehr3/balatro-gba/blob/e40f3c989f4c00199ff1e00ac794bd365d482bc9/tests/pool/Makefile#L1

Makefile
ricfehr3: answered here->https://github.com/cellos51/balatro-gba/pull/132#discussion_r2443306380

source/sprite.c
ricfehr3: I'll fix this

source/pool.c
ricfehr3: I see where you're coming from. I agree, it shouldn't effect the performance too much. What I'll do is I'll leave it as `div`/`modulo`, but I'll leave the faster version and a link to the discussion if it is ever needed.

source/pool.c
ricfehr3: I agree, I think I have `return -1` other places too, I'll hunt them down and fix it.

source/joker.c
ricfehr3: Absolutely. Great call.

ricfehr3 (commented):
Thanks for reviewing this @MeirGavish ! I'm going to apply some suggestions, and I added some explanations for some of the odd choices, definitely room for discussion.

I also wanted to reply to some comments that github won't let me respond to for some reason.

https://github.com/cellos51/balatro-gba/pull/132#discussion_r2443285808

**I'm okay with this, but why is the previous tracker not possible with the pools now?**

I think I got lost in the sauce there. Looking at the diff, I'm not sure why I added that extra variable to do what a NULL does. I'm going to revert that and see. The old way was cleaner too, imo.

https://github.com/cellos51/balatro-gba/pull/132#issuecomment-3419702337

**Sprite tearing on VRAM**

I agree, I think this also caused by the sprite updating on a VSYNC. VSYNC interrupts at 60 FPS, and it does seem like we are taking too long on some frames. Glad this PR is helping with that :) . I have some more ideas on how we can optimize further. I plan to make issues for them.

ricfehr3:
@MeirGavish Okay, updated again to cleanup for the PR besides any of the discussion points for testing.

source/pool.c
```diff
@@ -0,0 +1,49 @@
...
+    // Divide by 32 to get the word index
+    uint32_t i = idx / 32;
+    // Get last 5-bits, same as a modulo (% 32) operation on positive numbers
+    uint32_t b = idx % 32;
+
+    // Below are the "fast" forms of the above operations, respectively.
+    // These are more efficient, but removed for readability
+    // See: https://github.com/cellos51/balatro-gba/pull/132#discussion_r2365966071
+    //uint32_t i = idx >> 5;
+    //uint32_t b = idx & 0x1F;
```
MeirGavish:
```suggestion
    uint32_t i = idx / POOL_BITS_PER_WORD;
    uint32_t b = idx % POOL_BITS_PER_WORD;

    // Below are the "fast" forms of the above operations, respectively.
    // These are more efficient, but removed for readability
    // See: https://github.com/cellos51/balatro-gba/pull/132#discussion_r2365966071
    // Divide by 32 to get the word index
    //uint32_t i = idx >> 5;
    // Get last 5-bits, same as a modulo (% 32) operation on positive numbers
    //uint32_t b = idx & 0x1F;
```

Just moved the comments and made sure you use the define

include/pool.h
MeirGavish: Alright, makes sense, thanks! I do prefer uint32_t to unsigned int/unsigned long int where the bit width is not explicit.

include/card.h
MeirGavish: Thanks!

include/joker.h
MeirGavish: Thanks!

source/pool.c
MeirGavish: Okay, nice, thanks, just see my [new comment](https://github.com/cellos51/balatro-gba/pull/132#discussion_r2444661529).

.github/workflows/build_ci_workflow.yml
MeirGavish: We can leave it in, I guess it can be useful if a PR adds a new pool or increases the size of an existing one, then we can quickly see the new sizes of the pools.

source/pool.c
ricfehr3: Oh snap, my bad. You only told me twice :joy: Sorry, I went ahead and applied that.

.github/workflows/build_ci_workflow.yml
```diff
@@ -53,8 +53,11 @@ jobs:
           echo "UID=$(id -u)" >> "$GITHUB_ENV"
           echo "GID=$(id -g)" >> "$GITHUB_ENV"
 
+      - name: Run Tests
+        run: cd tests && ./run_tests.sh
```
MeirGavish, line 57: @ricfehr3 I missed this so commenting on a closed PR, maybe it's better to extract this test to another workflow with the same triggers? It seems quite independent since it's building a different target and I'm not sure it's a good idea to lump all out tests in a single workflow.

ricfehr3: @MeirGavish I agree, let me make an issue for it so I don't forget to do it.

PR #133, opened 2025-09-20
Cleanup, Refactor blinds into a map

ricfehr3:
Refactor the switch-cases out of `blind.c` and replace with a map to the properties of each type of blind. 

This aims to limit code reuse and centralize data for the blind system.


include/blind.h
```diff
@@ -39,12 +37,21 @@ enum BlindState
     BLIND_SKIPPED
 };
 
+typedef struct Blind {
+    enum BlindType type;
+    const u16* palette;
+    FIXED multiplier;
+    s32 reward;
+    u32 tid;
+    u32 pb;
```
ricfehr3, line 58: Thinking it may make sense to move `palette`,`tid`, and `pb` into their own struct for `palette_info` or something

include/blind.h
```diff
@@ -39,12 +38,27 @@ enum BlindState
...
+// TODO: Move this to a common interface for other palettes
+typedef struct {
+    const u16* palette;
+    u32 tid;
+    u32 pb;
+} PaletteInfo;
+
+typedef struct
+{
+    enum BlindType type;
```
ricfehr3, line 63: Having the type here isn't necessary. I find it to keep things neat when declaring though.

include/blind.h
```diff
@@ -31,6 +22,14 @@ static const int ante_lut[MAX_ANTE + 1] = {100, 300, 800, 2000, 5000, 11000, 200
 #define BLIND_BACKGROUND_SECONDARY_COLOR_INDEX 6
 #define BLIND_BACKGROUND_SHADOW_COLOR_INDEX 7
 
+enum BlindType
+{
+    SMALL_BLIND,
```
ricfehr3: May want to reverse this? Typically its `<NAME>_<TYPE>` with enums so `BLIND_SMALL` ... I think that's correct, I'll do that.

source/game.c
```diff
@@ -549,18 +558,22 @@ void change_background(int id)
...
         // Disable the button highlight colors
         // Select button PID is 15 and the outline is 18
         memcpy16(&pal_bg_mem[18], &pal_bg_mem[15], 1);
         // Skip button PID is 10 and the outline is 5
         memcpy16(&pal_bg_mem[10], &pal_bg_mem[5], 1);
 
-        for (int i = 0; i < MAX_BLINDS; i++)
+        // TODO: This should have it's data mapped with all these x,y constants
+        // and it should operate in a loop without the if/else blocks on those
+        // constants associated with each blind type
```
ricfehr3: Can I make an issue for this? @MeirGavish , if you like the idea?

ricfehr3:
:stop_sign: Do not merge yet.

@drewjohnson2 From talking today I've decided that here would be a good place to use X-macros for the declaration of blinds. What do you think?

I think it needs to be extensible because it looks like there's going to be a lot more blinds pretty soon: https://github.com/cellos51/balatro-gba/discussions/131

drewjohnson2:
That makes perfect sense to me.

source/blind.c
```diff
@@ -4,90 +4,94 @@
...
+#define BIG_BLIND_REWARD   4
+#define BOSS_BLIND_REWARD  5
+
+// +1 is added because we'll actually be indexing at 1, but if something causes you to go to ante 0, there will still be a value there.
+static const int ante_lut[MAX_ANTE + 1] = {100, 300, 800, 2000, 5000, 11000, 20000, 35000, 50000};
+
 // Palettes for the blinds (Transparency, Text Color, Shadow, Highlight, Main Color) Use this: http://www.budmelvin.com/dev/15bitconverter.html
 static const u16 small_blind_token_palette[PAL_ROW_LEN] = {0x0000, 0x7FFF, 0x34A1, 0x5DCB, 0x5104, 0x55A0, 0x2D01, 0x34E0};
 static const u16 big_blind_token_palette[PAL_ROW_LEN] = {0x0000, 0x2527, 0x15F5, 0x36FC, 0x1E9C, 0x01B4, 0x0D0A, 0x010E};
 static const u16 boss_blind_token_palette[PAL_ROW_LEN] = {0x0000, 0x2CC9, 0x3D0D, 0x5E14, 0x5171, 0x4D0F, 0x2CC8, 0x3089}; // This variable is temporary, each boss blind will have its own unique palette
```
ricfehr3, line 15: @MeirGavish Big question for ya here. Going forward will we be extracting the palette using grit when generating the sprite assets for blinds? 

We currently have a greyscale image and use these hardcoded values. Given we plan to have a lot more blind tokens, It would make sense to me to leave the palette generation to grit. 

I wouldn't mind implementing those new tokens in this PR. I would assume png per token, but I'm not sure if you prefer another way. Still, would rather refactor around implementing these new blinds.

MeirGavish: @ricfehr3 
Sorry, I missed this question while I was away. 
I think it would be better to refactor the blind palettes to leave the generation to grit, I'm not very fond of hardcoding the colors.
I thought of putting the small and big blind together in the same file so they share a palette since they are not very dynamic (and we save one palette bank) and for all the other (boss) blinds to have a single file for each blind.
However, notice that there is code that relies on the hard-coded palette indices when changing colors of background elements to match the blind colors so it's not a trivial refactor.

MeirGavish:
I'm requesting a review from Copilot to help me manage the more complex PRs, and I will review them myself later on, take his review with a grain of salt, I will reply to his comments with my opinion when I get to it, but it can be helpful to spot mistakes.

include/blind.h
```diff
@@ -31,20 +22,45 @@ static const int ante_lut[MAX_ANTE + 1] = {100, 300, 800, 2000, 5000, 11000, 200
...
+    const u16* palette;
+    u32 tid;
+    u32 pb;
+} PaletteInfo;
+
+typedef struct
+{
+    enum BlindType type;
+    PaletteInfo pal_info;
+    FIXED multiplier;
```
MeirGavish: Maybe `req_multiplier` or `score_req_multiplier` would be a clearer name?

source/game.c
MeirGavish: @ricfehr3 
Did you intend to map it in the same struct? Might be a bit too much for it since it's just for this blind select screen.
I actually was going to refactor this at some point and define constants for the x,y, see the TODO to use main_bg_se_copy_rect() below

source/game.c
```diff
@@ -533,14 +540,16 @@ void change_background(int id)
...
+        }
 
         const int default_y = 89 + (TILE_SIZE * 12); // Default y position for the blind select tokens. 12 is the amound of tiles the background is shifted down by
-        sprite_position(blind_select_tokens[SMALL_BLIND], 80, default_y);
-        sprite_position(blind_select_tokens[BIG_BLIND], 120, default_y);
-        sprite_position(blind_select_tokens[BOSS_BLIND], 160, default_y);
+        // TODO refactor magic numbers '80/120/160' into a map to loop with
+        sprite_position(blind_select_tokens[BLIND_TYPE_SMALL], 80, default_y);
+        sprite_position(blind_select_tokens[BLIND_TYPE_BOSS], 120, default_y);
+        sprite_position(blind_select_tokens[BLIND_TYPE_BIG], 160, default_y);
```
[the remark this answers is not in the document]
MeirGavish: @ricfehr3 
Seems like it spotted a mistake.

include/blind.h
MeirGavish: Yeah, I think it's good.

include/blind.h
ricfehr3: I agree :+1:

source/game.c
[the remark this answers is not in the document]
ricfehr3: Nice! It sure did, what a lifesaver.

source/game.c
ricfehr3: I didn't see that comment below. I'll come back to this and make a decision after I consider what the best option is

source/blind.c
ricfehr3: After checking this out more. I see that it really is a lot more complex to fix that. It looks like there's some comments around the use of `blind_get_color` saying the approach needs to be changed anyway. 

I'm going to defer that to another PR and make an issue.

include/blind.h
ricfehr3: Will do this in another branch. I don't think it would make sense to refactor where this could be used here.

ricfehr3:
@MeirGavish I'm calling this one good for now. I did end up trying to use `main_bg_se_copy_rect` but it looks like the border edge overlaps on SE offsets, which makes it a bit more difficult to implement and I think that's outside the scope of this PR.

include/blind.h
```diff
@@ -23,28 +14,62 @@ static const int ante_lut[MAX_ANTE + 1] = {100, 300, 800, 2000, 5000, 11000, 200
...
+    BLIND_MAIN_COLOR_INDEX = 4,
+    BLIND_BACKGROUND_MAIN_COLOR_INDEX = 5,
+    BLIND_BACKGROUND_SECONDARY_COLOR_INDEX = 6,
+    BLIND_BACKGROUND_SHADOW_COLOR_INDEX = 7,
+};
+
+#define BLIND_INFO_TABLE \
+    BLIND_INFO(SMALL, small, FIX_ONE) \
+    BLIND_INFO(BIG, big, FIX_ONE) \
+    BLIND_INFO(BOSS, boss, FIX_ONE)
```
MeirGavish: Did you intend to put FIX_ONE on all the values for multi here? Looking at the usage in blindMap[] it seems all the blinds will have score_req_multiplier of 1. 
Also if you put the score requirement in the xmacro, I think the rewards should also be there rather than in separate defines. The palettes are separate so I think it's good they are defined separately.

source/game.c
```diff
@@ -78,10 +78,17 @@ static CardObject *main_menu_ace = NULL;
 static Sprite *playing_blind_token = NULL; // The sprite that displays the blind when in "GAME_PLAYING/GAME_ROUND_END" state
 static Sprite *round_end_blind_token = NULL; // The sprite that displays the blind when in "GAME_ROUND_END" state
 
-static Sprite *blind_select_tokens[MAX_BLINDS] = {NULL}; // The sprites that display the blinds when in "GAME_BLIND_SELECT" state
+static Sprite *blind_select_tokens[BLIND_TYPE_MAX] = {NULL}; // The sprites that display the blinds when in "GAME_BLIND_SELECT" state
```
MeirGavish, line 81: Here, this should stay 3 and not grow when new boss blinds are added.

source/game.c
```diff
@@ -78,10 +78,17 @@ static CardObject *main_menu_ace = NULL;
...
 
-static int current_blind = SMALL_BLIND;
-static enum BlindState blinds[MAX_BLINDS] = {BLIND_CURRENT, BLIND_UPCOMING, BLIND_UPCOMING}; // The current state of the blinds, this is used to determine what the game is doing at any given time
+static int current_blind = BLIND_TYPE_SMALL;
+
+// The current state of the blinds, this is used to determine what the game is doing at any given time
+static enum BlindState blinds[BLIND_TYPE_MAX] =
```
MeirGavish, line 86: Here, this should stay 3 and not grow when new boss blinds are added.

source/game.c
ricfehr3: Oh nice catch. Yep, this and a few others I caught are fixed. Thanks!

include/blind.h
ricfehr3: I think that's a good idea. I'll get rework that.

ricfehr3: Also, FIXED_ONE was totally a mistake, I thought I had put in the multipliers previously. My bad, update coming soon.

MeirGavish:
After giving it just a bit more thought, I think it makes sense to separate the mappings - there are 3 general blind types - small, big, and boss and values of score requirement multiplier and reward should be mapped to them as they are the same for all boss blinds [according to the wiki](https://balatrogame.fandom.com/wiki/Blinds_and_Antes) apart from some that have the requirement modified (we should allow for that flexibility but I think it shouldn't be a problem). And then the palette info should have a separate mapping so any boss blind can have its own graphics.

ricfehr3:
> After giving it just a bit more thought, I think it makes sense to separate the mappings - there are 3 general blind types - small, big, and boss and values of score requirement multiplier and reward should be mapped to them as they are the same for all boss blinds [according to the wiki](https://balatrogame.fandom.com/wiki/Blinds_and_Antes) apart from some that have the requirement modified (we should allow for that flexibility but I think it shouldn't be a problem). And then the palette info should have a separate mapping so any boss blind can have its own graphics.

I agree. I can get this done after work today. But for clarity, when you say, "separate the mappings", are you saying:

1. We have a map for small/big/boss blinds with static information. Top of my head this would be like: `x/y` coords, `tid/pid` (bosses will share the same one), score multiplier, and reward.

2. Another mapping containing small/big graphic data, then an array of graphic data for bosses.

I'll need to think about this a bit... So I'll convert this to a draft. I will:

- [ ] : Create separate maps for "Blind Game Logic Data" and "Blind Graphics Data" (names subject to change.

And, up to you, but I can also do the following in this PR (though, I would like to separate this into another PR, just because it's going to become bigger):

- [ ] : Convert blind graphics to use grit output (involves a larger refactor for getting the background color from the blind palette.

MeirGavish:
Yes, that's what I mean, one mapping for the general data that will not grow with more boss blinds, I missed that boss blinds would share the same tid/pid. And another second mapping to include the graphic data for small, big, and each boss blind, but I will admit it's still not fully thought out, I guess currently it just means palette data. I think the bigger refactor for blinds to use grit should be separated to another PR and I suppose the idea is it should only affect the second mapping but you never know what problems you will encounter when getting around to it.

ricfehr3:
@MeirGavish Okay, I'm at a good place with this one again.

I went ahead and reverted back to using the TYPE of the blinds as the enum for simply SMALL, BIG, and BOSS blinds. I separated out the graphics for each blind, but haven't touched the palette yet, or it's background changing indices (that's for later). 

So now there's a map with the data of the 3 blinds, but I've added a function to overwrite the BOSS tile and palette data. Not sure how it will be implemented yet, but now you can adjust the graphics by type index, which I think covers what we need quite well. 

Of course, there's a couple of TODO's , but nothing we haven't discussed before.

MeirGavish:
Yeah, it seems good, there's no point to overthinking this now, we just keep in mind that the BlindType enum doesn't grow when boss blinds are added.

Issue #125, opened 2025-09-16
Refactor magic out of game.c


ricfehr3:
The `game.c` file contains a few magic numbers. Making it confusing to read. 

For example:
```c
...
        memset16(&pal_bg_mem[14], 0x32BE, 1); // Reset the shop lights to correct colors
        memset16(&pal_bg_mem[17], 0x4B5F, 1);
        memset16(&pal_bg_mem[22], 0x5F9F, 1);
        memset16(&pal_bg_mem[8], 0xFFFF, 1);
...
```

drewjohnson2:
I'm looking to get a little more familiar with the code base, I can take care of this if you don't mind.

MeirGavish:
@ricfehr3 
@drewjohnson2 
Thanks a bunch! I refactored a lot of magic numbers from the code when I started working on this project, but there are still many left. You don't have to do everything at once, do it in batches and maybe we can open sub-issues for different sections or just pass this issue between developers if you get tired of refactoring...
For the palette indices, you can find newer code I added that doesn't use magic numbers if it helps, I think it was for the game over screen or the joker selling mechanic in the shop.

drewjohnson2:
@MeirGavish 

No problem at all.

The real struggle has been finding which palette indices correspond to what, especially since some indices are used for more than one thing.

Unless you just think it's better and makes testing easier, I don't mind taking a big bite out of this issue with one PR. It's really easy, just tedious.

MeirGavish:
Whatever works for you, doing it in batches was just a suggestion in case it makes it more managable. 
And yeah, the difficulty of making the difficult to understand easy to understand, is that you have to understand it, and that's difficult. 😅

cellos51:
I'm sorry for writing bad code. 😢 I'll learn my lesson one day.

drewjohnson2:
I've written some outrageous code in my day. Far, far more egregious than this, so no sweat.

Glass houses, stones etc.

ricfehr3:
> I'm sorry for writing bad code. 😢 I'll learn my lesson one day.

Don't stress at all. It's a cool project that wasn't initially made for other people. Not an outrageous sin to have a little spaghetti :)

PR #134, opened 2025-09-20
Magic numbers

drewjohnson2:
This PR addresses Issue #125.

There's a lot of small changes here.

Some of these names will need to be changed to be more accurate, or changed to make the accurate at all.
Some magic numbers have been left for one of two reasons:

1. I couldn't make heads or tails of what that value was
 
or

2. I just overlooked it by mistake.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -30,38 +30,104 @@
 
 #include "list.h"
 
+#define ITEM_SHOP_Y 71 // TODO: Needs to be a rect?
+
+#define MAIN_MENU_BUTTONS 2
+#define MAIN_MENU_IMPLEMENTED_BUTTONS 1 // Remove this once all buttons are implemented
+
+//TODO: Properly define and use
+#define MENU_POP_OUT_ANIM_FRAMES 20
+#define GAME_OVER_ANIM_FRAMES 15
+
+#define SCORED_CARD_TEXT_Y 48
+
+#define HIGHLIGHT_COLOR 0xFFFF
+
+#define PITCH_STEP_DISCARD_SFX      (-64)
+#define PITCH_STEP_DRAW_SFX         24
+#define PITCH_STEP_UNDISCARD_SFX    2*PITCH_STEP_DRAW_SFX    
+
+// Timer defs
+#define TM_ZERO 0
+#define TM_RESET_STATIC_VARS 30
+#define TM_END_POP_MENU_ANIM 13
+#define TM_START_ROUND_END_MENU_AMIN 1
+#define TM_END_DISPLAY_FIN_BLIND 30
+#define TM_END_DISPLAY_SCORE_MIN 4
+#define TM_ELLIPSIS_PRINT_MAX_TM 16
+#define TM_DISPLAY_REWARDS_CONT_WAIT 30
+#define TM_HAND_REWARD_INCR_WAIT 45
+#define TM_DISMISS_ROUND_END_TM 20
+#define TM_CREATE_SHOP_ITEMS_WAIT 1
+#define TM_SHIFT_SHOP_ICON_WAIT 7
+#define TM_END_GAME_SHOP_INTRO 12
+#define TM_SHOP_PRC_INPUT_START 1
+#define TM_DISP_BLIND_PANEL_FINISH 7
+#define TM_DISP_BLIND_PANEL_START 1
+#define TM_BLIND_SELECT_START 1
+#define TM_END_ANIM_SEQ 12
+
 // This could go in game.h or some other header somewhere
 // but it's essentially private to this file, so might as well
 // leave it.
 typedef enum
 {
-	PLAY_HAND_BTN_SELECTED_BORDER = 1,
-	BOSS_BLIND_PRIMARY = 1,
-	BLIND_BG_SHADOW = 2,
-	MAIN_MENU_PLAY_BUTTON_OUTLINE = 2,
-	REROLL_BTN = 3,
-	BLIND_BG_SECONDARY = 5,
-	BLIND_SKIP_BTN = 5, // 5, in the blind select, is the score multiplier and deck PID index also. In shop it's the next round selected border.
-	MAIN_MENU_PLAY_BUTTON_MAIN_COLOR = 5,
-	NEXT_ROUND_BTN_SELECTED_BORDER = 5,
-	SHOP_PANEL_SHADOW = 6,
-	BOSS_BLIND_SHADOW = 7,
-	PLAY_HAND_BTN = 7,
-	REROLL_BTN_SELECTED_BORDER = 7,
-	SHOP_LIGHTS_1 = 8,
-	DISCARD_BTN_SELECTED_BORDER = 9,
-	BLIND_SKIP_BTN_SELECTED_BORDER = 10,
-	DISCARD_BTN = 12, // 12 appears to also be the score multiplier, discard button, and deck PID index
-	SHOP_LIGHTS_2 = 14,
-	BLIND_SELECT_BTN = 15,
-	NEXT_ROUND_BTN = 16, // 16, in shop, is the next round button, multiplier, and deck PID.
-	SHOP_LIGHTS_3 = 17,
-	BLIND_SELECT_BTN_SELECTED_BORDER = 18,
-	BLIND_BG_PRIMARY = 19,
-	REWARD_PANEL_BORDER = 19,
-	SHOP_LIGHTS_4 = 22,
-	SHOP_BOTTOM_PANEL_BORDER = 26
+    PLAY_HAND_BTN_SELECTED_BORDER = 1,
+    BOSS_BLIND_PRIMARY = 1,
+    BLIND_BG_SHADOW = 2,
+    MAIN_MENU_PLAY_BUTTON_OUTLINE = 2,
+    REROLL_BTN = 3,
+    BLIND_BG_SECONDARY = 5,
+    BLIND_SKIP_BTN = 5, // 5, in the blind select, is the score multiplier and deck PID index also. In shop it's the next round selected border.
+    MAIN_MENU_PLAY_BUTTON_MAIN_COLOR = 5,
+    NEXT_ROUND_BTN_SELECTED_BORDER = 5,
+    SHOP_PANEL_SHADOW = 6,
+    BOSS_BLIND_SHADOW = 7,
+    PLAY_HAND_BTN = 7,
+    REROLL_BTN_SELECTED_BORDER = 7,
+    SHOP_LIGHTS_1 = 8,
+    DISCARD_BTN_SELECTED_BORDER = 9,
+    BLIND_SKIP_BTN_SELECTED_BORDER = 10,
+    DISCARD_BTN = 12, // 12 appears to also be the score multiplier, discard button, and deck PID index
+    SHOP_LIGHTS_2 = 14,
+    BLIND_SELECT_BTN = 15,
+    NEXT_ROUND_BTN = 16, // 16, in shop, is the next round button, multiplier, and deck PID.
+    SHOP_LIGHTS_3 = 17,
+    BLIND_SELECT_BTN_SELECTED_BORDER = 18,
+    BLIND_BG_PRIMARY = 19,
+    REWARD_PANEL_BORDER = 19,
+    SHOP_LIGHTS_4 = 22,
+    SHOP_BOTTOM_PANEL_BORDER = 26
 } _UIElementPID;
+
+typedef enum
+{
+    GAME_SHOP_INTRO,
+    GAME_SHOP_ACTIVE,
+    GAME_SHOP_EXIT
+} _GameShopStates;
+
+typedef enum
+{
+    ROUND_END_START,
+    START_EXPAND_POPUP,
+    DISPLAY_FINISHED_BLIND,
+    DISPLAY_SCORE_MIN,
+    UPDATE_BLIND_REWARD,
+    BLIND_PANEL_EXIT,
+    DISPLAY_REWARDS,
+    DISPLAY_CASHOUT,
+    DISMISS_ROUND_END_PANEL,
+    ROUND_END_EXIT
+} _GameRoundEndStates;
+
+typedef enum
+{
+    START_ANIM_SEQ,
+    BLIND_SELECT,
+    BLIND_SELECTED_ANIM_SEQ,
+    DISPLAY_BLIND_PANEL
+} _BlindSelectStates;
 
 static uint rng_seed = 0;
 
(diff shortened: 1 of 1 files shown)


source/game.c
```diff
@@ -2819,61 +2959,60 @@ void game_blind_select()
...
-                timer = 0; // Reset the timer
+                timer = TM_ZERO; // Reset the timer
             }
             break;
         }
-        case 3: // Move the blind panel into view
+        case DISPLAY_BLIND_PANEL: // Move the blind panel into view
         {
-            if (timer < 7)
+            if (timer >= TM_DISP_BLIND_PANEL_FINISH)
```
drewjohnson2, line 2931: Just inverted the condition here. Getting rid of some identation.

source/game.c
```diff
@@ -29,6 +30,68 @@
 
 #include "list.h"
 
+// This could go in game.h or some other header somewhere
+// but it's essentially private to this file, so might as well
+// leave it.
+typedef enum
```
drewjohnson2: These enums could be in a header somewhere, I suppose, but since they're only used here might as well keep them private to the file.

source/game.c
```diff
@@ -29,6 +30,68 @@
...
+{
+    PLAY_HAND_BTN_SELECTED_BORDER = 1,
+    BOSS_BLIND_PRIMARY = 1,
+    BLIND_BG_SHADOW = 2,
+    MAIN_MENU_PLAY_BUTTON_OUTLINE = 2,
+    REROLL_BTN = 3,
+    BLIND_BG_SECONDARY = 5,
+    BLIND_SKIP_BTN = 5, // 5, in the blind select, is the score multiplier and deck PID index also. In shop it's the next round selected border.
```
drewjohnson2: PID's 5 and 12 seem to be used for several things.

There are several duplicated enum values, but I figured this was best just for clarity.

source/game.c
drewjohnson2: Alternatively these could be `#define`s, but personally I prefer the organization of an enum in this case. Willing to change to the former if anyone else prefers that, though.

ricfehr3:
Sweet! This is awesome.

source/game.c
```diff
@@ -756,72 +890,72 @@ void set_hand()
     {
```
ricfehr3, line 854: Not for this PR, let's make an issue. This is a good opportunity to refactor. Depending on how others feel about the design. I keep trying to sneak in the array maps in here, but I prefer them, and I think it would work great here. 

:heavy_check_mark: Extensible
:heavy_check_mark: Encapsulated data
:heavy_check_mark: Shrinks code of these switch cases.
:heavy_check_mark: Great opportunity to use X macros to further make it extendable and idomatic. 

Probably don't have to convince you too much @drewjohnson2 , but for anyone wondering my thought process about it.

drewjohnson2: I agree. 

I was thinking this, and many of the game state sub states, would be well served as a function table indexed with the enum values.

source/game.c
```diff
@@ -3039,4 +3178,4 @@ void game_update()
             game_win();
             break;
     }
-}
\ No newline at end of file
+}
```
ricfehr3, line 3144: Extreme nitpick! Remove this chunk.

source/game.c
ricfehr3: For others, here is an example of the result of a refactor similar to this. Look at `blind.c`

https://github.com/cellos51/balatro-gba/pull/133/files

source/game.c
drewjohnson2: I think the diffviewer is bugging out. I can't really see which chunk you're speaking about.

ricfehr3: Darn. No problem. Not a big deal at all. Btw, PR looks good to me!

drewjohnson2: Well now I have to know.

include/graphic_utils.h
```diff
@@ -67,6 +67,9 @@
 // By default TTE characters occupy a single tile
 #define TTE_CHAR_SIZE TILE_SIZE
 
+// Some defines for common register bits
+#define SE_HORIZONTAL_FLIP 0x0400
```
MeirGavish: tonc already has SE_HFLIP, use that instead

source/game.c
MeirGavish: I think it's better to have these in the .c file rather than the header, there's no reason they should be accessible elsewhere.

However, I do think these should be defines and not an enum - the main reason is that I don't think they should be all grouped together in a single type. Different game states have different background files loaded with different palette memory and mapping (though there isn't a 1:1 game state:background correspondence) so it makes more sense to group these by backgrounds and separate the constants to different files once the game states are separated to different files.
Other than that, these feel more like constant index values rather than something with some serial significance to warrant an enum.
And I also preferred that they have a "PID" identifier suffix in the name so the type of constant is recognizable, some of my old constants can be renamed to be shortened and better match yours "MAIN_MENU_PLAY_BTN_BORDER_CLR_PID" etc.

source/game.c
MeirGavish: Agreed, well done. As said, since different backgrounds are loaded, other colors will be at the same index.

source/game.c
```diff
@@ -276,6 +336,77 @@ static const BG_POINT JOKER_DISCARD_TARGET  = {240,     30};
...
+#define CARD_DRAW_POS_X 208
+#define CARD_DRAW_POS_Y 110
+#define CUR_BLIND_TOKEN_POS_X 8
+#define CUR_BLIND_TOKEN_POS_Y 18
+#define MAIN_MENU_ACE_TX 88
+#define MAIN_MENU_ACE_TY 26 
+#define CARD_DISCARD_PNT_X 240
+#define CARD_DISCARD_PNT_Y 70
+#define HAND_START_POS_X 120
+#define HAND_START_POS_Y 90
```
MeirGavish: For constants that have both an x and y value, I preferred using a static const POINT and access it's x, y fields, then you define a single value rather than two.

source/game.c
```diff
@@ -756,72 +890,72 @@ void set_hand()
...
+        chips = FLUSH_FIVE_CHIPS;
+        mult = FLUSH_FIVE_MULT;
         break;
     case NONE:
-        chips = 0;
-        mult = 0;
+        chips = NONE_CHIPS;
+        mult = NONE_MULT;
         break;
     }
```
MeirGavish, line 924: I'm not sure about defining constants for these, this function pretty much defines the mapping so in a way I think you can say these aren't really magic numbers even though they look like ones and adding the defines just adds a redundant layer.

However, I do think it would be good to make a static array of base chips and mults for the hands in preparation for planet cards and leveling up hands. Then the hands can start at the base values but this function would return the current value according to the level.
Maybe it would be better in a struct that would contain the chips and mult as fields.

source/game.c
```diff
@@ -756,72 +890,72 @@ void set_hand()
...
-        chips = 30;
-        mult = 3;
+        chips = THREE_OAK_CHIPS;
+        mult = THREE_OAK_MULT;
         break;
     case STRAIGHT:
         print_hand_type("STRT");
-        chips = 30;
-        mult = 4;
+        chips = STRAIGHT_FLUSH_CHIPS;
```
```suggestion
        chips = STRAIGHT_CHIPS;
```

The added layer adds a potential for errors, despite this fix suggestion I think the defines should be removed, see my other comment below

source/game.c
```diff
@@ -1206,8 +1340,8 @@ static void game_playing_process_hand_select_input()
     {
         if (discard_button_highlighted == false) // Play button logic
         {
-            memset16(&pal_bg_mem[1], 0xFFFF, 1);
-            memcpy16(&pal_bg_mem[9], &pal_bg_mem[12], 1);
+            memset16(&pal_bg_mem[PLAY_HAND_BTN_SELECTED_BORDER], 0xFFFF, 1);
```
```suggestion
            memset16(&pal_bg_mem[PLAY_HAND_BTN_SELECTED_BORDER], HIGHLIGHT_COLOR, 1);
```

Since we're already refactoring magic numbers

source/game.c
```diff
@@ -2781,24 +2919,26 @@ void game_blind_select()
...
+                    timer = TM_ZERO;
                 }
             }
 
             if (selection_y == 0)
             {
-                memset16(&pal_bg_mem[18], 0xFFFF, 1);
-                memcpy16(&pal_bg_mem[10], &pal_bg_mem[5], 1);
+				// 5 is the multiplier palette color and the skip button color
+                memset16(&pal_bg_mem[BLIND_SELECT_BTN_SELECTED_BORDER], 0xFFFF, 1);
```
```suggestion
                memset16(&pal_bg_mem[BLIND_SELECT_BTN_SELECTED_BORDER], HIGHLIGHT_COLOR, 1);
```

source/game.c
```diff
@@ -2781,24 +2919,26 @@ void game_blind_select()
...
+                memcpy16(&pal_bg_mem[BLIND_SKIP_BTN_SELECTED_BORDER], &pal_bg_mem[BLIND_SKIP_BTN], 1);
             }
             else
             {
-                memcpy16(&pal_bg_mem[18], &pal_bg_mem[15], 1);
-                memset16(&pal_bg_mem[10], 0xFFFF, 1);
+				// 15 is the select button color
+                memcpy16(&pal_bg_mem[BLIND_SELECT_BTN_SELECTED_BORDER], &pal_bg_mem[BLIND_SELECT_BTN], 1);
+                memset16(&pal_bg_mem[BLIND_SKIP_BTN_SELECTED_BORDER], 0xFFFF, 1);
```
```suggestion
                memset16(&pal_bg_mem[BLIND_SKIP_BTN_SELECTED_BORDER], HIGHLIGHT_COLOR, 1);
```

source/game.c
```diff
@@ -2781,24 +2919,26 @@ void game_blind_select()
...
-                memcpy16(&pal_bg_mem[10], &pal_bg_mem[5], 1);
+				// 5 is the multiplier palette color and the skip button color
+                memset16(&pal_bg_mem[BLIND_SELECT_BTN_SELECTED_BORDER], 0xFFFF, 1);
+                memcpy16(&pal_bg_mem[BLIND_SKIP_BTN_SELECTED_BORDER], &pal_bg_mem[BLIND_SKIP_BTN], 1);
             }
             else
             {
-                memcpy16(&pal_bg_mem[18], &pal_bg_mem[15], 1);
-                memset16(&pal_bg_mem[10], 0xFFFF, 1);
+				// 15 is the select button color
```
```suggestion
```

The comment can be removed once the magic number is refactored out.

source/game.c
```diff
@@ -2819,61 +2959,60 @@ void game_blind_select()
...
                     main_bg_se_copy_rect(from, to);
                 }
+            
+                int y = 6;
+                memset16(&se_mem[MAIN_BG_SBB][32 * (y - 1)], 0x0006, 1);
+                memset16(&se_mem[MAIN_BG_SBB][1 + 32 * (y - 1)], 0x0007, 2);
+                memset16(&se_mem[MAIN_BG_SBB][3 + 32 * (y - 1)], 0x0008, 1);
+                memset16(&se_mem[MAIN_BG_SBB][4 + 32 * (y - 1)], 0x0009, 4);
+                memset16(&se_mem[MAIN_BG_SBB][7 + 32 * (y - 1)], 0x000A, 1);
+                memset16(&se_mem[MAIN_BG_SBB][8 + 32 * (y - 1)], SE_HORIZONTAL_FLIP | 0x0006, 1);
```
```suggestion
                memset16(&se_mat[MAIN_BG_SBB][y - 1][0], 0x0006, 1);
                memset16(&se_mat[MAIN_BG_SBB][y - 1][1], 0x0007, 2);
                memset16(&se_mat[MAIN_BG_SBB][y - 1][3], 0x0008, 1);
                memset16(&se_mat[MAIN_BG_SBB][y - 1][4], 0x0009, 4);
                memset16(&se_mat[MAIN_BG_SBB][y - 1][7], 0x000A, 1);
                memset16(&se_mat[MAIN_BG_SBB][y - 1][8], SE_HORIZONTAL_FLIP | 0x0006, 1); 
```

While we're at it, this should also clean the 32 multiplication magic number. This pattern is in so many places but since we touched this one let's clean it at least.
These changes should be tested to make sure I didn't make any mistakes, I will run this PR to test before merging anyway.

source/game.c
```diff
@@ -1219,8 +1353,9 @@ static void game_playing_process_hand_select_input()
         }
         else // Discard button logic
         {
-            memcpy16(&pal_bg_mem[1], &pal_bg_mem[7], 1);
-            memset16(&pal_bg_mem[9], 0xFFFF, 1);
+			// 7 is score and play hand button color
+            memcpy16(&pal_bg_mem[PLAY_HAND_BTN_SELECTED_BORDER], &pal_bg_mem[PLAY_HAND_BTN], 1);
+            memset16(&pal_bg_mem[DISCARD_BTN_SELECTED_BORDER], 0xFFFF, 1);
```
```suggestion
            memset16(&pal_bg_mem[DISCARD_BTN_SELECTED_BORDER], HIGHLIGHT_COLOR, 1);
```

source/game.c
MeirGavish: I think it would be better to move the hand scores refactor to a different PR

source/game.c
drewjohnson2: I actually refactored this entire function in the state refactor PR. What's funny is that even though I branched off of this branch for that feature, I didn't use the constants I defined lol

drewjohnson2:
@MeirGavish I think I've got all of your concerns addressed and it should be good for a review and test.

source/game.c
```diff
@@ -651,16 +611,16 @@ void change_background(int id)
...
+        memset16(&pal_bg_mem[SHOP_PANEL_SHADOW_PID], 0x10B4, 1);
         
-        memset16(&pal_bg_mem[SHOP_LIGHTS_2], 0x32BE, 1); // Reset the shop lights to correct colors
-        memset16(&pal_bg_mem[SHOP_LIGHTS_3], 0x4B5F, 1);
-        memset16(&pal_bg_mem[SHOP_LIGHTS_4], 0x5F9F, 1);
-        memset16(&pal_bg_mem[SHOP_LIGHTS_1], 0xFFFF, 1);
+        memset16(&pal_bg_mem[SHOP_LIGHTS_2_PID], 0x32BE, 1); // Reset the shop lights to correct colors
+        memset16(&pal_bg_mem[SHOP_LIGHTS_3_PID], 0x4B5F, 1);
+        memset16(&pal_bg_mem[SHOP_LIGHTS_4_PID], 0x5F9F, 1);
+        memset16(&pal_bg_mem[SHOP_LIGHTS_1_PID], HIGHLIGHT_COLOR, 1);
```
```suggestion
        memset16(&pal_bg_mem[SHOP_LIGHTS_1_PID], 0xFFFF, 1);
```

I actually think HIGHLIGHT_COLOR is less fitting in this one since it's the shop lights and not a button highlight, I think the shop light just happens to have the same color and these colors should have a different source.

drewjohnson2: Ah, good call. Probably shouldn't have done a substitute on the whole file for that pattern now that I think about it.

drewjohnson2:
@MeirGavish Okay, now I think it's good.

PR #139, opened 2025-09-21
Wrap around hand selection when going out of bounds horizontally

mannodermaus:
Currently, it is not possible to e.g. move from the first card in hand to the last by pressing Left while the first card is focused. This is a simple fix to allow that.

https://github.com/user-attachments/assets/4ccab210-e266-4f01-9611-a160164eee02

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -851,8 +851,21 @@ void card_draw()
 
 void hand_set_focus(int index)
 {
-    if (index < 0 || index > hand_top || hand_state != HAND_SELECT) return;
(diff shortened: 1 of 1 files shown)


MeirGavish:
Does the original Balatro do that? I don't remember and only have the mobile version with me right now to check. 😅

mannodermaus:
It does, yeah! I've only ever played the PS5 version of Balatro, which is how I noticed it missing from this port in the first place. 😆

PR #143, opened 2025-09-22
Add uploading ROM on tag push

ricfehr3:
This PR adds uploading the ROM to github on tag pushes prefixed with `v*`

I notice a lot of people wanting to try the game out but struggling to build it in the youtube comments. This should address it.

@MeirGavish Once this is merged on your first tag push it will release. You can check the CI logs for the uploaded url.

### DIFF
--- a/.github/workflows/build_ci_workflow.yml
+++ b/.github/workflows/build_ci_workflow.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - "v*"
   pull_request:
     branches:
       - main
@@ -19,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Get docker image cache
         uses: actions/cache@v4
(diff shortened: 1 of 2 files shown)


ricfehr3:
Dont merge yet. I need to update the workdir so it makes the rom name consistent `balatro-gba.gba`

ricfehr3:
@MeirGavish This one is good now.

MeirGavish (commented):
@ricfehr3 
Cool, thanks! Just one thing I want to make sure - that the workflow won't accidently make unintended releases. 
The case I'm thinking about is when making a PR with a tagged commit, the workflow will run to test it, so I want to make sure it doesn't create a release then and would only create one once the PR is merged. Copilot said it's fine and it makes sense it would only happen on merge/push but I want to confirm this with you.

ricfehr3:
> @ricfehr3 Cool, thanks! Just one thing I want to make sure - that the workflow won't accidently make unintended releases. The case I'm thinking about is when making a PR with a tagged commit, the workflow will run to test it, so I want to make sure it doesn't create a release then and would only create one once the PR is merged. Copilot said it's fine and it makes sense it would only happen on merge/push but I want to confirm this with you.

The workflow will make a release on every tag push that is prefixed with `v*`. So, `v1.0.0` will release as will `v2` or even `v0.0.1-testing-alpha`. We can apply a regex to match a particular scheme, but that's a bit much IMO.

The tag applies to the commit, not the branch. So if you were to push a tag before a merge, it will make a release, but only if it has the `v*` prefix. The workflow should be: `merge` then `tag`. 

Since tags are so manual, I figure it should be fine. You can hold off on merging it if you wanted more control over the version scheme to prevent bad releases?

MeirGavish:
I guess it should be okay, it just means the workflow should be to tag previously merged commits and not tag unpushed commits.

PR #145, opened 2025-09-22
kind of fixed how game_speed behaves

notsomealex:
one thing that I instantaneously missed when I tried this out was more game speed, so I checked and see if it was implemented, and it was, except some animations such as the discard hand and draw cards to deck ones kept the same speed. and the score seemed to bug kinda badly, since it wouldn't add up fast enough to match the game speed and would stay at a value below the target score. from my testing, this PR should fix all of that. you still have to modify the source to change the game_speed variable, but at least now it should work as expected, for speeds 1-3 at least (maybe would be a good idea to work on a prototype of a menu, so that variable can be changed in-game).

**NOTE** that I made the game_speed variable not static, and included the game.h header in the sprite.c file, so that the function that moved the cards could access that variable. I really hope it doesn't bother.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -1279,7 +1279,7 @@ static void game_playing_process_input_and_state()
         if (mult > 0)
         {
             temp_score = chips * mult;
-            lerped_temp_score = int2fx(temp_score); // game speed affects how fast the score text increases.
+            lerped_temp_score = int2fx(temp_score);
             lerped_score = int2fx(score);
 
             display_temp_score(temp_score);


include/game.h
```diff
@@ -98,4 +98,8 @@ int get_deck_top(void);
 int get_num_discards_remaining(void);
 int get_money(void);
 
+extern int game_speed;  // had to do this use this it the sprites animations
```
ricfehr3: I would suggest keeping it `static` in the c file and instead use functions to get/set the static variable.

source/game.c
```diff
@@ -1269,7 +1269,7 @@ static void game_playing_process_input_and_state()
         if (mult > 0)
         {
             temp_score = chips * mult;
-            lerped_temp_score = int2fx(temp_score);
+            lerped_temp_score = (float)(1.0f / game_speed) * int2fx(temp_score); // game speed affects how fast the score text increases.
```
ricfehr3: Floating point operations are sluggish on GBA. It's best to use fixed point numbers.

include/game.h
notsomealex: noted, will do that. i really didn't like making it not static when the rest of the variables are, but didn't come up with that.

source/game.c
notsomealex: I actually just tested this on my DS Lite with GBARunner2 and the performance is fine, I don't have a true GBA to test it, though, but the DS uses the same hardware for running GBA games as the GBA.
but I will try to do it that way, for the sake of doing things right and in case performance turns out to be an issue when more things are added.

ricfehr3: Sweet, thank you!

The GBA doesn't have any hardware to handle floating point values at all. Not sure if that's the case on the NDS. So the compiler has to emulate it in software anyway. Its better to use fixed unless high precision is really required.

Libtonc has some really good docs on it: https://gbadev.net/tonc/fixed.html?highlight=Fixed#b-fixed-point-numbers-and-luts

notsomealex: the DS doesn't have a FPU either, and I've made some homebrew for it that uses float variables and division, and performance isn't an issue at all on real hardware. I don't know if for small implementations of these things, performance is not affected that much, or if devkitARM under the hood converts floating point  operations to fixed point, and converts them back to float to use them.

ricfehr3: It absolutely does work! You're right too, realistically there won't be a noticeable difference. 

However, while in some cases introducing known inefficient code has benefits (simplicity, time-crunch... etc.), I'd argue that there's no reason to introduce it here. There is a good reason that machines without an FPU avoid the `(float)` specifier, the difference is hidden but rather drastic. With devkitpro's toolchain, the compiler (GCC) uses a library to emulate floating point calls. This library is called **softfp**, you can [check out the code](https://bellard.org/softfp/). Maintaining the float format([IEEE 754](https://en.wikipedia.org/wiki/IEEE_754)) is not a simple task. As a reminder, a 32-bit float looks like this:



Note that the *mantissa* is the decimal portion of the value. Reserving 23 bits for precision. To maintain this precision, on arithmetic operations, there is work required to:
1. dissect these components to represent the actual number. 
2. Perform the arithmetic on these sections, including accounting for rounding and changes in the exponent.
3. Put all these components back together in the IEEE 754 format.

Since so much work is required, the improvement between fixed-point and `float` is drastic. Compare `libtonc`'s math library vs `softfloat.c` :

**Softfp**
```c
float32 float32_div(float32 a, float32 b, float_status *status)
{
    flag aSign, bSign, zSign;
    int_fast16_t aExp, bExp, zExp;
    uint32_t aSig, bSig, zSig;
    a = float32_squash_input_denormal(a, status);
    b = float32_squash_input_denormal(b, status);

    aSig = extractFloat32Frac( a );
    aExp = extractFloat32Exp( a );
    aSign = extractFloat32Sign( a );
    bSig = extractFloat32Frac( b );
    bExp = extractFloat32Exp( b );
    bSign = extractFloat32Sign( b );
    zSign = aSign ^ bSign;
    if ( aExp == 0xFF ) {
        if (aSig) {
            return propagateFloat32NaN(a, b, status);
        }
        if ( bExp == 0xFF ) {
            if (bSig) {
                return propagateFloat32NaN(a, b, status);
            }
            float_raise(float_flag_invalid, status);
            return float32_default_nan;
        }
        return packFloat32( zSign, 0xFF, 0 );
    }
    if ( bExp == 0xFF ) {
        if (bSig) {
            return propagateFloat32NaN(a, b, status);
        }
        return packFloat32( zSign, 0, 0 );
    }
    if ( bExp == 0 ) {
        if ( bSig == 0 ) {
            if ( ( aExp | aSig ) == 0 ) {
                float_raise(float_flag_invalid, status);
                return float32_default_nan;
            }
            float_raise(float_flag_divbyzero, status);
            return packFloat32( zSign, 0xFF, 0 );
        }
        normalizeFloat32Subnormal( bSig, &bExp, &bSig );
    }
    if ( aExp == 0 ) {
        if ( aSig == 0 ) return packFloat32( zSign, 0, 0 );
        normalizeFloat32Subnormal( aSig, &aExp, &aSig );
    }
    zExp = aExp - bExp + 0x7D;
    aSig = ( aSig | 0x00800000 )<<7;
    bSig = ( bSig | 0x00800000 )<<8;
    if ( bSig <= ( aSig + aSig ) ) {
        aSig >>= 1;
        ++zExp;
    }
    zSig = ( ( (uint64_t) aSig )<<32 ) / bSig;
    if ( ( zSig & 0x3F ) == 0 ) {
        zSig |= ( (uint64_t) bSig * zSig != ( (uint64_t) aSig )<<32 );
    }
    return roundAndPackFloat32(zSign, zExp, zSig, status);

}
```

**Tonc:**
```c
//! Divide two fixed point values.
INLINE FIXED fxdiv(FIXED fa, FIXED fb)
{	return ((fa)*FIX_SCALE)/(fb);			}
```

If precision is an issue, `libtonc` even provides a macro for 64-bit operations:

```c

//! Divide two fixed point values using 64bit math.
INLINE FIXED fxdiv64(FIXED fa, FIXED fb)
{	return ( ((s64)fa)<<FIX_SHIFT)/(fb);	}
```

Sorry if the explanation was a bit much, but I wanted to share for anyone else reading :)

MeirGavish: We should avoid floating point operations altogether and avoid any unnecessary performance degradation even if it's currently not visible. We want to have the swirling HBLANK background effect from the main menu during gameplay too but currently it causes performance issues so we want to keep the code as optimized as possible so it might be possible eventually.

MeirGavish:
Thanks for working on this! I think it never fully worked since it was added but honestly I kind of forgot about game speed at some point during development 😅

notsomealex:
it should be done now. the only thing is that in the sprite_object_update integer division is used, in the following way:
`sprite_object->vx += (sprite_object->tx - sprite_object->x) / (8 / get_game_speed());`
which makes not possible for the speed to be more than 8, because of division by zero. I left it there because a game_speed of 8 is extremely fast anyways, and using fixed division for more precision would be a waste of resources.

I have a question, though, how is integer division handled by the GBA? I know it doesn't have a division unit, so I'd be hesitant to use division at all, but I see it used widely. my guess is the ARM7 handles it just fine (?)

ricfehr3:
> it should be done now. the only thing is that in the sprite_object_update integer division is used, in the following way: `sprite_object->vx += (sprite_object->tx - sprite_object->x) / (8 / get_game_speed());` which makes not possible for the speed to be more than 8, because of division by zero. I left it there because a game_speed of 8 is extremely fast anyways, and using fixed division for more precision would be a waste of resources.
> 
> I have a question, though, how is integer division handled by the GBA? I know it doesn't have a division unit, so I'd be hesitant to use division at all, but I see it used widely. my guess is the ARM7 handles it just fine (?)

Instead of `a / (8 / b)` use `(a * b) / 8` for the safer/faster equivalent.

For your question, `how is integer division handled by the GBA`? That is a good question. Every CPU will have hardware to do division. The math behind how it does it is pretty nuts, I don't fully understand it, not even close ([wikipedia page](https://en.wikipedia.org/wiki/Division_algorithm#Fast_division_methods) if you're brave).

In general you don't need to worry about optimizing division. Little things like bit-shifts to divide by two come in handy and are always welcome, but in general division is fine.

I looked up the division assembly instructions in the GBA CPU datasheet. I was surprised to see that on top of the additional cycles it takes to perform division, there is some overhead with operation. It makes sense, it needs to protect the hardware from division by 0 conditions. There is some code that gets ran to let devs know there's a problem, specifically a divide by 0 exception.   

Anyway, it looks like it does a lot more than just a mulitply call! So good catch on being suspicious of it. Totally you should multiply if possible over division.

image

Ref [ARMv7-M Manual](https://web.eecs.umich.edu/~prabal/teaching/eecs373-f10/readings/ARMv7-M_ARM.pdf)

source/sprite.c
```diff
@@ -157,9 +158,14 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
 
 void sprite_object_update(SpriteObject* sprite_object)
 {
```
ricfehr3, line 160: Instead of a / (8 / b) use (a * b) / 8 for the safer/faster equivalent.

(This is in another comment but github is weird with what a "review" is)

notsomealex: done. also thanks for your explanations:D

ricfehr3: Sweet, looks great! And no problem at all.

ricfehr3: @pansythoughts FYI, you have sent me down a rabbit hole into the magic of compilers. I came across this video and immediately thought of our discussion on division. It never ceases to amaze me how complex computers are. He explains it better than I can:

https://www.youtube.com/watch?v=ssDBqQ5f5_0

source/game.c
```diff
@@ -182,6 +182,16 @@ int get_num_discards_remaining(void) {
...
 }
+int get_game_speed(void)
+{
+    return game_speed;
+}
+
+// for the future when a menu actually lets this variable be changed.
+void set_game_speed(int game_speed)
+{
+    game_speed = game_speed;
```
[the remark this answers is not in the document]
notsomealex: completely true. my bad.

source/game.c
```diff
@@ -1269,7 +1279,7 @@ static void game_playing_process_input_and_state()
         if (mult > 0)
         {
             temp_score = chips * mult;
-            lerped_temp_score = int2fx(temp_score);
+            lerped_temp_score = fx2int(fxdiv(int2fx(1), int2fx(game_speed)) * int2fx(temp_score)); // game speed affects how fast the score text increases.
```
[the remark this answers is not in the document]
notsomealex: I originally thought of doing what Copilot says, since it makes sense, but is actually wrong. if it's done that way, the score is (my guess), the original score times the game speed. so, if the game speed is 2, then the temp score is double the original score, as seen here:
bad

whereas the original code I provided (which divides the temp_score by the game speed), does behave correctly:

good

I don't exactly know why this works, but it does, (i tested it through most hands at many different amounts of scores, and found no issues). especially since the output is an integer, insted of a fixed point number. any way I tried to get this to work with the output being a fixed, just wouldn't.

PD: the old code (which just copies the temp score converted to fixed) behaves as follows:

old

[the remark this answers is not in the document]
ricfehr3: `lerped_temp_score` is a FIXED value, but Copilot wanted you to turn it into an integer with fx2int. So you're right to think it's wrong.

What's going on is the higher the game speed, the smaller the `lerped_temp_score`. Below in this file you can see that `lerped_temp_score` is subtracted from until it's below 0, then it displays the score.
 
I do think there is a bug with:

```c
lerped_temp_score = fx2int(fxdiv(int2fx(1), int2fx(game_speed)) * int2fx(temp_score));
```

I think what you want is:

```c
lerped_temp_score = fxdiv(FIXED_ONE, int2fx(game_speed)) * temp_score;
```

You can multiply a fixed point by an integer and it returns the fixed point version. So you get a ratio of the game speed (FIXED_ONE is a shortcut of int2fx(1)), that is then multiplied by `temp_score` which results in a fixed point value. 

Not 100% sure though, let me know how that works out. 

---
__Quick edit__: I think the equation is correct as it also matches your original intension with the `float` version (though note it still has a bug since `temp_score` should not be converted to a FIXED value): 

```c
(float)(1.0f / game_speed) * int2fx(temp_score);
```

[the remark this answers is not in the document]
notsomealex: interestingly enough, your snippet and mine behave exactly the same, just as seen on the second GIF in my previous comment. I do will leave yours since it makes the most sense.
I think both expressions are the same, because I just straight up multiply (not using fxmul) two fixed point numbers, which results in the fixed scale being applied twice, if that makes sense. So when it gets divided by the fixed scale in fx2int, it just leaves the regular fixed point number.
I admit that at first I was just playing around with the expression until it just worked and left it like that without much thought lol.

source/game.c
```diff
@@ -182,6 +182,16 @@ int get_num_discards_remaining(void) {
...
+int get_game_speed(void)
+{
+    return game_speed;
+}
+
+// for the future when a menu actually lets this variable be changed.
+void set_game_speed(int game_speed_)
+{
+    game_speed = game_speed_;
+}
```
```suggestion
void set_game_speed(int new_game_speed)
{
    game_speed = new_game_speed;
}
```

This trailing underscore just feels a little awkward

source/game.c
```diff
@@ -1269,7 +1279,7 @@ static void game_playing_process_input_and_state()
         if (mult > 0)
         {
             temp_score = chips * mult;
-            lerped_temp_score = int2fx(temp_score);
+            lerped_temp_score = fxdiv(FIX_ONE, int2fx(game_speed)) * temp_score;; // game speed affects how fast the score text increases.
```
MeirGavish: So there's been some discussion about this calculation but there's something else I want to point out about it. I actually wouldn't say this is the right place to do this, what happens here is that during the accumulation animation, the scores themselves are divided by the game speed so they are actually incorrect and they become correct when the animation is done.
Look at these screenshots with game speed 2, the pair is scored 30x2 = 60, but during the animation the score for the hand is shown to be 30 and when it's done it's back to 60.

Image
Before animation 1

Image
Before animation 2

Image
During the accumulation animation

Image
After the animation

It just so happens that it makes it move faster since the target value is divided by the game speed.

I actually think this calculation should be done in the PLAY_ENDED stage where the animation happens.

source/sprite.c
```diff
@@ -157,8 +158,8 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
-    sprite_object->vx += (sprite_object->tx - sprite_object->x) / 8;
-    sprite_object->vy += (sprite_object->ty - sprite_object->y) / 8;
+    sprite_object->vx += ((sprite_object->tx - sprite_object->x) * get_game_speed()) / 8; // SPEED
```
MeirGavish:
```suggestion
    sprite_object->vx += ((sprite_object->tx - sprite_object->x) * get_game_speed()) / 8; 
```

This comment seems redundant.

source/sprite.c
```diff
@@ -157,8 +158,8 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
+    sprite_object->vy += ((sprite_object->ty - sprite_object->y) * get_game_speed()) / 8;
```
MeirGavish:
```suggestion
    sprite_object->vy += ((sprite_object->ty - sprite_object->y) * get_game_speed()) / 8;
```
Another trailing whitespace

source/joker.c
```diff
@@ -83,7 +83,7 @@ static int get_unused_joker_pb()
 }
 
 static int allocate_pb_if_needed(u8 joker_id)
-{
+{
```
```suggestion
{
```

Sneaky hidden whitespaces

source/game.c
notsomealex: I simply learned to do it that way and have kept doing it with the underscore out of habit, but fair.

source/sprite.c
notsomealex: lol, sorry. I don't even remember when I made that comment and forgot to remove it.

source/game.c
notsomealex: true. as I said in another comment, I was just trying things out until I stumbled upon what worked, and left it like that. I'll commit what you have pointed out, thanks!

MeirGavish:
Looks good now

PR #146, opened 2025-09-22
Added more Jokers all outside of scope, some with art available already from MathisMartin31

noctowlhoot:
I was unsure of the significance of the number immediately after joker rarity in the joker registry, so I haven't put the joker effects in there. If anyone can tell me what the significance of that number is I can add them in :)

Jokers added
The Duo
The Trio
The Family
The Order
The Tribe
Bootstraps
Shoot the moon
Triboulet

### DIFF
--- a/source/joker_effects.c
+++ b/source/joker_effects.c
@@ -571,13 +571,10 @@ static JokerEffect shoot_the_moon_joker_effect(Joker *joker, Card *scored_card) 
     int hand_size = hand_get_size();
     for (int i = 0; i < hand_size; i++ )
     {
-        switch (hand[i]->card->rank) {
-            case QUEEN:
-            
-                    effect.mult += 13;
-            default:
-                break;
-        }
+        if (hand[i]->card->rank == QUEEN)
+        {
+             effect.mult += 13;
+        } 
     }
 
     return effect;


JustinL-12:
Hi, it looks like the second value in the joker registry is just the base cost of the joker. This page has a list of all of them. :) 
https://balatrowiki.org/w/Jokers

noctowlhoot:
Legend, thank you! Will pop them in, though not sure what to do about triboulet then which doesn't have a cost, though spectral cards are likely a while away anyway

MeirGavish:
Interesting that you added Triboulet, I haven't considered legendary jokers, this isn't just making an unlockable joker available, legendary jokers have their own system for obtaining. If you add it the same way as the other jokers, it will just be assigned a value and be able to appear in the shop like the other jokers (and be sold) but wouldn't that be a little weird?

MeirGavish:
Also, I now see that there's a PR for properly weighted jokers. I think implementing that would mean that legendary jokers would just not appear in the shop or in the game at all even if their code and art exist.

JustinL-12:
> Also, I now see that there's a PR for properly weighted jokers. I think implementing that would mean that legendary jokers would just not appear in the shop or in the game at all even if their code and art exist.

Yeah, I had it implemented to match the chances from the Balatro wiki, but im thinking now in the spirit of this project it might make sense to give legendary jokers like a 1% chance to be in the shop, as opposed to never showing up at all,. I'm not sure if there will ever be unlocking jokers in this port (or at least not for a while), so I'd say we should have a method of seeing them. And I guess give them the highest cost?

noctowlhoot:
Justin's idea seems plausible for now, but yeah from my perspective these are just here ready to go when and if you want to implement them!

source/joker_effects.c
```diff
@@ -464,6 +464,143 @@ static JokerEffect odd_todd_joker_effect(Joker *joker, Card *scored_card) {
...
+    int hand_size = hand_get_size();
+    for (int i = 0; i < hand_size; i++ )
+    {
+        switch (hand[i]->card->rank) {
+            case QUEEN:
+            
+                    effect.mult += 13;
+            default:
+                break;
+        }
```
```suggestion
        if (hand[i]->card->rank == QUEEN)
        {
             effect.mult += 13;
        } 
```

Since queen is the **only** case here, there's not much point for a switch case and it can be replaced with an if statement.

noctowlhoot: Done! Thank you for the tip that does make way more sense.

PR #149, opened 2025-09-22
Fix Dynamic Badges (hopefully)

GreenNinja2525:
I think I fixed the dynamic badges, it just needed a maxAge parameter to properly tell GitHub to recache the data. I can't really test this as it is set to refresh each badge every 2 hours. I also fixed some minor formatting in the build instructions. I also added a discussions badge so people can read through them before contributing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,9 @@
 # Balatro-GBA
 
-[![Build Status](https://img.shields.io/github/actions/workflow/status/cellos51/balatro-gba/build_ci_workflow.yml?style=flat&logo=github&branch=main&label=Builds&labelColor=gray&color=default&v=1)](https://github.com/cellos51/balatro-gba/actions)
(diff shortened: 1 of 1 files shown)


GreenNinja2525:
@MeirGavish to properly test this, if no other issues or pull requests are resolved, it should update the pull request badge to 14 on my repository since I opened this one in 2 hours. If it does update, then it can be merged. If it doesn't, I'll redo the badges entirely. I set it to 2 hours as setting it to an hour or lower can cause issues and make the badges freeze. Hopefully this fixes the issue and I don't have to make different badges.

GreenNinja2525:
@MeirGavish it did not work, I will redo the badges later, I really thought it would work this time. Oh well I guess...

GreenNinja2525:
@MeirGavish I think I finally fixed the badges. They should automatically recache themselves and have custom icons. I used Shields.io to make the badges and they are much more reliable than what I was using before. I can't really test these quite yet but it would be fine to merge as they look identical to the old ones, after merging, it should update and go down by one within an hour (after a page refresh usually) The only difference is the added discussion badge that wasn't part of the original set. These should update automatically, just to reiterate.

PR #153, opened 2025-09-24
Add git hash baked into ROM for bug reports

ricfehr3:
Bake in a git commit hash to help identify card versions for dev. Helpful for bug reports.

https://github.com/cellos51/balatro-gba/issues/141

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,10 +38,16 @@ GRAPHICS	:= graphics
 #---------------------------------------------------------------------------------
 ARCH	:=	-mthumb -mthumb-interwork
 
+GIT_DIRTY := $(shell git diff-index --quiet HEAD -- || echo "-dirty")
+GIT_HASH := $(shell git rev-parse --short HEAD)
+GIT_C_FLAGS := -DGIT_HASH=\"$(GIT_HASH)\" -DGIT_DIRTY=\"$(GIT_DIRTY)\"
+
 CFLAGS	:= -g -O3 -Wall -Werror\
         -mcpu=arm7tdmi -mtune=arm7tdmi \
         -ffast-math -fomit-frame-pointer -funroll-loops \
         $(ARCH)
+
+CFLAGS  += $(GIT_C_FLAGS)
 
 CFLAGS	+=	$(INCLUDE)
 
--- a/include/version.h
+++ b/include/version.h
@@ -1,0 +1,7 @@
+#ifndef VERSION_H
+#define VERSION_H
+
+__attribute__((used))
+const char gba_version[] = "GBA_VERSION:" GIT_HASH GIT_DIRTY;
+
+#endif // VERSION_H
--- a/source/game.c
+++ b/source/game.c
@@ -28,6 +28,7 @@
 #include "soundbank.h"
 
 #include "list.h"
+#include "version.h"
 
 static uint rng_seed = 0;
 


Makefile
```diff
@@ -38,11 +38,17 @@ GRAPHICS	:= graphics
 #---------------------------------------------------------------------------------
 ARCH	:=	-mthumb -mthumb-interwork
 
+GIT_DIRTY := $(shell git diff-index --quiet HEAD -- || echo "-dirty")
```
MeirGavish, line 41: What does "dirty" mean?

source/version.c
```diff
@@ -0,0 +1,15 @@
...
+#ifndef GIT_HASH
+#define GIT_HASH "undef"
+#endif
+
+#ifndef GIT_DIRTY
+#define GIT_DIRTY "-dirty"
+#endif
+
+__attribute__((used))
+const char balatro_version[] = "GBALATRO_VERSION:" GIT_HASH GIT_DIRTY;
```
MeirGavish, line 10: Not too fond of having a global variable definition in a header... It could cause issues if it's included in multiple places.. which I guess theoretically shouldn't happen but still, maybe this definition should be in a .c file.

Makefile
ricfehr3: It's semi-common terminology for describing a git repo with uncommitted local changes. "Dirty worktree".

source/version.c
ricfehr3: So true. I'll switch it to a C file and make it static.

source/version.c
```diff
@@ -0,0 +1,10 @@
+#ifndef GIT_HASH
+#define GIT_HASH "undef"
+#endif
+
+#ifndef GIT_DIRTY
+#define GIT_DIRTY "-dirty"
+#endif
+
+__attribute__((section(".version"), used))
```
ricfehr3, line 9: It creates a custom `version` section in the elf. You can checkout this section with good ol' `objdump`

```sh
/opt/devkitpro/devkitARM/bin/arm-none-eabi-objdump -s -j .version build/balatro-gba.elf

build/balatro-gba.elf:     file format elf32-littlearm

Contents of section .version:
 819fbe4 4742414c 4154524f 5f564552 53494f4e  GBALATRO_VERSION
 819fbf4 3a353662 62633266 2d646972 747900    :56bbc2f-dirty.
```

Makefile
```diff
@@ -38,17 +38,23 @@ GRAPHICS	:= graphics
...
 
+CFLAGS  += $(GIT_C_FLAGS)
+
 CFLAGS	+=	$(INCLUDE)
 
 CXXFLAGS	:=	$(CFLAGS) -fno-rtti -fno-exceptions
 
 ASFLAGS	:=	-g $(ARCH)
-LDFLAGS	=	-g $(ARCH) -Wl,-Map,$(notdir $*.map)
+LDFLAGS	=	-g $(ARCH) -Wl,-Map,$(notdir $*.map),--undefined=balatro_version
```
ricfehr3, line 57: This makes the linker put it in the elf file, even if it's unused.

scripts/build_bug_report.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
```
ricfehr3, line 1: This script's name isn't 100% accurate yet. 

We can choose to keep it or not, but I wanted a portable way to get that `GBALATRO_VERION` string, so I chose python.

drewjohnson2:
This is really cool. Great idea.

Makefile
MeirGavish: @ricfehr3 and does that mean we will see if there are uncommitted changes? What about if they're not staged?

MeirGavish (approved):

Makefile
ricfehr3: @MeirGavish Hey, thanks for merging this. It won't say 'dirty' if there's a local commit, but it will have a new commit hash.

MeirGavish: Yeah, I was just wondering if it matters if the uncommitted changes are staged.

ricfehr3: Oh my bad. Yes it will say 'dirty' on staged but uncommitted changes.

PR #155, opened 2025-09-26
State refactor

drewjohnson2:
Still currently a work in progress while I self-review.

Merging is dependent on the merge of PR #134 since this branch has been based off of that branch. Because of that the diff is a bit larger than it actually is.


include/game.h
```diff
@@ -75,10 +76,18 @@ enum HandType
     FLUSH_FIVE
 };
 
+typedef struct 
+{
+    enum GameState state;
```
drewjohnson2: I don't even think the state variable is used. Can probably remove.

source/game.c
```diff
@@ -2,9 +2,12 @@
 
 #include <maxmod.h>
 #include <tonc.h>
-#include <string.h>
 #include <stdlib.h>
 
+#include "tonc_input.h"
```
drewjohnson2: Clangd really loves importing these headers when using autocomplete even though it's unnecessary.

source/game.c
```diff
@@ -23,22 +25,185 @@
...
+// https://github.com/cellos51/balatro-gba/issues/137#issuecomment-3322485129
+static void _noop() {}
+
+// These functions need to be forward declared
+// so they're visible to the state_info array,
+// and the sub-state function tables.
+// This could be done, and maybe should be done,
+// with an X macro, but I'll leave that to the 
+// reviewer(s).
+static void game_main_menu_on_init();
```
drewjohnson2, line 83: The comment above is very relevant because the forward declaration of all of these functions unfortunately contributes further to the bloat of this file.

source/game.c
```diff
@@ -23,22 +25,185 @@
...
 static uint rng_seed = 0;
 
 static uint timer = 0; // This might already exist in libtonc but idk so i'm just making my own
 static int game_speed = 1;
 static int background = 0;
 
-static enum GameState game_state = GAME_SPLASH_SCREEN; // The current game state, this is used to determine what the game is doing at any given time
+static const StateInfo state_info[] = 
+{
+#define DEF_STATE_INFO(stateEnum, init_fn, update_fn, exit_fn) \
```
drewjohnson2, line 123: I like an X macro here. Leaves room, and saves _some_ typing, for new possible game states (not that I know there will be any new ones).

source/game.c
```diff
@@ -23,22 +25,185 @@
...
+    game_round_end_update_blind_reward, game_round_end_panel_exit,
+    game_round_end_display_rewards, game_round_end_display_cashout,
+    game_round_end_dismiss_round_end_panel
+};
+
+static enum GameState game_state = GAME_STATE_SPLASH_SCREEN; // The current game state, this is used to determine what the game is doing at any given time
 static enum HandState hand_state = HAND_DRAW;
 static enum PlayState play_state = PLAY_PLAYING;
-static int state = 0; // General state variable, used for switch statements in each game state related function
+static int substate = 0; // maybe put this in StateInfo struct?
```
drewjohnson2: `substate` seemed like a more accurate name since it really tracks the state of the game state.

Also: This could maybe live in the state info struct?

source/game.c
```diff
@@ -2992,9 +3123,69 @@ static void game_lose()
...
     }
+
+    if (key_hit(KEY_START)) game_change_state(GAME_STATE_BLIND_SELECT);
 }
 
-static void game_win()
+// This function isn't set in stone. This is just a placeholder
+// allowing the player to restart the game. Thought it would be nice to have
+// util we decide what we want to do after a game over.
+static void game_lose_on_exit()
```
drewjohnson2: Comment above is very important. This function is really just a placeholder.

I think it would be good to keep for now so the game can be reset.

source/main.c
```diff
@@ -77,6 +77,7 @@ void init()
     blind_init();
     joker_init();
     game_init();
+    game_change_state(GAME_STATE_SPLASH_SCREEN);
```
drewjohnson2, line 80: For the state refactor to work as intended, I needed to remove the call to `game_change_state` from `game_init`. Now that I think of it, I'm not ever sure we really need to call the change state function. Will investigate.

drewjohnson2: nvm

include/game.h
ricfehr3: We don't need it. If it's not useful we can add it later. If you were declaring a map statically it could work, but here, eh.

include/game.h
```diff
@@ -28,14 +28,15 @@
...
 {
-    GAME_SPLASH_SCREEN,
-    GAME_MAIN_MENU,
-    GAME_PLAYING,
-    GAME_ROUND_END,
-    GAME_SHOP,
-    GAME_BLIND_SELECT,
-    GAME_LOSE,
-    GAME_WIN
+    GAME_STATE_SPLASH_SCREEN,
```
ricfehr3: This can be X-macro'd as well. You may want to since you define the enum names in the def table.

source/game.c
ricfehr3: I think that static declarations are fine for forward declaration since the type specifier locks it to one memory location. Not sure tho. Something we could test to be sure.

I am thinking that maybe these functions should be in their own files. Remove bloat from this file and isolate functionality of each state into the scope of a single file. Though, that method may actually add to the bloat.

source/game.c
```diff
@@ -135,6 +300,18 @@ static inline Card *discard_pop()
     return discard_pile[discard_top--];
 }
 
+static inline void reset_background_color()
```
ricfehr3: I'm not sure why y is 6 here, or why 1 is subtracted to it.

I don't think any new code needs to be written, just a comment and changing y to either a more detailed name and be declared `const`

source/game.c
```diff
@@ -52,6 +213,10 @@ static Sprite *blind_select_tokens[MAX_BLINDS] = {NULL}; // The sprites that dis
 static int current_blind = SMALL_BLIND;
 static enum BlindState blinds[MAX_BLINDS] = {BLIND_CURRENT, BLIND_UPCOMING, BLIND_UPCOMING}; // The current state of the blinds, this is used to determine what the game is doing at any given time
 
+static int blind_reward = 0;
+static int hand_reward = 0;
+static int interest_reward = 0;
```
ricfehr3, line 192: Definitely something for later is encapsulating things into structs. Not necessary for this PR, but something to think of going forward and taking opportunities to fix when possible.

source/game.c
```diff
@@ -520,16 +765,16 @@ void change_background(int id)
...
-        memset16(&pal_bg_mem[8], 0xFFFF, 1);
+        memset16(&pal_bg_mem[SHOP_LIGHTS_2], 0x32BE, 1); // Reset the shop lights to correct colors
+        memset16(&pal_bg_mem[SHOP_LIGHTS_3], 0x4B5F, 1);
+        memset16(&pal_bg_mem[SHOP_LIGHTS_4], 0x5F9F, 1);
+        memset16(&pal_bg_mem[SHOP_LIGHTS_1], 0xFFFF, 1);
 
-        memcpy16(&pal_bg_mem[7], &pal_bg_mem[3], 1); // Disable the button highlight colors
-        memcpy16(&pal_bg_mem[5], &pal_bg_mem[16], 1);
+        memcpy16(&pal_bg_mem[REROLL_BTN_SELECTED_BORDER], &pal_bg_mem[REROLL_BTN], 1); // Disable the button highlight colors
+        memcpy16(&pal_bg_mem[NEXT_ROUND_BTN_SELECTED_BORDER], &pal_bg_mem[NEXT_ROUND_BTN], 1);
```
ricfehr3: memcpy16(&pal_bg_mem[<offset>], <some u16>, 1); 

Is used quite a bit. I think it would make sense to have an inline wrapper, something like, `set_bg(_UIElementPID pal_id, u16 val)`

Also here, it may make more sense to pull out setting the shop "scene" as `void set_shop_scene(...)`, to set all these lights and borders. 

This doesn't need to be done now. I think what you have is great and additional diffs will get heavier to integrate.

source/game.c
```diff
@@ -2677,211 +2812,203 @@ void game_shop()
...
         }
-        default:
-            state = 0;
-            timer = 0;
-            selection_y = 0;
-            background = UNDEFINED;
-            game_set_state(GAME_PLAYING);
-            break;
+    
+        int y = 6;
```
ricfehr3: Not sure what y is here

ricfehr3 (changes requested):
Looks great! This is a big step up. I'm probably going to keep coming back and checking stuff more, it's a lot to diff. But yeah, this is awesome.

drewjohnson2:
It'll be a lot better once the magic numbers PR is merged and I can rebase onto that, but it'll still be hefty.

In retrospect I probably could have broken it up a little. ¯\_(ツ)_/¯

source/game.c
drewjohnson2: I have no clue either. This was just used in a few different places, so I thought I'd move it out into its own function

source/game.c
drewjohnson2: ¯\_(ツ)_/¯

Probably have to do a git blame on the original section of code. 

If you glance at the magic numbers  PR, which this is based on, you'll see I couldn't figure it out there either.

ricfehr3:
> It'll be a lot better once the magic numbers PR is merged and I can rebase onto that, but it'll still be hefty.
> 
> In retrospect I probably could have broken it up a little. ¯_(ツ)_/¯

Oh yeah, my bad I should have read what you said there. 

All good on big PR's. I forget what book it was in exactly, but it said something like, "A kernel developer isn't afraid to completely re-write a system from scratch if it makes more sense." Not the exact quote, regardless I agree with it. Big diffs are worth it for big positive changes.

Sorry for poppin' in early. I'll let it cook before I come back for a review :sunglasses:

include/game.h
drewjohnson2: I 100% agree. I had that thought at some point, but I think it got wiped from my memory at some point when I was removing all of those substate switch statements.

source/game.c
drewjohnson2: The probably should be moved to another file, which is what i pursued until I realized:

All of these functions rely on file scope variables. In the future when we have these variables added to some sort of _superstructure_ we can then move those functions to another file and have each function take a pointer to that structure.

But until then we're stuck with this.

source/game.c
drewjohnson2: I agree with this 100% and had thought about it, but it was just too many tedious changes for me to make for this issue.

source/game.c
drewjohnson2: After doing some investigation it appears that the name of this function is not correct.

So I have no clue what it's doing outside of copying some tiles into screen entries on lines 307-311 and line 312 is a horizontal flip and setting a tile.

source/game.c
```diff
@@ -135,6 +299,20 @@ static inline Card *discard_pop()
     return discard_pile[discard_top--];
 }
 
+// Resets bg tiles after shop or blind panel is dismissed
+// to match the rest of the game panel background.
+static inline void reset_bg_on_panel_dismissed()
+{
```
drewjohnson2, line 280: Turns out this code just resets the background of a small section on the left hand game panel after either the blind panel or shop panel dismisses.

source/game.c
```diff
@@ -135,6 +299,20 @@ static inline Card *discard_pop()
...
+    int y = 6;
+
+    memset16(&se_mem[MAIN_BG_SBB][32 * (y - 1)], 0x0006, 1);
+    memset16(&se_mem[MAIN_BG_SBB][1 + 32 * (y - 1)], 0x0007, 2);
+    memset16(&se_mem[MAIN_BG_SBB][3 + 32 * (y - 1)], 0x0008, 1);
+    memset16(&se_mem[MAIN_BG_SBB][4 + 32 * (y - 1)], 0x0009, 3);
```
drewjohnson2, line 286: previously the halfword count passed in on this line was 4, which unless I have a misunderstanding (which is likely) is incorrect. The correct halfword count to fill SE 4, 5 and 6 is 3. I may be misunderstanding something though.

source/game.c
ricfehr3: Oh yup, that makes sense. I can make it a new issue and reference this one.

source/game.c
ricfehr3: Another good one to make an issue for :) Later though, so it can be made with the context of the work finished in this pr.

source/game.c
ricfehr3: Done: https://github.com/cellos51/balatro-gba/issues/156

include/game.h
```diff
@@ -26,16 +26,13 @@
...
-    GAME_SPLASH_SCREEN,
-    GAME_MAIN_MENU,
-    GAME_PLAYING,
-    GAME_ROUND_END,
-    GAME_SHOP,
-    GAME_BLIND_SELECT,
-    GAME_LOSE,
-    GAME_WIN
+#define DEF_STATE_INFO(stateEnum, on_init, on_update, on_exit) stateEnum,
+#include "../include/def_state_info_table.h"
```
ricfehr3: Two things here:

1. I appreciate the detail of the full include path, but we can drop the `../include/` part. Thoughts?

2. Since we now have the def files for `joker_gfxs`,`pools`, and now `states` we could put them in their own directory `include/def` or similar?

source/game.c
```diff
@@ -23,22 +22,184 @@
...
 #include "background_main_menu_gfx.h"
-#include "affine_main_menu_background_gfx.h"
 
 #include "soundbank.h"
 
 #include "list.h"
 
+// This could go in game.h or some other header somewhere
+// but it's essentially private to this file, so might as well
+// leave it.
```
ricfehr3: Agreed

source/game.c
```diff
@@ -23,22 +22,184 @@
...
+    char *display_name;
+} HandValues;
+
+typedef void (*SubStateActionFn)();
+
+// Used as a No Operation for game states that have no init and/or exit function.
+// ricfehr3 did the work of determining whether a noop or a NULL check was more 
+// efficient. Well, this is the answer.
+// Thanks!
+// https://github.com/cellos51/balatro-gba/issues/137#issuecomment-3322485129
```
ricfehr3, line 74: :D

source/game.c
```diff
@@ -1419,8 +1574,8 @@ void card_in_hand_loop_handle_discard_and_shuffling(int card_idx, bool* discarde
     {
         if (!*discarded_card)
         {
-            *hand_x = int2fx(240);
-            *hand_y = int2fx(70);
+            *hand_x = int2fx(CARD_DISCARD_PNT_X);
+            *hand_y = int2fx(CARD_DISCARD_PNT_Y);
```
ricfehr3: Not sure if these are used elsewhere, but you could embed the `int2fx` in the macro itself?

Same for `HAND_START_POS_X` on 1640 below.

ricfehr3 (commented):
No real changes here, it looks great. Just some food for thought and opinions, but it looks really good.

include/game.h
drewjohnson2: That seems like a good idea to me.  You want to add those directories in your PRs or do you want me to add them in mine? It honestly doesn't matter, just depends on how soon you'd want it merged into master. Your PRs _should_ be merged before mine. But again, it doesn't matter at all

source/game.c
drewjohnson2: Probably not a bad idea, I can do that. That won't go into this PR though since this change is actually from the magic numbers PR.

In hindsight I think me opening a PR based off of a branch that's also in PR wasn't the most helpful.

Most changes like this in this PR are not actually from this PR.

ricfehr3: It's all good. Nothing a few git headaches can't fix haha.

include/game.h
ricfehr3: Why don't we save it until after all these other ones are merged. No need to add more git headaches with a small change.

MeirGavish:
I'm requesting a review from Copilot to help me manage the more complex PRs, and I will review them myself later on, take his review with a grain of salt, I will reply to his comments with my opinion when I get to it, but it can be helpful to spot mistakes.

source/splash_screen.c
```diff
@@ -6,16 +6,19 @@
...
 static const Rect COUNTDOWN_TIMER_RECT = {208, 144, 240, 152};
+static uint timer = 0;
 
-void splash_screen_init()
+void splash_screen_on_init()
 {
     tte_printf("#{P:72,8; cx:0xF000}DISCLAIMER");
     tte_printf("#{P:8,24; cx:0xF000}This project is NOT endorsed \n by or affiliated with \n Playstack or LocalThunk.\n\n If you have paid for this, \n you have been scammed \n and should request a refund \n IMMEDIATELY. \n\n The only official place \n to obtain this is from: \n\n 'github.com/\n  cellos51/balatro-gba'");
     tte_printf("#{P:8,144; cx:0xF000}(Press any key to skip)");
 }
```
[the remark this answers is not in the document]
drewjohnson2: Seems reasonable. 

Note: I did not change functionality, I just changed the way it's accessed. I don't mind making the change though, since it's small, but it is outside of the scope of this PR.

source/game.c
```diff
@@ -752,79 +1002,13 @@ void set_hand()
...
-        mult = 0;
-        break;
-    }
 
+    HandValues hand = hand_base_values[hand_type];
+
+    chips = hand.chips;
+    mult = hand.mult;
+
+    print_hand_type(hand.display_name);
```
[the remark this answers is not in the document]
drewjohnson2: Reasonable. There's no reason that `display_name` should ever be null, but there's nothing wrong with defensive programming (unless this check adds an obscene amount of code when compiled).

source/game.c
```diff
@@ -135,6 +300,20 @@ static inline Card *discard_pop()
...
+// Resets bg tiles after shop or blind panel is dismissed
+// to match the rest of the game panel background.
+static inline void reset_bg_on_panel_dismissed()
+{
+    int y = 6;
+
+    memset16(&se_mem[MAIN_BG_SBB][32 * (y - 1)], 0x0006, 1);
+    memset16(&se_mem[MAIN_BG_SBB][1 + 32 * (y - 1)], 0x0007, 2);
+    memset16(&se_mem[MAIN_BG_SBB][3 + 32 * (y - 1)], 0x0008, 1);
+    memset16(&se_mem[MAIN_BG_SBB][4 + 32 * (y - 1)], 0x0009, 3);
```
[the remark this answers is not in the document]
drewjohnson2: @MeirGavish I'll definitely want you to look into this one for me. I left a comment on this line specifically explaining this change. This change may be incorrect, but I have a feeling this change is right.

include/game.h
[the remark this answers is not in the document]
drewjohnson2: This one can absolutely be ignored.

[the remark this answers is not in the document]
MeirGavish: Doesn't `#include "def_state_info_table.h"` work though?

source/game.c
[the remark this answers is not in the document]
MeirGavish: @drewjohnson2 
Yeah, I think you're right, the next line just overwrites the last tile from this one.

source/game.c
```diff
@@ -166,6 +267,20 @@ static inline Card *discard_pop()
     return discard_pile[discard_top--];
 }
 
+// Resets bg tiles after shop or blind panel is dismissed
+// to match the rest of the game panel background.
+static inline void reset_bg_on_panel_dismissed()
```
```suggestion
// Resets bottom row bg tiles of the top left panel (shop/blind) after 
// it is dismissed to match the rest of the game panel background.
static inline void reset_top_left_panel_bottom_row()
```

Just a name and documentation change suggestion to make this clearer, I will change the calls too. Nice job figuring out what this does! I wish it was documented before.

I will add suggestions at the call sites too.

source/game.c
```diff
@@ -2126,271 +2145,257 @@ void game_playing()
...
-            {
-                tte_erase_rect_wrapper(BLIND_REWARD_RECT);
-                tte_erase_rect_wrapper(BLIND_REQ_TEXT_RECT);
-                obj_hide(playing_blind_token->obj);
-                affine_background_load_palette(affine_background_gfxPal);
-                state = BLIND_PANEL_EXIT;
-                timer = TM_ZERO;
-            }
-            break;
+            reset_bg_on_panel_dismissed();
```
```suggestion
            reset_top_left_panel_bottom_row();
```

source/game.c
```diff
@@ -2777,13 +2782,7 @@ static void game_shop_outro()
...
         }
 
-        int y = 6;
-        memset16(&se_mat[MAIN_BG_SBB][y - 1][0], 0x0006, 1);
-        memset16(&se_mat[MAIN_BG_SBB][y - 1][1], 0x0007, 2);
-        memset16(&se_mat[MAIN_BG_SBB][y - 1][3], 0x0008, 1);
-        memset16(&se_mat[MAIN_BG_SBB][y - 1][4], 0x0009, 4);
-        memset16(&se_mat[MAIN_BG_SBB][y - 1][7], 0x000A, 1);
-        memset16(&se_mat[MAIN_BG_SBB][y - 1][8], SE_HFLIP | 0x0006, 1);
+        reset_bg_on_panel_dismissed();
```
```suggestion
        reset_top_left_panel_bottom_row();
```

source/game.c
```diff
@@ -2821,211 +2820,200 @@ void game_shop()
...
         }
-        default:
-            state = 0;
-            timer = TM_ZERO;
-            selection_y = 0;
-            background = UNDEFINED;
-            game_set_state(GAME_PLAYING);
-            break;
+    
+        reset_bg_on_panel_dismissed();
```
```suggestion
        reset_top_left_panel_bottom_row();
```

source/splash_screen.c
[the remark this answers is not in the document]
MeirGavish: I agree with Copilot here, the state shouldn't have it's own separate timer while the others use the global one. You can add a get_timer() function to access it rather than pass it as a parameter.
This theoretically would change functionality if we return to the splash screen though in practice it's not supposed to happen.

source/game.c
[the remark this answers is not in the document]
MeirGavish: There's actually a NULL-check in `print_hand_type()` (which I labelled paranoia but it's a healthy one) so there's no point adding this check, it is redundant.

source/game.c
MeirGavish: I'm okay with the forward declarations, when the game states are separated to files, they will also be separated.

source/game.c
MeirGavish: New game states would probably need to be added for implementing booster packs, or for a settings menu.

source/game.c
MeirGavish: Does it work though? Does the game properly reset if you change the state after loss? 
Also it should be renamed `game_over_on_exit()` and then be used for both game_win and game_lose

source/game.c
```diff
@@ -3136,9 +3124,66 @@ static void game_lose()
...
+    current_blind = SMALL_BLIND;
+    blinds[0] = BLIND_CURRENT;
+    blinds[1] = BLIND_UPCOMING;
+    blinds[2] = BLIND_UPCOMING;
+    round = 0; 
+    ante = 1;
+    money = 4;
+    score = 0;
+    hands = max_hands;
+    discards = max_discards;
```
MeirGavish: I know you said this function is not set in stone but since this is already done in `game_init()` it shouldn't be here too.

source/game.c
```diff
@@ -3136,9 +3124,66 @@ static void game_lose()
...
+{
+    timer = TM_ZERO;
+    current_blind = SMALL_BLIND;
+    blinds[0] = BLIND_CURRENT;
+    blinds[1] = BLIND_UPCOMING;
+    blinds[2] = BLIND_UPCOMING;
+    round = 0; 
+    ante = 1;
+    money = 4;
+    score = 0;
```
MeirGavish: I think these should be moved to `game_init()` actually... This can wait until a proper restart is implemented.

source/game.c
```diff
@@ -3136,9 +3124,66 @@ static void game_lose()
...
+    display_money(money);
+    tte_printf(
+        "#{P:%d,%d; cx:0x%X000}%d#{cx:0x%X000}/%d",
+        ANTE_TEXT_RECT.left,
+        ANTE_TEXT_RECT.top,
+        TTE_YELLOW_PB,
+        ante,
+        TTE_WHITE_PB,
+        MAX_ANTE
+    ); // Ante
```
MeirGavish, line 3221: Since this is already done in `game_start()` I don't think it should also be done here. A proper restart is outside the scope of this PR but I guess if it were "New Run" then it would call `game_start()` too and if it were "Main Menu" it would let the main menu handle it. I guess you can add a call to `game_start()` after `game_init()` here for now and it might work.

source/game.c
```diff
@@ -2126,271 +2145,257 @@ void game_playing()
     game_playing_ui_text_update();
 }
 
-void game_round_end_cleanup()
+static void game_round_end_on_exit()
 {
     // Cleanup blind tokens from this round to avoid accumulating 
     // allocated blind sprites each round
+    timer = TM_ZERO;
+    state_info[game_state].substate = 0;
```
MeirGavish: I think it would be better to set this on `game_change_state()` so every state starts at substate 0 rather than have every state do the reset.

source/game.c
```diff
@@ -2821,211 +2820,200 @@ void game_shop()
...
+        BG_POINT to = {0, y_to};
+    
+        main_bg_se_copy_rect(from, to);
+    }
+}
+
+static void game_blind_select_on_exit()
+{
+    state_info[game_state].substate = 0;
+    timer = 0;
```
MeirGavish: I'm pretty sure the timer reset can be removed from here (and many other places) now. Not critical for this PR.

source/game.c
```diff
@@ -3136,9 +3124,66 @@ static void game_lose()
...
+        "#{P:%d,%d; cx:0x%X000}%d#{cx:0x%X000}/%d",
+        ANTE_TEXT_RECT.left,
+        ANTE_TEXT_RECT.top,
+        TTE_YELLOW_PB,
+        ante,
+        TTE_WHITE_PB,
+        MAX_ANTE
+    ); // Ante
+
+    affine_background_load_palette(affine_background_gfxPal);
```
MeirGavish, line 3223: Is this because we're in the boss blind? This should also probably be done elsewhere.

include/game.h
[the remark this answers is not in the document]
drewjohnson2: Turns out I'm a fool. I thought this was a comment on `game.c`, not `game.h`.

MeirGavish:
Oh, and also hand_state and play_state also act as sort of substate variables for the GAME_STATE_PLAYING state so theoretically they can be refactored into this system but it's quite a bit to untangle so I don't know if it's worth it.

source/game.c
drewjohnson2: @MeirGavish Doing this causes an endless loop, it appears, and causes all printed values on screen to be erased and redrawn constantly.

I can look into it more later, but this request, at least as stated in this comment, is not possible.

source/game.c
drewjohnson2: @MeirGavish No, this is done because the palette is swapped when you get a game over. This is, as far as I can tell, the only way to reset the palette back to the normal game palette.

We could add it elsewhere, but if we add it to some sort of startup function we'll be adding an unnecessary function call to the stack since when the game starts this palette is already active. Personally, I think this function call fits in best here.

MeirGavish: Right, yeah, I agree.

source/game.c
MeirGavish: Odd, I guess this can wait for a proper restart. I also now see that game_start() does a few things that would make it problematic to call from other places. For one it destroys the main menu ace, which should be done on the main menu's on_exit().

source/game.c
drewjohnson2: It does indeed work and everything appears to properly reset after a loss.

drewjohnson2:
@MeirGavish I think I've addressed all of your concerns.

As far as merge order goes, and obviously I'm biased in this matter, but I think merging this PR might be the simplest solution. Any rebase woes on other branches _should_ be easy now since the state machine is, in my opinion, simplified. But it doesn't matter too terribly much to me.

> Oh, and also hand_state and play_state also act as sort of substate variables for the GAME_STATE_PLAYING state so theoretically they can be refactored into this system but it's quite a bit to untangle so I don't know if it's worth it.

Since there's, at a glance, a non-trivial amount of investigation that I'd need to do to incorporate that, I can probably add it in a follow up PR. This change set is already bloated enough, imo.

MeirGavish:
> > Oh, and also hand_state and play_state also act as sort of substate variables for the GAME_STATE_PLAYING state so theoretically they can be refactored into this system but it's quite a bit to untangle so I don't know if it's worth it.
> 
> Since there's, at a glance, a non-trivial amount of investigation that I'd need to do to incorporate that, I can probably add it in a follow up PR. This change set is already bloated enough, imo.

Of course, not for this PR, just noting it for possible future refactors.

As another such note, the code tends to reset things a lot on exit (previously on the default case), doing the same thing in many places, and it will be better to centralize these and also move towards each state initializing what it needs on entrance instead.

source/splash_screen.c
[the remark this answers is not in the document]
MeirGavish: So now the state has its own timer but resets it on init, I'll take it, might be even better to do the same for the other states once they are separated to files.

source/game.c
```diff
@@ -3177,9 +3160,64 @@ static void game_lose()
     {
         tte_printf("#{P:%d,%d; cx:0x%X000}GAME OVER", GAME_LOSE_MSG_TEXT_RECT.left, GAME_LOSE_MSG_TEXT_RECT.top, TTE_RED_PB);
     }
+
+    if (key_hit(KEY_ANY)) game_change_state(GAME_STATE_BLIND_SELECT);
```
MeirGavish, line 3178: Did you intend to have a call to actually trigger the reset? I'm okay with leaving this since it works but I think `game_over_on_exit()` should be cleaned up a bit then and it should also be added for `game_win_on_update()`.

source/game.c
```diff
@@ -3177,9 +3160,64 @@ static void game_lose()
...
+        ANTE_TEXT_RECT.left,
+        ANTE_TEXT_RECT.top,
+        TTE_YELLOW_PB,
+        ante,
+        TTE_WHITE_PB,
+        MAX_ANTE
+    ); // Ante
+
+    affine_background_load_palette(affine_background_gfxPal);
+    game_init();
```
```suggestion
    list_destroy(&jokers_available_to_shop);
    
    game_init();
    
    display_round(round);
    display_score(score);
    display_chips(chips);
    display_mult(mult);
    display_hands(hands);
    display_discards(discards);
    display_money(money);
    tte_printf(
        "#{P:%d,%d; cx:0x%X000}%d#{cx:0x%X000}/%d",
        ANTE_TEXT_RECT.left,
        ANTE_TEXT_RECT.top,
        TTE_YELLOW_PB,
        ante,
        TTE_WHITE_PB,
        MAX_ANTE
    ); // Ante

    affine_background_load_palette(affine_background_gfxPal);

```

Since we rely on hand and discard initialization on game_init() it should be moved so initialization happens before display.

source/game.c
```diff
@@ -3177,9 +3160,64 @@ static void game_lose()
...
+{
+    timer = TM_ZERO;
+    current_blind = BLIND_TYPE_SMALL;
+    blinds[0] = BLIND_STATE_CURRENT;
+    blinds[1] = BLIND_STATE_UPCOMING;
+    blinds[2] = BLIND_STATE_UPCOMING;
+    round = 0; 
+    ante = 1;
+    money = 4;
+    score = 0;
```
MeirGavish: And I will insist on moving this to `game_init()`, also I think money shouldn't use a magic number, I guess you can add a `starting_money` variable along with `max_hands` and `max_discards` to make it consistent.

source/game.c
drewjohnson2: I mean, yeah. There has to be some sort of a call for the thing to reset.

It's already added for `game_win_on_update()`.

source/game.c
drewjohnson2: I don't mind doing this, but this is another instance of duplicating behavior on game startup that doesn't need to be duplicated. The reason it was kept in the game over function was because these values are already set when the game starts.

Fortunately, unnecessarily setting values on startup doesn't really kill performance.

source/game.c
drewjohnson2: Added this, but it doesn't appear to have made any difference. I'll keep it in since it appears not to have mattered.

source/game.c
MeirGavish: Yes, I prefer this, I think it is the correct place to set it rather than at the global definition. Theoretically if we allow for the player to choose different decks with different starting money, it will be the place to change.

source/game.c
MeirGavish: Okay, cool, I just wasn't sure since you said the function was a placeholder.

source/game.c
MeirGavish: I think I did see wrong values for hands (0) and discards on restart.

drewjohnson2: I may have corrected it and forgot, but it looks like those values are good now.

drewjohnson2:
@MeirGavish Okay, I think that now it's probably good to merge (depending on what you decide to do with merge order).

drewjohnson2:
Yeah, it's going to be nasty and I do not look forward to it even a bit.

MeirGavish:
I kinda want this in, I'll merge, conflicts can be dealt with when the time comes.

Issue #163, opened 2025-10-08
Game can be softlocked if hand is played and selection is cancelled at the same time


MeirGavish:
Steps to reproduce:

During round,
1. Select hand cards
2. Highlight "Play Hand" button
3. Press **both** A and B (select button and cancel hand selection)
4. The game state will change to playing hand state but the selection is also cancelled so no hand can be played, the game is then softlocked.

https://github.com/user-attachments/assets/43159e25-f991-4edb-8975-7a7bffed030b

ricfehr3:
I think I found the issue and addressed it in the PR. Graceful state transitions can be another battle if you just wanted the quick fix.

PR #164, opened 2025-10-09
Fix softlock related to #163

ricfehr3:
Quick fix for #163


source/game.c
ricfehr3, line 1228: Looks like it's reset right here on the A+B in one frame situation.

MeirGavish:
Nice, it looks like a simple 2-row fix but the code is quite convoluted so finding this isn't trivial.

Issue #151, opened 2025-09-23
Update List Implementation


ricfehr3:
🛑 Blocked on: https://github.com/cellos51/balatro-gba/pull/132

Once memory pools are implemented. The list implementation can be modified to utilize the array in the memory pool. 

Currently the `List` implementation in `list.h/c` uses a pointer to an array to hold it's data. A few malloc/realloc calls are required currently. There is also a `int` wrapper provided. I think it can be reduced to one implementation with no data copies.

Since objects are pooled into an array now. References to objects in this pool now only require an `int` to the index in the pool's array. This can allow the List to be redesigned to use integers as offsets into the pool array. Reducing the complexity needed to store a list as simply a list of ints for each type. 

I suggest we borrow an element from the [linux kernel linked list](https://kernelnewbies.org/FAQ/LinkedLists) design. Instead of creating a list object that tracks a list in an array, a `struct list_head` is saved in struct. This `struct list_head` would be as small as possible:

```c
struct {
    int next;
    int prev;
} list_head;
```

Where `next` and `prev` are integer offsets into the memory pool's array of the same type.

Eventually the access pattern would maybe look like this:

```c
// here '-1' is NULL for the int offset
// Using "Joker" object for the example, assume that it now has a "list_head" member in the struct
Joker* current = get_joker_list_head(); // this function doesn't exist, yet...
while(current->list.next > 0)
{
    Joker* next = POOL_AT(current->list.next); // New macro along with other POOL_ macros
    current = next;
}
```

Since the list is essentially typeless now. It's extendible to anything that uses a pool, which should be the only objects that need a list anyway.

There's other ways to handle this as well. Another idea is to use pointers to the addresses in the pool directly. Though type safety becomes and issue, as well as NULL pointers.

I'm going to tackle this after the mentioned PR is merged, unblocking this.

ricfehr3:
I'm on the fence about this design. I'm still thinking through it. 

The thing is, it would require a macro to use to type to the correct pool. I'm totally down for that (it's what linux does with `container_of` but I'm still thinking it through. Free forum for anyone to add to.

ricfehr3:
Putting the `list_head` in the struct is not very helpful to us. An object can only be in one list at a time. 

Instead. I think I'll use `ListNode` structs in a pool. Then with some macros to find quick offsets via the pool.

[...]

Issue #158, opened 2025-09-30
Add swapping order of Cards and Jokers


MathisMartin31:
Quick idea I had, since the implementation for Lists is going to be reworked as mentioned by issue #151 and it looks like it will add some flexibility.

Some Jokers rely on the order the Cards and Jokers themselves are in, but I haven't found a way to swap them around so I assume it's not implemented yet.

As of now, you can sell/select Cards/Jokers by pressing A. What I'm proposing is that we keep this behavior, but trigger the action on button release with the `key_released()` function, and allow moving the Card/Joker while the A button is held (`key_held()`) by pressing the Left/Right arrows. When doing this, the action would switch from "selling" to "moving". The action would then still trigger when A is released, rearranging the Card/Joker in the list by swapping their next/prev properties around as one would normally do with a linked list.

I wanted to look at it but I think waiting for the List implementation to drop would be best.

MeirGavish:
This would be good to add, especially since more xmult jokers are being added and even more so since it seems like Blueprint and Brainstorm will be added.
Jokers are sold with the L button, not the A, so A can be used to rearrange them.
I for the Joker selling mechanic I added the selection_grid system (which honestly I'm not fully happy with but it generalizes selection at least), it is currently used only by the shop and not by the game round state. The game round should be refactored to use it and then the same code can be used for both the shop and game state. Then joker rearranging can be implemented and jokers can be both sold and rearranged in both shop and round.

Hand card rearranging wouldn't affect much currently so I guess it can be postponed for now but I think we can consolidate selection and rearranging, we need to see how it is handled in Balatro with a controller.

MathisMartin31:
Yes selling with L completely went over my head lol ^^'. And I think you're right about rearranging cards, it can be taken care of later.

I think the selection system being used only in the Shop isn't much of an issue right now if we use it to rearrange Jokers. We can add the rearranging with it now, and refactor the round selection to use it later.

Imo this could even be a "feature": think twice before going to the next round because you won't be able to change the Joker order during the round. (I may or may not have played like this for several months before realizing changing the order on the fly was a thing 👀).

Also, I played some Balatro today and it behaves as I proposed: hold A to keep a card selected, then press left/right to move it.

noctowlhoot:
Just played the latest build and glad to see this has already been mentioned. And I agree with Martin about only being able to rearrange in the shop as being a feature. With a smaller joker roster, adds another layer of strategy/difficulty to the game

[...]

PR #152, opened 2025-09-23
Weight jokers by rarity in the shop

JustinL-12:
From the Balatro wiki, when jokers are generated there are specific chances to get one of each rarity:
image

These changes implement this behavior for the jokers generated in the shop. The new joker function joker_get_random_rarity can also be used in the future for when jokers get generated by other means.

PR #168, opened 2025-10-12
Update list implementation to work with memory pools

ricfehr3:
For issue: #151 
:stop_sign: Blocked on PR #132 

PR for updating list implementation to work with the new memory pools. This completely removes malloc from the codebase.

This is blocked on the memory pool PR, and this will be in WIP (needs cleanup and I'd like to test more). However, I think it's at a pretty good spot right now, so I'll get the PR going so others can check it out.

It is a pretty confusing thing to digest. I'm working on more diagrams to explain the entire memory system with the pools and how that relates to lists, but below should explain how this list implementation works a bit. 

Listimplementation

### DIFF
--- a/include/list.h
+++ b/include/list.h
@@ -85,9 +85,8 @@ typedef struct
 /**
  * Create a list.
  *
- * If you are recreating and existing list, the list must be freed with @ref list_clear.
- *
  * While this function does not allocate memory for the list itself, the list does allocate memory for each element.
+ * So every created list must be freed with @ref list_clear to ensure the list's nodes are deleted properly.
  *
  * @return A @ref List with head and tail reset.
  */
@@ -176,7 +175,10 @@ ListItr list_itr_create(List* list);
 void* list_itr_next(ListItr* itr);
 
 /**
- * Remove the previous @ListNode from the iterator
+ * Remove the current @ref ListNode from the iterator.
+ *
+ * The "current node" corresponds to the list node associated with the
+ * most recently returned valu from @ref list_itr_next()
  *
  * @param itr pointer to the @ref ListItr
  */
--- a/source/game.c
+++ b/source/game.c
@@ -422,10 +422,10 @@ void add_joker(JokerObject *joker_object)
     }
 }
 
-void remove_held_joker(JokerObject* joker_object, int shop_joker_idx)
+void remove_owned_joker(int owned_joker_idx)
 {
     // TODO: Extract to on_joker_removed() callback
-    //JokerObject* joker_object = list_get_at_idx(&_owned_jokers_list, joker_id);
+    JokerObject* joker_object = list_get_at_idx(&_owned_jokers_list, owned_joker_idx);
     // In case the player gets multiple Four Fingers Jokers,
     // and only reset the size when all of them have been removed
     if (joker_object->joker->id == FOUR_FINGERS_JOKER_ID) 
@@ -443,7 +443,7 @@ void remove_held_joker(JokerObject* joker_object, int shop_joker_idx)
     }
 
     _set_shop_joker_avail(joker_object->joker->id, true);
-    list_remove_at_idx(&_owned_jokers_list, shop_joker_idx);
+    list_remove_at_idx(&_owned_jokers_list, owned_joker_idx);
 }
 
 int get_deck_top(void)
@@ -1422,6 +1422,7 @@ void game_init()
     _owned_jokers_list = list_create();
     _discarded_jokers_list = list_create();
     _shop_jokers_list = list_create();
+    // TODO: Move this to an initialization of the play scoring states
     _joker_scored_itr = list_itr_create(&_owned_jokers_list);
 
     jokers_available_to_shop_init();
@@ -2963,7 +2964,7 @@ void game_sell_joker(int joker_idx)
     display_money(money);
     erase_price_under_sprite_object(joker_object->sprite_object);
 
-    remove_held_joker(joker_object, joker_idx);
+    remove_owned_joker(joker_idx);
 
     joker_start_discard_animation(joker_object);
 }
--- a/source/list.c
+++ b/source/list.c
@@ -206,6 +206,7 @@ static ListNode* _list_itr_node_next(ListItr* itr)
 
 void list_itr_remove_current_node(ListItr* itr)
 {
+    if(!itr || !itr->current_node) return;
     ListNode* tmp_prev = itr->current_node->prev;
     _list_remove_node(itr->list, itr->current_node);
     itr->current_node = tmp_prev;


source/game.c
```diff
@@ -188,14 +189,14 @@ int get_played_top(void) {
     return played_top;
 }
 
-List *get_jokers(void) {
-    return jokers;
+ListHead get_jokers_list(void) {
```
ricfehr3: return a pointer here, callers should be able to modify the list.

source/game.c
```diff
@@ -188,14 +189,14 @@ int get_played_top(void) {
...
+    return _all_jokers_list;
 }
 
 bool is_joker_owned(int joker_id) {
-    for (int k = 0; k < list_get_size(jokers); k++)
+    for (int k = 0; k < list_get_len(_all_jokers_list); k++)
```
ricfehr3: bad rebase, this should be an iterator now.

source/game.c
```diff
@@ -1224,11 +1220,13 @@ void game_set_state(enum GameState new_game_state)
 void jokers_available_to_shop_init()
 {
     int num_defined_jokers = get_joker_registry_size();
-    jokers_available_to_shop = list_new(num_defined_jokers);
+
+    _avail_shop_jokers_list = list_new();
```
ricfehr3: Add list delete above to parse the list and free each node.

source/game.c
```diff
@@ -1237,11 +1235,8 @@ void game_init()
     jokers_available_to_shop_init();
 
     // Initialize jokers list
-    if (jokers) list_destroy(&jokers);
-    jokers = list_new(MAX_JOKERS_HELD_SIZE);
-
-    if (discarded_jokers != NULL) list_destroy(&discarded_jokers);
-    discarded_jokers = list_new(MAX_JOKERS_HELD_SIZE);
+    _all_jokers_list = list_new();
```
ricfehr3: Add list delete above to parse the list and free each node.

source/game.c
```diff
@@ -1259,7 +1254,8 @@ void game_init()
 
 void game_start()
 {
-    set_seed(rng_seed);
+    //set_seed(rng_seed);
+    set_seed(9); // 9 is a full house
```
ricfehr3: Remove, but for anyone who sees this before I do, 9 is always a full house as the first hand, and it's 3 aces and 2 jacks so you beat the first hand in one go.

source/game.c
```diff
@@ -2493,23 +2498,31 @@ static void game_shop_reroll(int *reroll_cost)
 {
     money -= *reroll_cost;
     display_money(money); // Update the money display
-    for (int i = 0; i < list_get_size(shop_jokers); i++)
+
+    ListItr itr = list_itr_new(&_all_jokers_list);
```
ricfehr3: `itr` is a bad name, fix it for all others

source/list.c
```diff
@@ -2,100 +2,181 @@
...
+    return ln;
 }
 
-void* list_get(List *list, int index)
+void list_remove_at_idx(ListHead* p_list, int elem_idx)
 {
-    if (index < 0 || index >= list->size) 
-        return NULL;
-    return list->_array[index];
+    int len = 0;
```
ricfehr3, line 156: You'll see the use of this new iterator scattered about this PR. The only way I could think to simplify re-use of this code is maybe like the `list_for_each` macro in [linux](https://docs.kernel.org/core-api/list.html#traversing-the-list), but that's too much imho.

include/list.h
```diff
@@ -2,24 +2,40 @@
...
+} ListHead;
+
+typedef struct
+{
+    const ListHead *p_list;
+    ListNode *p_current_node;
+} ListItr;
+
+// add push back
+// add iterator to fix while loop
```
ricfehr3: Delete

ricfehr3:
@MeirGavish Squashed all my commits and rebased this one. 

One question. I updated the change from #169 . I think it's working, but just to confirm, **only the one joker your testing should appear?**

MeirGavish:
My intention was that if you don't own it, it would always appear in the shop but other jokers will continue appearing normally

drewjohnson2:
I appreciate the chart immensely.

include/def_balatro_mempool.h
```diff
@@ -1,10 +1,12 @@
...
 #include "card.h"
+#include "list.h"
 
 POOL_ENTRY(Sprite, MAX_SPRITES);
 POOL_ENTRY(SpriteObject, MAX_SPRITE_OBJECTS);
 POOL_ENTRY(Joker, MAX_ACTIVE_JOKERS);
 POOL_ENTRY(JokerObject, MAX_ACTIVE_JOKERS);
 POOL_ENTRY(Card, MAX_CARDS);
 POOL_ENTRY(CardObject, MAX_CARDS_ON_SCREEN);
+POOL_ENTRY(ListNode, 128);
```
drewjohnson2: I'm sure you're going to do this, especially since this is just a draft, but thought I'd throw a reminder in here:

Don't forget to define a macro for the pool capacity on this line.

Couldn't find anything else to nitpick. Really good stuff in this PR.

ricfehr3:
@drewjohnson2 Thanks for checking this out. I took a real long hard look at this and decided to do some changes to what I had. Still working on cleaning this up, but writing it here to just log it in detail anyway.

So, I realized that using integer offsets wasn't the vibe:
1. It offered nothing in terms of saving space, `sizeof(int) == sizeof(void*)`
2. It only abstracted data types specifically for objects that are pooled. You still have to cast the result anyway too.
3. A pointer to the data works better since it decouples the list from the other pools entirely. It also limits the interface to the pool.
4. Accessing via the offsets was cumbersome and would have required macros to cleanup.

Now, with this commit as of this comment. I have used a union of an `int` and `void*` so we can support the original **list** and **int list**. However, I think I'm going to remove that, unless it is helpful elsewhere, because I don't think we need to have the **int list** anymore. Why? Because we can use the pool's bitmap tracker to do the same thing but faster. This would also limit the use of our largest list. 

Going to clean this up more and switch to the bitmap stuff.

MeirGavish:
> So, I realized that using integer offsets wasn't the vibe:
> 
> 1. It offered nothing in terms of saving space, `sizeof(int) == sizeof(void*)`
> 2. It only abstracted data types specifically for objects that are pooled. You still have to cast the result anyway too.
> 3. A pointer to the data works better since it decouples the list from the other pools entirely. It also limits the interface to the pool.
> 4. Accessing via the offsets was cumbersome and would have required macros to cleanup.
> 
> Now, with this commit as of this comment. I have used a union of an `int` and `void*` so we can support the original **list** and **int list**. However, I think I'm going to remove that, unless it is helpful elsewhere, because I don't think we need to have the **int list** anymore. Why? Because we can use the pool's bitmap tracker to do the same thing but faster. This would also limit the use of our largest list.
> 
> Going to clean this up more and switch to the bitmap stuff.

@ricfehr3 So this will generalize the list to non-pooled objects? Also using the bitmap as an int list sounds a bit hacky but if there's a clear and simple wrapper API to use it, it's okay. Otherwise if you don't want a union you can always use the same method I did with intptr_t but I have no idea if it's preferable.

source/list.c
MeirGavish: I don't mind a "for each" macro if you want to add it

ricfehr3:
> > So, I realized that using integer offsets wasn't the vibe:
> > 
> > 1. It offered nothing in terms of saving space, `sizeof(int) == sizeof(void*)`
> > 2. It only abstracted data types specifically for objects that are pooled. You still have to cast the result anyway too.
> > 3. A pointer to the data works better since it decouples the list from the other pools entirely. It also limits the interface to the pool.
> > 4. Accessing via the offsets was cumbersome and would have required macros to cleanup.
> > 
> > Now, with this commit as of this comment. I have used a union of an `int` and `void*` so we can support the original **list** and **int list**. However, I think I'm going to remove that, unless it is helpful elsewhere, because I don't think we need to have the **int list** anymore. Why? Because we can use the pool's bitmap tracker to do the same thing but faster. This would also limit the use of our largest list.
> > Going to clean this up more and switch to the bitmap stuff.
> 
> @ricfehr3 So this will generalize the list to non-pooled objects? Also using the bitmap as an int list sounds a bit hacky but if there's a clear and simple wrapper API to use it, it's okay. Otherwise if you don't want a union you can always use the same method I did with intptr_t but I have no idea if it's preferable.

Yeah, it's getting all a bit confusing. I think I owe you and anyone looking at this a real good explanation for wtf is going on here. I'll try to clear any confusion. I go over some basic stuff here, not for you, but for anyone wanting to chime in and to make it easier for my brain to put into words.

First, to quickly answer your first question:

_So this will generalize the list to non-pooled objects?_ : The list is generic now, just holding a pointer. It leaves it up to the programmer to use pooled objects. However, there is nothing stopping you from using any pointer. I'll try to explain that below.

Let's look at the **previous implementation** and **current implementation** of this change. You'll see below that:
- The previous implementation offered no benefits 
- The previous implementation was way more cumbersome
- The previous implementation was slower

---

### Prerequisites
The diagrams I put in the [Wiki](https://github.com/cellos51/balatro-gba/wiki/Data-Structures) so far are relevant for the **previous implementation** as well as the **Memory Pool** and **Bitmap Tracker**. (*going to update these diagrams soonish, may not be valid in a while*).

For context and clarity here are some definitions:
- **List**: A doubly-linked list holding **ListNode**s. Represented by `head` and `tail` variables pointing to the first/last **ListNode** respectively, or some undefined value if it's empty 
- **ListNode**: An entry in the **List**, it has `prev` and `next` variables pointing to other **ListNode**s in the same **List**.
- **Memory Pool**: A data structure to fetch free pointers of a defined **Object** type. It stores this as a regular array of **Object**s and tracks with a **Bitmap Tracker**. Gives a `malloc`/`free` replacement
- **Bitmap Tracker**: A data structure of flags representing 256 entries via an 8-word array.

---

### Getting Data From the List
#### Previous implementation
- I had the **ListNode** "pointers" represented as integer offsets into a corresponding **Memory Pool**. The idea was it allowed all data types in any list to be represented as an `integer`, essentially "_abstracting_" the data that was stored. This straight up was a bad idea, you'll see below:

> ```c
> typedef struct 
> {
> 	int prev;        // An offset into the ListNode memory pool
> 	int data_offset; // An offset into another memory pool
> 	int next;        // An offset into the ListNode memory pool
> } ListNode;
> ```
  
- The problem here begins at accessing an object.
- To access an object at an offset in a pool, you use the `POOL_AT` macro. For example: 

> ```c
> MyObject* my_object = POOL_AT(MyObject, 19);
> ```

We want to access the object at the `data_offset` of a specific **ListNode** in a **List**. One example of this in the code is we needed to get the **ListNode** at the N'th offset in a **List** to [choose the shop joker to purchase based on the cursor offset](https://github.com/ricfehr3/balatro-gba/commit/bfd332909b6697611b52a3f0db8ccdd6db5901ed#r168834371). To do this we needed a function: `int list_get_at_idx(List* list, int idx)` that returns the `int data_offset` of a **ListNode** at the specific __idx__(_N'th_) index value in a **List** (_confused yet? This is a benefit of the current implementation as well, much more straight forward. The diagrams in the wiki could help_).  

So, now that call to select a joker from the pool ended up looking something like this:
> ```c
> int offset = next_selection;
> MyObject* my_object = POOL_AT(MyObject, list_get_at_idx(&my_list, offset));
> ```

Let's compare that to how the new implementation looks:

#### Current implementation
- **ListNode** now uses actual pointers. `prev`  and `next` point to __ListNode*__, and `int data_offset` is replaced with a `void* data`.

>```c
>typedef struct 
>{
>	ListNode* prev;    // A pointer to a member of the ListNode memory pool
>	void* data_offset; // A pointer to any data
>	ListNode* next;    // A pointer to a member of the ListNode memory pool
>} ListNode;
>```

- Immediately, you can see that this approach doesn't care about the pool at all. This has some benefits:
1. It removes the need to use the `POOL_AT` macro, since your pointer is already available (_likely with `POOL_GET` anyway_). Now it returns `void* data` from a **ListNode** with `list_get_at_idx`. Imo, it's far more readable and also it's not a macro so IDEs and compilers should be far happier:
>```c
> MyObject* my_object = (MyObject*)list_get_at_idx(&my_list, idx)
>```
2. Also, you can see in the code snippet above that it prevents the operation of indexing the pool by removing `POOL_AT` , so it's also faster.
3. It helps encapsulate the use of the **Memory Pool** to only using **POOL_GET** and **POOL_FREE** in game logic code, completely swapable with `malloc`/`free`.  This also removes the requirement of a strongly dependent **List** implementation to match the **Memory Pool**. They shouldn't know about each other.

### Putting data in the list

Both the **previous** and **current** implementations only provide `list_push_front` and `list_push_back`, and `list_remove_at_idx`. No insertion. For the context of this discussion, the only important difference *is the way you get the data to push*. 

Once again, when pushing objects to the `tail`of the **List** with the **previous implementation**, you needed to first get the **Memory Pool** index of the pointer you wanted to represent, (_taken from [here](https://github.com/ricfehr3/balatro-gba/commit/bfd332909b6697611b52a3f0db8ccdd6db5901ed#r168834756)_):

> ```c
> list_push_back(&_all_jokers_list, POOL_IDX(JokerObject, joker_object));
> ```

But, with the **new implementation** you just push the pointer itself:

> ```c
> list_push_back(&_all_jokers_list, joker_object);
> ```

## The case for the bitmap tracker swap

I found that the `int list` was being used in one place and that was keeping track of what jokers are available, or rather, not on screen (hand/shop/being discarded). This was done, as you had it, by representing the `void* data` pointer in a **ListNode** as an `intptr_t`. It was used as follows:

- On game start, the available jokers integer list is filled  with `0 - NUMBER_OF_JOKERS` entries, each holding it's corresponding iteration value as an `intptr_t`.
- On leaving the list, the joker would track it's ID, retrieved from the available jokers integer list.
- This ID stays with the joker until it's freed back into the available jokers list. The freed joker ID is set as the value in a new entry into available jokers integer list, completing the cycle.

This operation in itself is very similar to what is done with the **Memory Pool** already to track what offsets are available via the **Bitmap Tracker**. Switching it over was actually pretty easy as instead of appending and removing integers from the list, it just had to flip a bit in the **Bitmap Tracker** at the Joker ID offset. 

The only spot this was any trouble was a simple call to get a random joker index:

> ```c
> joker_idx = random() % list_get_len(_avail_shop_jokers_list);
> ```

Switching this over is not trivial. Unlike a **List**, the **Bitmap Tracker** can have gaps. So two additional operations are needed to get correct indices for Joker ID's:
- The **Bitmap Tracker** needs to count how many 1's there are in it's array, essentially replicating getting the **List** length. Luckily another gcc builtin can do that quickly: [`__builtin_popcount`](https://gcc.gnu.org/onlinedocs/gcc/Bit-Operation-Builtins.html#index-_005f_005fbuiltin_005fpopcount)
- The **Bitmap Tracker** needs to skip over invalid entries (or bits that are `0`). To do this, I'm trying the following (not set in stone at all, example [here](https://github.com/ricfehr3/balatro-gba/commit/eae4750b2a15cf71b9098d3f918ebb71a5ec7c86#r168834919)). Given a random number between 0 and the available jokers length. Starting at the first integer of the bitmap tracker:
	1. Count the number of `1`'s in the integer. If it's greater than the number, proceed to the next step, otherwise go to the next integer in the array and try again, summing the previous number of `1`'s in the integer
	2. Once the sum of all the `1`'s is greater than the index you want, operate on that integer. You now know that the flag of your offset is one of the flags in the bitmap array integer you just found. So in a for loop check each flag.
	3. Counting up from the sum of the previous trackers, plus the inner `for loop` tracker, you can find the offset in the bitmap array that corresponds to the correct Joker ID skipping any empty ("in-use") entries.

I feel this is an acceptable complexity given that:
1. This random number only needs to be calculated per joker in the shop (*twice right now*).
2. Can only have a maximum of 7 (*for the number of bytes*) + 32 (*for the number of ints*) = 39 iterations for the worst case scenario. 
3. This operation would be faster than traversing a **List**, consistently.

I don't think we need to abstract the fact that it's a **Bitmap Tracker** that's being used to track the available jokers. It's doing exactly what it's made for. I think putting the **Bitmap Tracker** into it's own file with a clearer API that works in both the context of the **Memory Pool** and the available jokers information is needed. Currently I'm piggy backing off of the **Memory Pool** file `pool.c`, but that isn't going to stay like that.

source/list.c
ricfehr3: Cool, I think I'm going to add that.

MeirGavish:
Okay, maybe just wrappers like `is_joker_available_to_shop()`, `get_num_jokers_available_to_shop()`, `set_joker_shop_availability()` to abstract the implementation so the list/bitmap implementation is not too coupled with the usage. 
Also since PR #152 got merged the implementation changed to weight the jokers by rarity by making a temporary array of list indices that match the rarity, so that will need to be accommodated in the bitmap tracker implementation. Maybe making a bitmap for each rarity would work? Or the same implementation could work with bitmap indices but would just need to make sure the joker is available to shop (the index is really set) before putting it in the temporary array.

ricfehr3:
> Okay, maybe just wrappers like `is_joker_available_to_shop()`, `get_num_jokers_available_to_shop()`, `set_joker_shop_availability()` to abstract the implementation so the list/bitmap implementation is not too coupled with the usage. Also since PR #152 got merged the implementation changed to weight the jokers by rarity by making a temporary array of list indices that match the rarity, so that will need to be accommodated in the bitmap tracker implementation. Maybe making a bitmap for each rarity would work? Or the same implementation could work with bitmap indices but would just need to make sure the joker is available to shop (the index is really set) before putting it in the temporary array.

I agree, I'll make sure the interface to the jokers is clear with those functions. Good point. 

When I rebased last night, I noticed that random joker function and I think I've got solution. Though, I'm really hoping to get your opinion on it.

- _Note_, I changed `bitmap` to `bitset` because that is what the data structure is actually called, and I was being silly trying to "map" the "bits" to the array.
- In the function [`game_shop_get_random_joker_idx()`](https://github.com/cellos51/balatro-gba/blob/main/source/game.c#L2520), the operation we are replacing is:
  - _go through the list of Jokers that are not in-play from the `jokers_available_to_shop` list._ 
  - _Go through the list one-by-one and get the JokerId from it's entry_ 
  - _The `JokerId` is an offset into the `JokerInfo` array. We use it to get data from the array_
- The goal of that operation for our use can be also be seen as: 
  - _If a Joker is not in-play, get it's ID, an offset into the `JokerInfo` array._
  
Cool, so each flag in our bitset now represents if a Joker is in-play. So, now instead of a single list to parse through, we need to get the offset at every set flag in the bitset.

There's a couple ways to do that, currently I'm doing it a slow way. I won't explain it too deeply here because I'm going to change it, but here is a permalink to it: https://github.com/ricfehr3/balatro-gba/blob/feature/index-list/source/bitset.c#L89 . Basically, you can use the iterator to get the Joker Id's as you can see here: https://github.com/ricfehr3/balatro-gba/blob/feature/index-list/source/game.c#L2500

The much faster way to do this is to change the iterator method. Instead of `bitset_find_idx_of_nth_set` (still helpful for finding a single index, quickly, but doesn't iterate well.) I plan on just using 3 `int`s. 
- One to track the current word in the array.
- One to track the current bit in the word. 
- One to track the current actual offset (which will be the Joker Id for us).

This operation is quick and simple compared to a `List` iterator, and frees up 40 (or how many jokers we have later) list nodes to be something else that would benefit from a list more.

---

**Edit**: The bitset iterator is now updated, still could be improved.

include/bitset.h
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * @brief Current bit the iterator is on
+     */
+    int bit;
+
+    /**
+     * @brief Number of bits that have been iterated through in total
+     */
+    int itr; 
+} BitsetItr;
```
ricfehr3, line 92: I realize now I need to be more descriptive in what this is exactly. 

This ONLY iterates over set flags, not over every entry. I don't think we need the functionality of going over every entry (if we do we can add it). 

Describe this better

include/game.h
```diff
@@ -1,6 +1,8 @@
 #ifndef GAME_H
 #define GAME_H
 
+#include "list.h"
```
ricfehr3: Fix, redo the forward declaration

MeirGavish:
Yeah, it sounds good. 
`bitset` is definitely a better name as it clearly reflects the usage as a "set of bits" and "bitmap" confusingly coincides with "bitmap background modes/bitmap images".
This solution seems to work well, and it makes sense to optimize the iterator, I'm not sure I have any other input about it.
The only thing that now maybe `game_shop_get_random_joker_idx()` should be renamed to something like `game_shop_get_rand_available_joker_id()` to signify that now you get a joker ID and not an index to another list.

source/game.c
```diff
@@ -2734,19 +2724,24 @@ static int game_shop_get_random_joker_idx()
...
 
-    for (int i = 0; i < jokers_avail_size; i++)
+    BitsetItr itr = bitset_itr_new(&_avail_jokers_bitset);
+
+    int i = 0;
+    intptr_t joker_id = UNDEFINED;
+    while((joker_id = bitset_itr_next(&itr)) != UNDEFINED)
     {
-        intptr_t joker_id = int_list_get(jokers_available_to_shop, i);
+        //intptr_t joker_id = int_list_get(jokers_available_to_shop, i);
```
ricfehr3: delete

source/joker_effects.c
```diff
@@ -266,22 +267,23 @@ static JokerEffect joker_stencil_effect(Joker *joker, Card *scored_card, enum Jo
...
+    ListNode* ln;
+
+    while((ln = list_itr_next(&itr)))
     {
-        JokerObject* joker_object = list_get(jokers, i);
-        if (joker_object->joker->id == JOKER_STENCIL_ID)
-        {
-            effect.xmult++;
-        }
+        //JokerObject* joker_object = POOL_AT(JokerObject, ln->elem_idx);
```
ricfehr3: delete

source/game.c
```diff
@@ -254,14 +259,45 @@ bool is_shortcut_joker_active(void)
...
-static int four_fingers_joker_count = 0;
+//static int four_fingers_joker_count = 0;
 static int straight_and_flush_size = STRAIGHT_AND_FLUSH_SIZE_DEFAULT;
 
 int get_straight_and_flush_size(void) 
 {
     return straight_and_flush_size;
 }
 
+static inline void _set_shop_joker_avail(int joker_id, bool avail)
```
ricfehr3, line 268: @MeirGavish Here are the interfaces to the bitset we talked about. There is one use of a `BitsetItr` below, but it's in another `Joker` function, so I'm happy with it.

ricfehr3:
Moving this out of draft and opening it up for review.

Please do not merge yet, I'm writing tests but hopefully I won't be changing much with what I have here. 

I was going to make a macro to iterate through the `List` and `Bitset`, but I found that the iterator pattern worked nicely and I didn't feel a need to make it more complicated. It's a pretty simple interface:

```c
    ListItr itr = list_itr_new(&list);
    ListNode* ln;

    while((ln = list_itr_next(&itr)))
    {
        MyObject* obj = ln->data; 
        // do stuff
        ...
    }
```

ricfehr3:
@MeirGavish Okay, added some tests covering every function. The tests are a bit spaghetti and could be better. However, I'm pretty confident they cover everything in the state they are in.

ricfehr3:
Finishing this up to unblock new work. Mainly ~#207~ #158

@MathisMartin31 How did you end up testing the jokers exactly? I would like make sure I'm not breaking any of your work from my rebase on your joker effects work. 

@MeirGavish Can you request copilot to review this? I don't have that authority.

MeirGavish:
> Finishing this up to unblock new work. Mainly #207

@ricfehr3 You meant to unblock #158?

source/list.c
```diff
@@ -1,121 +1,171 @@
...
+    }
+
+    return len;
 }
 
-int list_get_size(List *list)
+void* list_get_at_idx(const List* list, int n)
 {
-    if (list == NULL)
+    if(n >= list_get_len(list)) return NULL;
```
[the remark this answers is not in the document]
MeirGavish: `list_get_len()` does not iterate the entire list.

source/game.c
```diff
@@ -228,9 +230,12 @@ static int selection_y = 0;
 
 static bool sort_by_suit = false;
 
-static List *jokers = NULL;
-static List *discarded_jokers = NULL;
-static List *jokers_available_to_shop; // List of joker IDs
+static List _active_jokers_list;
```
MeirGavish: I think a better name here would be `_owned_jokers_list`, "active" feels like it could imply jokers at the shop but not yet owned.

source/game.c
```diff
@@ -2767,39 +2761,33 @@ static int game_shop_get_random_joker_idx()
...
 
-    shop_jokers = list_new(MAX_SHOP_JOKERS);
+    if (_no_avail_jokers()) return;
+
+    list_destroy(&_shop_jokers_list);
+    _shop_jokers_list = list_new();
 
     for (int i = 0; i < MAX_SHOP_JOKERS; i++)
     {
         intptr_t joker_id = 0;
```
```suggestion
        int joker_id = 0;
```

I guess this should be changed from inptr_t now as there's no reason for it

source/game.c
```diff
@@ -346,16 +382,18 @@ int get_scored_card_index(void)
...
-{
-    return jokers;
+List* get_jokers_list(void) {
+    return &_active_jokers_list;
 }
 
-bool is_joker_owned(int joker_id)
-{
-    for (int k = 0; k < list_get_size(jokers); k++)
+// is it possible to... directly map a jokers ID to the offset?
```
MeirGavish: I don't think so, the list is very dynamic and can contain any jokers in any order.

source/game.c
```diff
@@ -3444,10 +3459,10 @@ static void held_jokers_update_loop()
 
     FIXED hand_x = int2fx(HELD_JOKERS_POS.x);
 
-    int jokers_top = list_get_size(jokers) - 1;
+    int jokers_top = list_get_len(&_active_jokers_list) - 1;
     for (int i = jokers_top; i >= 0; i--)
     {
-        JokerObject *joker = list_get(jokers, i);
+        JokerObject *joker = (JokerObject*)list_get_at_idx(&_active_jokers_list, i);
```
MeirGavish: Why is this not using the iterator? Because it's from last to first? I guess we can implement a reverse iterator... I'm not even sure why it's in that order or if it has to be.
I'm a little worried about this since it's called for every frame.

source/game.c
```diff
@@ -366,47 +404,7 @@ bool is_joker_owned(int joker_id)
...
-        {
-            straight_and_flush_size = STRAIGHT_AND_FLUSH_SIZE_FOUR_FINGERS;
-        }
-        four_fingers_joker_count++;
-    }
-
-    if (joker_object->joker->id == SHORTCUT_JOKER_ID) 
-    {
-        shortcut_joker_count++;
-    }
```
MeirGavish, line 386: This removes Four Fingers and Shortcut jokers code and would break them.

source/game.c
```diff
@@ -2893,38 +2891,43 @@ static void jokers_sel_row_on_selection_changed(SelectionGrid *selection_grid,
...
 void joker_start_discard_animation(JokerObject *joker_object)
 {
     joker_object->sprite_object->tx = int2fx(JOKER_DISCARD_TARGET.x);
     joker_object->sprite_object->ty = int2fx(JOKER_DISCARD_TARGET.y);
-    list_append(discarded_jokers, joker_object);
+    list_push_back(&_discarded_jokers_list, joker_object);
 }
 
-void game_sell_joker(int joker_idx)
+void game_sell_joker(int joker_id)
```
MeirGavish: Why is this an ID and not an index now?

source/game.c
```diff
@@ -366,47 +404,7 @@ bool is_joker_owned(int joker_id)
...
-    }
-
-    if (joker_object->joker->id == SHORTCUT_JOKER_ID) 
-    {
-        shortcut_joker_count--;
-    }
-
-    list_remove_by_idx(jokers, joker_idx);        
+    list_push_back(&_active_jokers_list, joker_object);
 }
```
MeirGavish, line 447: Here too, keep this wrapper, we will need it...

source/game.c
```diff
@@ -254,14 +259,45 @@ bool is_shortcut_joker_active(void)
...
 #define STRAIGHT_AND_FLUSH_SIZE_FOUR_FINGERS 4
 #define STRAIGHT_AND_FLUSH_SIZE_DEFAULT 5
-static int four_fingers_joker_count = 0;
+//static int four_fingers_joker_count = 0;
```
```suggestion
static int four_fingers_joker_count = 0;
```

And uncomment

source/game.c
```diff
@@ -228,9 +230,12 @@ static int selection_y = 0;
 
 static bool sort_by_suit = false;
 
-static List *jokers = NULL;
-static List *discarded_jokers = NULL;
-static List *jokers_available_to_shop; // List of joker IDs
+static List _active_jokers_list;
+static List _discarded_jokers_list;
+
+#define MAX_JOKERS_IN_BITMAP 64
```
MeirGavish: This is another number that bounds the number of jokers and is not tied to the registry... I guess we can use MAX_DEFINABLE_JOKERS here though it's bigger and will take more memory. I'm thinking of going against my previous principle and maybe externing the joker registry so its size is known at compile time and can be used in macros and definitions.

source/game.c
```diff
@@ -1376,25 +1374,17 @@ void game_change_state(enum GameState new_game_state)
...
 {
-    jokers_available_to_shop_init();
-
-    // Initialize jokers list
-    if (jokers) list_destroy(&jokers);
-    jokers = list_new(MAX_JOKERS_HELD_SIZE);
+    // Initialize all jokers list once
+    _active_jokers_list = list_new();
+    _discarded_jokers_list = list_new();
+    _shop_jokers_list = list_new();
```
MeirGavish: Do we not need to free previously allocated lists? If we can't do it here it should be done in `game_over_on_exit()` since it calls game_init() every time.

source/game.c
```diff
@@ -2097,10 +2088,10 @@ static void cards_in_hand_update_loop(bool* discarded_card, int* played_selectio
 // returns true if a joker was scored, false otherwise
 static bool check_and_score_joker_for_event(int* iteration_start, Card* played_card, enum JokerEvent joker_event)
 {
-    for (int k = *iteration_start; k < list_get_size(jokers); k++)
+    for (int k = *iteration_start; k < list_get_len(&_active_jokers_list); k++)
     {
         (*iteration_start)++;
-        JokerObject *joker = list_get(jokers, k);
+        JokerObject *joker = list_get_at_idx(&_active_jokers_list, k);
```
MeirGavish: Maybe here we can pass a `ListItr*` instead of `iteration_start` so we can optimize the iteration here.

source/game.c
```diff
@@ -2893,38 +2891,43 @@ static void jokers_sel_row_on_selection_changed(SelectionGrid *selection_grid,
...
     if (prev_selection->y == row_idx)
     {
-        JokerObject* joker_object = list_get(jokers, prev_selection->x);
-        erase_price_under_sprite_object(joker_object->sprite_object);
-        sprite_object_set_focus(joker_object->sprite_object, false);
+        JokerObject* joker_object = (JokerObject*)list_get_at_idx(&_active_jokers_list, prev_selection->x);
+        if(joker_object != NULL) {
+            erase_price_under_sprite_object(joker_object->sprite_object);
+            sprite_object_set_focus(joker_object->sprite_object, false);
+        }
```
```suggestion
        if (joker_object != NULL) 
        {
            erase_price_under_sprite_object(joker_object->sprite_object);
            sprite_object_set_focus(joker_object->sprite_object, false);
        }
```

Allman style, we should reformat everything and add clang enforcement. But it's good to add this NULL-check.

source/game.c
```diff
@@ -2893,38 +2891,43 @@ static void jokers_sel_row_on_selection_changed(SelectionGrid *selection_grid,
...
     if (new_selection->y == row_idx)
     {
-        JokerObject* joker_object = list_get(jokers, new_selection->x);
-        sprite_object_set_focus(joker_object->sprite_object, true);
-        print_price_under_sprite_object(joker_object->sprite_object, joker_get_sell_value(joker_object->joker));
+        JokerObject* joker_object = (JokerObject*)list_get_at_idx(&_active_jokers_list, new_selection->x);
+        if(joker_object != NULL) {
+            sprite_object_set_focus(joker_object->sprite_object, true);
+            print_price_under_sprite_object(joker_object->sprite_object, joker_get_sell_value(joker_object->joker));
+        }
```
```suggestion
        if (joker_object != NULL) 
        {
            sprite_object_set_focus(joker_object->sprite_object, true);
            print_price_under_sprite_object(joker_object->sprite_object, joker_get_sell_value(joker_object->joker));
        }
```

source/game.c
```diff
@@ -2893,38 +2891,43 @@ static void jokers_sel_row_on_selection_changed(SelectionGrid *selection_grid,
 {
     if (prev_selection->y == row_idx)
     {
-        JokerObject* joker_object = list_get(jokers, prev_selection->x);
-        erase_price_under_sprite_object(joker_object->sprite_object);
-        sprite_object_set_focus(joker_object->sprite_object, false);
+        JokerObject* joker_object = (JokerObject*)list_get_at_idx(&_active_jokers_list, prev_selection->x);
```
MeirGavish: Originally I wasn't worried about this since I said the list wouldn't get very long but now I do wonder how significant the increase in runtime is here.

source/game.c
ricfehr3: I agree, that was annoying me too. I'm switching it to `_owned_jokers_list`

source/game.c
ricfehr3: Whoops, that was an early note for myself, haha. Removing

source/game.c
ricfehr3: Readded. Sorry, sloppy rebase work there.

source/game.c
ricfehr3: ~This does need to be `joker_idx`~. It's a bit confusing as the Joker's ID is also it's index into the bitset. The ideas are different and shouldn't be treated the same (as I did here), but it does work, somewhat unfortunately. 

~Definitely moving back to `joker_idx`, tho.~

I'm looking at this more and I do think we should combine the ideas that a jokers ID is it's "index", but drop the idea of an "index" all together. We already use it's ID as an index into the joker_registery, which relates to the index in it's bitset. The idea of an index isn't needed, it would be better anyway to operate on ID's alone. It just is convenient that the ID's are one-to-one.

What do you think?

source/game.c
ricfehr3: I like the idea of using `MAX_DEFINABLE_JOKERS`. For the `bitset` in particular there is no additional memory. They are always 8 ints for now, allowing for 256 entries by default. 

Switching it for now to `MAX_DEFINABLE_JOKERS`. 

I will say though, maybe we should consider switching the joker register to an X macro after all? The big benefit that allows is for an `enum` to be defined to show us the number of `JOKERS_IN_REGISTERY`, which I think would justify the X-macro since we are relying on that variable in a few places loosely.

source/game.c
ricfehr3: Yes, 100% we do. I'll put them in that function. 

I also think it would be appropriate to do it before call `list_new()`, but I think it looks more "correct" in the `game_over_on_exit()` function.

source/game.c
ricfehr3: Caught me red handed. I agree, I had a good `.clang-format` working here -> https://github.com/ricfehr3/balatro-gba/blob/feature/cleanup-code-style/.clang-format

After this PR, I wouldn't mind finishing that work up.

source/game.c
ricfehr3: Here I think we are good. Since the number of "owned" jokers is always going to be within a reasonable value, and this operation only happens in the shop on a sale. 

However, I haven't tested performance. No easy way to, as far as I know?

source/game.c
ricfehr3: I just tried it with a regular iterator and it seems to work fine. As far as I can tell it doesn't matter if you iterate up or down, since it would always map the same `joker` to the same `i` value from the list. 

Good point on the usage of these list functions in update functions like this, absolutely it is a source of slowdown when used this frequently. But the iterator cleans up that extra overhead here.

source/game.c
ricfehr3: I agree. It's hard to test... I might need some ideas to make sure I'm not breaking anything, but the iterator replacement seems to be working great.

ricfehr3:
@MeirGavish Thanks for taking a look at this. I appreciate the level of detail. I've addressed all the comments (so far). But yeah, keep coming back to it, I am too. These big PRs can be a pain, I appreciate you rocking through them with me. 

From our convo so far, sounds like we need to make issues for:

- Using X-macros for the joker registry: https://github.com/cellos51/balatro-gba/pull/168#discussion_r2487048651

- Adding list insertion and reverse iterators.

source/game.c
MeirGavish: I think it should be fine if jokers are correctly activated.

source/game.c
MeirGavish: But here it's the index of the joker in the shop list, which is different to its ID obtained through `joker_object->joker->id` and is the actual value passed to the bitset.

source/game.c
```diff
@@ -2095,12 +2123,17 @@ static void cards_in_hand_update_loop(bool* discarded_card, int* played_selectio
...
 
 // returns true if a joker was scored, false otherwise
-static bool check_and_score_joker_for_event(int* iteration_start, Card* played_card, enum JokerEvent joker_event)
+//static bool check_and_score_joker_for_event(int* iteration_start, Card* played_card, enum JokerEvent joker_event)
+static bool check_and_score_joker_for_event(ListItr* itr, Card* played_card, enum JokerEvent joker_event)
 {
-    for (int k = *iteration_start; k < list_get_size(jokers); k++)
+    ListNode* ln;
+
+    //for (int k = *iteration_start; k < list_get_len(&_owned_jokers_list); k++)
```
ricfehr3: delete these comments, me.

source/game.c
MeirGavish: Yeah, what worried me is that this is called for every move of the selection when highlighting different jokers in the shop.
I tried testing performance by activating the HBLANK affine background effect on the rest of the game by simply removing the `IRQ_HBLANK` toggle in `affine_background_change_background()`. Doing that creates other performance artifacts but I don't see anything when selecting shop jokers so it seems to be fine.

MeirGavish:
I know you might be in the middle of changing things but I noticed some things when running this I thought I should mention.
Owned jokers can reappear in the shop, you can test it by giving yourself a lot of `STARTING_MONEY` and rerolling a lot.

balatro-gba-5
balatro-gba-6


Also the motion effects when rerolling are bugged.

https://github.com/user-attachments/assets/f8283830-ac35-4b2f-a48c-801ac561e2b7

source/game.c
MeirGavish: @MathisMartin31 Actually, if you can share how you tested the retriggers it can be helpful.

ricfehr3:
@MeirGavish You beat me to it, ha. I'll have to come back to this tomorrow. I thought I had fixed it, but I guess not. I'll need to sleep before looking at this again, haha. Cheers

ricfehr3:
@MeirGavish Alrighty, after a sleeping and working, I found the issue right away (_I think_). Good catch on that. I'm pretty happy with this, besides how to test the jokers exactly.

MeirGavish:
Nice! I'll try to take another look tomorrow or during the weekend

source/game.c
MeirGavish: @ricfehr3 Yeah, I'm going to insist on `joker_idx` here, we can change it to `owned_joker_idx` but it's important to differentiate the index and the ID here. I know it's confusing the joker ID is technically the index into the registry and bitset and that's why I think the code should never name the ID as index/idx.

source/game.c
```diff
@@ -2893,38 +2928,45 @@ static void jokers_sel_row_on_selection_changed(SelectionGrid *selection_grid,
...
     
-    JokerObject *joker_object = list_get(jokers, joker_idx);
+    JokerObject* joker_object = (JokerObject*)list_get_at_idx(&_owned_jokers_list, joker_id);
     money += joker_get_sell_value(joker_object->joker);
     display_money(money);
     erase_price_under_sprite_object(joker_object->sprite_object);
 
-    remove_held_joker(joker_idx);
-    int_list_append(jokers_available_to_shop, (intptr_t)joker_object->joker->id);
+    list_remove_at_idx(&_owned_jokers_list, joker_id);
```
MeirGavish: This should call `remove_held_joker()` so it calls everything it encapsulates such as the Four Fingers/Shortcut logic

source/game.c
```diff
@@ -2095,12 +2123,13 @@ static void cards_in_hand_update_loop(bool* discarded_card, int* played_selectio
 }
 
 // returns true if a joker was scored, false otherwise
-static bool check_and_score_joker_for_event(int* iteration_start, Card* played_card, enum JokerEvent joker_event)
+static bool check_and_score_joker_for_event(ListItr* itr, Card* played_card, enum JokerEvent joker_event)
```
MeirGavish: Maybe rename `itr` to something like `starting_joker_itr`? Just to make the name a little more meaningful and clear

source/game.c
```diff
@@ -2893,38 +2928,45 @@ static void jokers_sel_row_on_selection_changed(SelectionGrid *selection_grid,
...
+    JokerObject* joker_object = (JokerObject*)list_get_at_idx(&_owned_jokers_list, joker_id);
     money += joker_get_sell_value(joker_object->joker);
     display_money(money);
     erase_price_under_sprite_object(joker_object->sprite_object);
 
-    remove_held_joker(joker_idx);
-    int_list_append(jokers_available_to_shop, (intptr_t)joker_object->joker->id);
+    list_remove_at_idx(&_owned_jokers_list, joker_id);
+    
+    _set_shop_joker_avail(joker_object->joker->id, true);
```
MeirGavish: This should call remove_held_joker() so it calls everything it encapsulates such as the Four Fingers/Shortcut logic, at the current state selling these jokers will not deactivate them and that's a bug. I'm also thinking the call to `_set_shop_joker_avail` should be moved there.

include/list.h
```diff
@@ -1,27 +1,179 @@
...
+
+    /**
+     * @brief The current node in the list
+     */
+    ListNode* current_node;
+} ListItr;
+
+/**
+ * Create a new list.
+ *
```
MeirGavish, line 90:
```suggestion
 *
 * The list must be freed with @ref list_destroy
 *
 * While this function does not allocate memory for the list itself, the list does allocate memory for each element.
 *
```

Add some documentation instructing that the list must freed as well and how.

include/list.h
```diff
@@ -1,27 +1,179 @@
...
+int list_get_len(const List* list);
+
+/**
+ * Create a new @ref ListItr
+ *
+ * @param list pointer to a @ref List
+ *
+ * @return A new @ref ListItr
+ */
+ListItr list_itr_new(const List* list);
```
MeirGavish: As with `list_new()`, these are the first times the phrase "new" is in a function without actually allocating anything that needs to be freed. It's a bit of an inconsistency so either add a comment in the documentation explaining this does not allocate memory. Alternatively I'm wondering if we can drop or replace the "new", maybe it can be dropped for this function but for `list_new()` it feels like the name becomes too generic then.

source/game.c
```diff
@@ -3029,9 +3072,9 @@ static void shop_top_row_on_selection_changed(SelectionGrid* selection_grid, int
         }
         else 
         {
-            JokerObject *joker = list_get(shop_jokers, new_selection->x - 1);
-            sprite_object_set_focus(joker->sprite_object, true); 
-            // -1 to account for next round button
+            int idx = new_selection->x - 1;
```
```suggestion
            int idx = new_selection->x - 1; // -1 to account for next round button
```

Keep the comment though I know it's duplicated in other uses.

source/game.c
```diff
@@ -3014,8 +3056,9 @@ static void shop_top_row_on_selection_changed(SelectionGrid* selection_grid, int
         }
         else 
         {
-            JokerObject *joker = list_get(shop_jokers, prev_selection->x - 1);
-            sprite_object_set_focus(joker->sprite_object, false); 
+            int idx = prev_selection->x - 1;
+            JokerObject *joker_object = (JokerObject*)list_get_at_idx(&_shop_jokers_list, idx);
+            sprite_object_set_focus(joker_object->sprite_object, false); 
             // -1 to account for next round button
```
```suggestion
            int idx = prev_selection->x - 1;  // -1 to account for next round button
            JokerObject *joker_object = (JokerObject*)list_get_at_idx(&_shop_jokers_list, idx);
            sprite_object_set_focus(joker_object->sprite_object, false); 
```

And maybe move the comment here too

source/game.c
ricfehr3: Absolutely, that's fair. Sorry, I hadn't seen that comment before. I'll change that on my next push.

ricfehr3: @MeirGavish Fixed it. I agree with your reasoning 100%. It's called as the idx in the game shop itself. Makes so much more sense now, haha.

include/list.h
```diff
@@ -1,27 +1,179 @@
...
+    /**
+     * @brief A pointer to the @ref List this is iterating through
+     */
+    const List* list;
+
+    /**
+     * @brief The current node in the list
+     */
+    ListNode* current_node;
+} ListItr;
```
MeirGavish, line 83: The documentation for each field here feels a little excessive to me, I think all of these are obvious from the name and to whoever knows what a linked list looks like. Maybe it's better to avoid excessive comments and let the code be self-documenting but I don't know, maybe there is value to documenting every field.
For the bitset I do feel the documentation is helpful so maybe it's just me, or maybe it means the field names there should be improved there...

include/list.h
```diff
@@ -1,27 +1,179 @@
...
+ListItr list_itr_new(const List* list);
+
+/**
+ * Get the next entry in a @ref ListItr
+ *
+ * @param itr pointer to the @ref ListItr
+ *
+ * @return A pointer to the next @ref ListNode if valid, otherwise return NULL.
+ */
+ListNode* list_itr_next(ListItr* itr);
```
MeirGavish: This is more of a design comment that applies to several places in the code but I thought this is the best place to put it.
I don't think the only way to iterate over the list should give the user access to the nodes, in most usages the list user just wants access to the data and gets it from the node so this iterator should return the data. This will remove code duplication for accessing the data, decouple iteration from the list implementation, and will be safer since the list user doesn't get access to an internal structure, and preserve the list's responsibility for the nodes.
The one place where the node is needed is when removing during iteration. For that I actually think it would be best to still encapsulate it within the list iterator API with a function such as `list_iter_remove_prev()` that would prevent the user from somehow removing a node that is further ahead in the iteration, potentially breaking it.
If that proves problematic, at the very least provide separate functions - `list_itr_next()` that returns the data and `list_itr_next_node()` that returns the node, and in the documentation discourage usage of the latter over the former and document that removing nodes with the iterator must be with the node returned from `list_itr_next_node`
I think we may end up breaking this separation to some extend and provide access to the nodes for performance but any function that provides access to nodes should have a preferred alternative.

tests/list/list_test.c
```diff
@@ -0,0 +1,301 @@
...
+
+    // verify no data 
+    assert(my_cool_list.head == NULL);
+    assert(my_cool_list.tail == NULL);
+    assert(list_get_len(&my_cool_list) == 0);
+    assert(list_is_empty(&my_cool_list));
+
+    list_destroy(&my_cool_list);
+
+    assert(list_is_empty(&my_cool_list));
```
MeirGavish, line 24: I would expect the list to not be valid after a call to `list_destroy`, I think maybe it should be separated to `list_clear()` that removes the entries and promises validity and a wrapper `list_destroy` that doesn't promise validity, then it's more of an intended documented behavior than a side-effect.

include/bitset.h
```diff
@@ -0,0 +1,198 @@
...
+ * @brief Get a new @ref BitsetItr
+ *
+ * @param bitset A @ref Bitset to operate on
+ *
+ * @return A newly constructed BitsetItr
+ */
+BitsetItr bitset_itr_new(const Bitset* bitset);
+
+/**
+ * @brief Get the next index from a @ref BitsetItr
```
MeirGavish:
```suggestion
 * @brief Get the index of the next set bit in the bitset from a @ref BitsetItr
```

Just to make the documentation a little clearer as to what the iterator is iterating on

source/joker_effects.c
```diff
@@ -859,8 +863,8 @@ static JokerEffect brainstorm_joker_effect(Joker *joker, Card *scored_card, enum
         return effect;
     }
 
-    List* jokers = get_jokers();
-    JokerObject* first_joker = list_get(jokers, 0);
```
MeirGavish, line 863: I think this should still use `list_get_at_idx(jokers, 0)` rather than directly access the list's internal data. It would still return NULL if the list is empty, right? Though I think the following NULL-check is defensively redundant since the list will at least contain Brainstorm itself.

include/bitset.h
```diff
@@ -0,0 +1,198 @@
...
+ * @param idx the index of the flag to get
+ *
+ * @return the value of the flag as `true` or `false`
+ */
+bool bitset_get_idx(Bitset *bm, int idx);
+
+// Get the next free (set to 0) index in the bitset.
+// It also sets the bit which it maybe should do... It really shouldn't do two things
+// But it's such a fast operation idk. // TODO: decide what you wanna do
+int bitset_get_free_idx(Bitset *bitset);
```
MeirGavish: I think there isn't a real performance cost to separating this into get and set operations.
Alternatively, you can rename this to something like `bitset_allocate_idx` and then the multiple operations are implied from the name a bit like `stack_pop()` etc.

scripts/get_memory_map.sh
MeirGavish, line 1: Shouldn't this include the list nodes too now?

source/game.c
```diff
@@ -346,16 +380,17 @@ int get_scored_card_index(void)
     return scored_card_index;
 }
 
-List *get_jokers(void)
-{
-    return jokers;
+List* get_jokers_list(void) {
+    return &_owned_jokers_list;
```
```suggestion
List* get_jokers_list(void) 
{
    return &_owned_jokers_list;
```

Allman

source/game.c
```diff
@@ -346,16 +380,17 @@ int get_scored_card_index(void)
...
 }
 
-bool is_joker_owned(int joker_id)
-{
-    for (int k = 0; k < list_get_size(jokers); k++)
+bool is_joker_owned(int joker_id) {
+    ListItr itr = list_itr_new(&_owned_jokers_list);
```
```suggestion
bool is_joker_owned(int joker_id) 
{
    ListItr itr = list_itr_new(&_owned_jokers_list);
```

Allman

source/game.c
```diff
@@ -220,17 +222,18 @@ static int hand_selections = 0;
 static int scored_card_index = 0;
 
 // Keeping track of what Jokers are scored at each step
-static int joker_scored_index = 0;
-static int joker_round_end_index = 0;
+static ListItr _joker_scored_itr;
```
MeirGavish, line 225: There's a problem with this replacement because the `joker_scored_index` was previously initialized here while the iterator is being created with no initialization. There needs to be a corresponding call to `list_itr_new()` when the first joker scoring state (which is actually the card scoring state) is first entered, otherwise this could cause bugs.

source/game.c
```diff
@@ -2326,8 +2356,6 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
                                 *played_selections = 0;
                                 played_top = -1; // Reset the played stack
                                 scored_card_index = 0;
-                                joker_scored_index = 0;
```
MeirGavish, line 2329: I'm pretty sure removing this index reset can also cause bugs. The correct thing to do is replace it with an initialization on entering the scoring state as mentioned in a previous comment.

source/list.c
```diff
@@ -1,121 +1,170 @@
...
+    return list->len;
 }
 
-int list_get_size(List *list)
+void* list_get_at_idx(const List* list, int n)
 {
-    if (list == NULL)
+    if(n >= list_get_len(list) || n < 0) return NULL;
+
+    int len = 0;
```
MeirGavish: Maybe a more correct name would be `curr_idx` since this isn't really a length of anything.

MeirGavish (changes requested):
These are good improvements, well done! I also tested it and it seems the bugs we've seen really are fixed. There are still a few things that need fixing and this time I gave a closer look to the bitset and list APIs.

One more thing, you're the one who first started adding unit tests so huge props for that! I want to keep that habit of adding tests to added code when possible and I've been thinking that now that the bitset is its own separate module (which is also great), it should also have tests along with the pool and list.

I think I'm pretty much done reviewing this PR, but I will want to give game.c another pass since it's a big file with many changes.

source/game.c
MeirGavish: Yeah, I think it was especially confusing when the jokers available to shop was a list of joker IDs, so you would have a `joker_idx` to that list to get a `joker_id` but now that it's a bitset I think that confusion is gone.

source/game.c
ricfehr3: Love it, adding.

source/game.c
ricfehr3: Done, I did modify the `remove_held_joker()` function to take the joker_object and the index in the owned joker list to prevent traversing the list more than once.

include/list.h
ricfehr3: agreed

include/list.h
ricfehr3: That's a fair point. I think maybe the phrase, "declare" is more accurate here and in `list_new()`. Going to change that now and you let me know what you think?

include/list.h
ricfehr3: @MeirGavish This was a sneaky attempt from me to introduce Doxygen to this project. The code should be self documenting 100%, and I feel that it does a good job at that right now. But Doxygen on top of it enforces a consistent style and Imo is much easier to parse through once it's in a webpage. 

Totally not necessary though, however it sure is nice to look at once it's done. I'll let you decide on what you wanna do. I personally would like to have the whole repo documented this way (only header files look like this), and I would like to do the work for it. But I'll let you stew on it and decide. I'm good either way.

Here's some pictures of what it looks like with these changes. You can link every variable this way as well.

image

image

tests/list/list_test.c
ricfehr3: Since I've moved the name from `list_new` to `list_declare`. I think that this should just be moved to `list_clear` since there isn't anything `new` to destroy in relation to the `list` object.

include/bitset.h
ricfehr3: I'm also going to use the `declare` terminology here.

source/joker_effects.c
ricfehr3: That is fair. It will return NULL and it makes more sense than directly accessing the internals of the list.

include/bitset.h
ricfehr3: I think I'll choose to rename this for now.

include/list.h
MeirGavish: Maybe "init"?

scripts/get_memory_map.sh
ricfehr3: It automatically gets anything with a pool, so it should work. Here is the output on my host machine:

```sh
Sprite          | 0x03003120 | 2560       | 40         | 32
SpriteObject    | 0x03002d80 | 896        | 40         | 32
Joker           | 0x03002578 | 64         | 36         | 32
JokerObject     | 0x030025b8 | 64         | 36         | 32
Card            | 0x030025f8 | 104        | 36         | 32
CardObject      | 0x03002c80 | 128        | 36         | 32
ListNode        | 0x03002660 | 1536       | 40         | 32
```

source/game.c
ricfehr3: That's fair. I moved that initialization to right after the _owned_joker_list is declared.

source/game.c
ricfehr3: I got you, done.

ricfehr3:
@MeirGavish Huge huge thank you for reviewing this PR. I appreciate your great attention to detail. I went ahead and applied most of your changes. But I'm still working on some. Summarizing here so I can track and keep the convo going and not lost in comments:

1. Address your concern with the list iterator access (list node vs data itself): https://github.com/cellos51/balatro-gba/pull/168#discussion_r2507897390 . I do agree that it should be the data in the list node, but I do have the iterator deleting nodes within the iterator. It would be useful to have the node there and I need to think about it.

2. Doxygen inclusion. Summarized in this comment: https://github.com/cellos51/balatro-gba/pull/168#discussion_r2508303334

3. Adding tests for bitset

include/list.h
ricfehr3: Idk, "init" doesn't have to do with creating the structure. What about `_create()` for everything, `bitset`,`list`,`list_itr`?

include/list.h
ricfehr3: Okay, after thinking on this. I think I'm going to take your suggestions, but try my best scope the iterator into only the `list.c` file.

ricfehr3: Holy smokes, that actually was a really clean refactor. Good call. The access to the `List` internals are well abstracted now and the cumbersome handling of the list nodes is only in `list.c` now.

ricfehr3:
@MeirGavish I've refactored the iterator according to your suggestions. I've also added tests for bitsets.

The last thing to address is the Doxygen stuff, but I'll wait for your word on that.

For the iterator, the internals of the list are "invisible" now. It really cleaned up the cumbersome interface to the iterator, big thank you.

include/list.h
MeirGavish: @ricfehr3 I thought of "init" because it conveys that there's no allocation, just an initialized struct returned but I think "create" is also good, it has the same meaning as "new" but it's a different word so it can be associated with a slightly different functionality.

include/list.h
MeirGavish: Alright, Doxygen is good.

include/list.h
```diff
@@ -76,15 +76,15 @@ typedef struct
 } ListItr;
 
 /**
- * Declare a list.
+ * Create a list.
  *
- * If the list is being re-declared, the list must be freed with @ref list_destroy.
+ * If you are recreating and existing list, the list must be freed with @ref list_clear.
```
MeirGavish: I think it shouldn't be only when re-creating, it's better to not leave unused allocated lists and free the nodes for other lists when the list is no longer needed.

include/list.h
```diff
@@ -169,15 +164,22 @@ int list_get_len(const List* list);
...
  *
  * @param itr pointer to the @ref ListItr
  *
- * @return A pointer to the next @ref ListNode if valid, otherwise return NULL.
+ * @return A pointer to the data pointer at the next @ref ListNode if valid, otherwise return NULL.
+ */
+void* list_itr_next(ListItr* itr);
+
+/**
+ * Remove the previous @ListNode from the iterator
```
MeirGavish: Isn't it the current node? Maybe specify it's the one that was returned from the last call to `list_itr_next()` if it occurred

source/list.c
```diff
@@ -158,13 +157,56 @@ bool list_remove_at_idx(List* list, int n)
...
+    itr->next_node = NULL;
+    return ln;
+}
+
+void list_itr_remove_node_current(ListItr* itr)
+{
+    ListNode* tmp_prev = itr->current_node->prev;
+    _list_remove_node(itr->list, itr->current_node);
+    itr->current_node = tmp_prev;
+}
```
MeirGavish: I think you need a NULL-check here, what happens if the current node is removed from a newly created iterator? What happens if the list is empty?

source/game.c
```diff
@@ -386,10 +422,10 @@ void add_joker(JokerObject *joker_object)
     }
 }
 
-void remove_held_joker(int joker_idx)
+void remove_held_joker(JokerObject* joker_object, int shop_joker_idx)
 {
     // TODO: Extract to on_joker_removed() callback
-    JokerObject* joker_object = list_get(jokers, joker_idx);
+    //JokerObject* joker_object = list_get_at_idx(&_owned_jokers_list, joker_id);
```
MeirGavish: It should be `owned_joker_idx` (or `held_joker_idx` but if we're at it I think the function should be renamed `remove_owned_joker`), the index is not related to the shop.
Also, I actually think it would be better to only pass the index rather than the `joker_object` too and uncomment this line back, it adds a bit of performance cost but it it makes sure there's a single source of truth, the function is not called with mismatching joker object and index, and it doesn't require the caller to always provide both.

source/game.c
```diff
@@ -1376,25 +1413,18 @@ void game_change_state(enum GameState new_game_state)
...
-    jokers_available_to_shop_init();
-
-    // Initialize jokers list
-    if (jokers) list_destroy(&jokers);
-    jokers = list_new(MAX_JOKERS_HELD_SIZE);
+    // Initialize all jokers list once
+    _owned_jokers_list = list_create();
+    _discarded_jokers_list = list_create();
+    _shop_jokers_list = list_create();
+    _joker_scored_itr = list_itr_create(&_owned_jokers_list);
```
MeirGavish, line 1426:
```suggestion
    // TODO: Move this to an initialization of the play scoring states
    _joker_scored_itr = list_itr_create(&_owned_jokers_list);
```

This isn't the most correct place to do this, but I guess it's better to leave this as a TODO for a future refactor.

scripts/get_memory_map.sh
MeirGavish: Cool!

source/game.c
MeirGavish: I added a comment for this.

MeirGavish:
Oh, and I want to note one last thing, I think what we will need to add to this API for #158 is a `list_swap()` function that will allow efficiently swapping elements in the list, rather than a `list_insert()` function. This can be done in a separate issue or as part of #158.

ricfehr3:
> Looks good, just a few small changes left.
> 
> Since concerns over list traversal performance keep coming up I do want to note that if it does prove to be a problem in the future maybe it will be worth it to change some of the lists back to array lists, maybe just pre-allocate a large enough maximum size for them so there's no need to re-allocate, it might be good for small lists. I know we've been working a lot on this, but I just want to note the thoughts that come up over these concerns.

I have no problem switching back to the old method. I think there's value in this method (especially for insertions and `list_swap()` as you said). However, it may be worth it to add an array list (I'm going to call it a **vector** though, similar to c++). 

If the linked list methods work, I would say there's a benefit to only having one list type (less places for problems and less overhead), but I'm more than down to re-implement some of this as a vector too.

source/list.c
ricfehr3: You're totally right, Adding

ricfehr3:
> Oh, and I want to note one last thing, I think what we will need to add to this API for #158 is a `list_swap()` function that will allow efficiently swapping elements in the list, rather than a `list_insert()` function. This can be done in a separate issue or as part of #158.

I think we should do this in another PR, this one is just getting too big IMO.

include/list.h
ricfehr3: I agree, the "previous" was left over from a different commit. Fixed now.

source/game.c
ricfehr3: 1. I agree, changing the name to `remove_owned_joker()`
2. `owned_joker_idx` is better, using it here.
3. In relation to changing the linked list to a vector, I think this is now the only "big" performance cost when comparing to the array method. However, with only 5 jokers, I think its fine and doesn't quite justify needing a vector yet, though, I wouldn't mind writing one to get that extra bit of performance out if needed in the future.

include/list.h
```diff
@@ -85,9 +85,8 @@ typedef struct
 /**
  * Create a list.
  *
- * If you are recreating and existing list, the list must be freed with @ref list_clear.
- *
  * While this function does not allocate memory for the list itself, the list does allocate memory for each element.
+ * So every created list must be freed with @ref list_clear to ensure the list's nodes are deleted properly.
```
MeirGavish, line 89: Nice!

MeirGavish (approved):
Looks great!

As for changing back to vector, right now it doesn't feel like we need to, it doesn't seem too bad for the small lists, but I did want to note it as an option.
I think this list should be used by the deck and hand too when we need to make them more dynamic. There are many benefits for this list for them. For example we see that drawing cards is a critical operation and we want to avoid reallocations there. Also the deck acts pretty much as a stack. These are bigger lists but maybe we can optimize critical parts to avoid traversing them too much.

I will probably merge this soon. Also thinking of #218 but I think this will go first.

MeirGavish:
> > Oh, and I want to note one last thing, I think what we will need to add to this API for #158 is a `list_swap()` function that will allow efficiently swapping elements in the list, rather than a `list_insert()` function. This can be done in a separate issue or as part of #158.
> 
> I think we should do this in another PR, this one is just getting too big IMO.

Yes, definitely makes sense.

source/game.c
MeirGavish: There are a few other places, namely in the row selection, but maybe that can be optimized.

MathisMartin31:
> How did you end up testing the jokers exactly? I would like make sure I'm not breaking any of your work from my rebase on your joker effects work. 

I'm sooo sorry I didn't se your message ><. I'm refreshing my inbox constantly but there's so many messages here it got flooded

I put a message on the testing issue thread I believe you've seen it already. 

You don't have to worry about breaking my pending PRs, yours is very high priority so I'll pull your changes when you're done :3

Also my tests are in a stash locally I didn't plan on merging them ^^

PR #169, opened 2025-10-13
Implement joker testing

MeirGavish:
Allow defining an ID for a joker to always appear in shop and be tested.
For example
`#define TEST_JOKER_ID JOKER_BRAINSTORM_ID`


MeirGavish:
@MathisMartin31 This is what you wanted to ease joker testing? It does require a small tweak in the code but that's what I could easily implement. You can also change the money you start with in the code so you have enough to buy it.

MeirGavish:
@ricfehr3 This does add another function to the list API for a search, I suppose this shouldn't be much of a problem for #168?

ricfehr3:
> @ricfehr3 This does add another function to the list API for a search, I suppose this shouldn't be much of a problem for #168?

Won't be a problem at all. I recommend merging this one first, then let me update my PR to use the new list implementation.

MathisMartin31:
> @MathisMartin31 This is what you wanted to ease joker testing? It does require a small tweak in the code but that's what I could easily implement. You can also change the money you start with in the code so you have enough to buy it.

Yes that's a very big step thank you ^^

PR #172, opened 2025-10-17
Changed music to start after Splash Screen

MeirGavish:
I think it feels much better when the music starts on the main menu when the game actually starts rather than on the splash screen.

The state refactor in #155 made this simple to implement by doing it in the Splash Screen on_exit() function, the only side effect being that the music would restart if we ever returned to the splash screen though it should never happen.

I had to make a few changes since the splash screen was the default state, will comment them.

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -29,11 +29,11 @@
 // Enum value names in ../include/def_state_info_table.h
 enum GameState
 {
-    GAME_STATE_UNDEFINED = -1,
 #define DEF_STATE_INFO(stateEnum, on_init, on_update, on_exit) stateEnum,
 #include "def_state_info_table.h"
 #undef DEF_STATE_INFO
-    GAME_STATE_MAX
+    GAME_STATE_MAX,
+    GAME_STATE_UNDEFINED
 };
 
 enum HandState


include/game.h
```diff
@@ -29,6 +29,7 @@
 // Enum value names in ../include/def_state_info_table.h
 enum GameState
 {
+    GAME_STATE_UNDEFINED = -1,
```
MeirGavish: Since the splash screen was the default state, the first state change would call its on_exit() and start the music, I had to change the default state one so I added an "undefined" value in the enum.
@ricfehr3 @drewjohnson2 I would like your opinion on whether it makes sense to do it like this with the xmacro.

I also changed `game_change_state()` to check for this value.

ricfehr3: I like what you have here. I usually do it after the Xmacro:

```
enum GameState
{
#define DEF_STATE_INFO(stateEnum, on_init, on_update, on_exit) stateEnum,
#include "def_state_info_table.h"
#undef DEF_STATE_INFO
    GAME_STATE_UNDEFINED,
    GAME_STATE_MAX
};
```

I don't see anything wrong with doing it the opposite way, as long as it's consistent when we do it again :sunglasses:

drewjohnson2: I'll need to take a closer look later, but having an explicit value before the expansion of the macro is something I've done before, so at a glance I don't see any issues with what you've done.

ricfehr3: Okay, so, I did realize that here `GAME_STATE_MAX` includes everything **besides** `GAME_STATE_UNDEFINED` in it's value, when doing your method of prefixing with `GAME_STATE_UNDEFINED = -1` whereas my method would have GAME_STATE_MAX include the number of all states **including** `GAME_STATE_UNDEFINED`.

With that, I think it is better to include GAME_STATE_UNDEFINED at the end. My reasoning:

1. It's more idiomatic when X-macros are used for mapping an array. 
  - For example, below a similar change required by it being -1 looks like:

```c
GameState _state = GAME_STATE_UNDEFINED;

... Mapping of GameStates here

// If GAME_STATE_UNDEFINED == -1
void do_update() 
{
    if(_state == GAME_STATE_UNDEFINED)
    {
        return;
    }

    _game_states[_state].on_update();

    ...
}

// If GAME_STATE_UNDEFINED == GAME_STATE_MAX - 1
// You would just have a function that returns nothing so it isn't NULL
// @drewjohnson2 and I found that a noop is best: https://github.com/cellos51/balatro-gba/issues/137#issuecomment-3322485129
void do_update()
{
    // no need to check the state, all GameState's are valid in this map now
    _game_states[_state].on_update();
}
```

ricfehr3 (changes requested):

include/game.h
drewjohnson2: I agree with @ricfehr3.

Something else to consider, although it's not really applicable in this situation, is using enum values as conditions for a loop.  If you have an additional enum value (aside from the MAX) that won't be used as an index it should almost certainly be listed after the MAX enum value because of the following:

Take this enum:
```
// Let's say the expansion of the macro gives us FOO_1 ... FOO_N
enum FooEnum 
{
   FOO_UNDEFINED,
#define SOME_FOO(foo) foo,
#include "sometable.h"
#undef SOME_FOO
   FOO_MAX
};
```

Let's also assume this enum is used to index a lookup table.

You can easily see how the code below would become an issue with the enum definition above.

```
for (size_t i = 0; i < FOO_MAX; i++)
{
    // do something with myFooTable[i];
}
```

Ric's comment should make this pretty obvious, but it was something that I had overlooked when I first peeked at this PR, so I figured I'd point it out in case anyone else was on the slow train like me.

tl;dr

Ric is correct, GAME_STATE_MAX should come at the end, but before GAME_STATE_UNDEFINED.

MeirGavish:
In reply to @ricfehr3's [request](https://github.com/cellos51/balatro-gba/pull/172#discussion_r2441457227) (GitHub doesn't let me reply directly),

I can put it at the end after GAME_STATE_MAX, but for the question of having a check for the state in `game_update()` it wouldn't matter because the sate array `state_info`'s size is determined by the length of the X-macro so GAME_STATE_UNDEFINED (and actually GAME_STATE_MAX too) would still be out of bounds.
I agree and much prefer to not have a check in `game_update()` because it is called on every frame and should be as optimized as possible. I think there isn't really a need for one either way because the checks I added in `game_change_state()` prevents a switch to the undefined state and the initial state is switched on `init()` before there's a chance to call `game_update()`. 
Come to think of it, I think I should upgrade the check in `game_change_state()` to a full bounds checking, I will add another commit for that.

Personally I prefer -1 because it's a more obvious invalid value but I don't mind putting it at the if it's more idiomatic.

MeirGavish:
@ricfehr3 
I missed an invalid access to the game state array so I fixed that, and I moved the undefined value to the end.

PR #176, opened 2025-10-22
Moved pool file out of external definition and replaced it with a flag

MeirGavish:
Just a small change to the pool testing following PR #132.
I moved the pool file definition out of the external definition and replaced it with a flag that the code itself checks.
This separates the responsibilities, making the code responsible for determining which pool file is included rather than have it injected by the makefile, as well as making the production one the default. This also allows IDE parsers (such as IntelliSense) to know that the file is actually being included and not show errors.

### DIFF
--- a/tests/pool/Makefile
+++ b/tests/pool/Makefile
@@ -1,7 +1,7 @@
 
 CC := gcc
 CFLAGS := -I../../include -I. \
-          -g -O3 -Wall -Werror -DPOOLS_TEST_ENV=t 
+          -g -O3 -Wall -Werror -DPOOLS_TEST_ENV=yes
 
 SRC            := pool_test.c ../../source/pool.c
 OUT            := build/pool_test


tests/pool/Makefile
```diff
@@ -1,8 +1,7 @@
-POOLS_DEF_FILE := \"def_test_mempool.h\"
 
 CC := gcc
 CFLAGS := -I../../include -I. \
-          -g -O3 -Wall -Werror -DPOOLS_DEF_FILE=$(POOLS_DEF_FILE) 
+          -g -O3 -Wall -Werror -DPOOLS_TEST_ENV=t
```
ricfehr3: nitpik: Change 't' to 'test' or 'yes'. nbd

ricfehr3 (changes requested):
This is good! 

I'm requesting changes because I missed something in my original PR that I think you can add here. I was going to make another commit but it's very small.

Can you make `-1` here `"UNDEFINED` as it is below in the same file:

https://github.com/cellos51/balatro-gba/blob/e45c1eeb22a84ab8cb764615798609e0019620d7/source/pool.c#L38

Otherwise, once you change that go ahead and merge. :+1:

tests/pool/Makefile
MeirGavish:
```suggestion
          -g -O3 -Wall -Werror -DPOOLS_TEST_ENV=yes
```

PR #178, opened 2025-10-25
Added option for a second test joker

MeirGavish:
Added option for a second test joker so 2 can be tested at the same time to see interactions. A single test joker can still be defined.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -199,8 +199,8 @@ static int hands = 0;
 static int discards = 0;
 
 static int round = 0;
-static int ante = 1;
-static int money = 4;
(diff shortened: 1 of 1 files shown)


source/game.c
```diff
@@ -199,8 +199,8 @@ static int hands = 0;
 static int discards = 0;
 
 static int round = 0;
-static int ante = 1;
-static int money = 4;
+static int ante = 0;
+static int money = 0;
```
MeirGavish, line 203: I changed all the dummy initial values to 0 to have a single place of initialization since the actual values are being initialized in game_init()

Issue #179, opened 2025-10-26
Create Tests Workflow


ricfehr3:
Currently in our CI we are lumping everything into one github action workflow for builds and tests. We should separate tests and build. See @MeirGavish's comment: https://github.com/cellos51/balatro-gba/pull/132#discussion_r2462722987

PR #160, opened 2025-10-04
Refactor of scoring flow and first steps for Joker callbacks

MathisMartin31:
Hello again ^^

Creating this as part of discussion #103

I've been tweaking the code for scoring Cards/Jokers for a few days trying to implement retriggers, and I found that Joker callbacks were kind of the first step towards that goal, so I added what I think could be a way to do it.

Since this is a fairly large PR, and because it contains code from my other PR #159 , I'm putting this as a draft for now so we can discuss things further.

Also, discussion #156 is blocking, or at least relevant and would make the variables I added in `game.c` a lot cleaner by encapsulating them in a nice struct.

Issue #141, opened 2025-09-21
Add versioning


ricfehr3:
Some form of versioning would be helpful for devs. I don't think there needs to be a specific version scheme yet, but as people are expected to build the ROM themselves right now, some form of reliable tracking would be helpful as bugs are reported.

I believe it would be helpful to have a script that saves the current commit hash and a (dirty/clean) state of the hash into an easily 'greppable' variable. This could be used in a few ways:
- A script could be provided to grep the ROM for the string to be given for bug reports.
- It can be displayed on the splash screen

ricfehr3:
Did an experiment:

Added this:

```c
__attribute__((used))
const char gba_version[] = "GBA_VERSION:0x123456789ABCDEF";
```

`used` indicated to compile and skip optimizations.

Next, I simply:

```sh
strings build/balatro-gba.gba | grep GBA_VERSION
GBA_VERSION:0x123456789ABCDEF
```

So, the makefile can be updated to get the hash:



```c
GIT_DIRTY := $(shell git diff-index --quiet HEAD -- || echo "-dirty")
GIT_HASH := $(shell git rev-parse --short HEAD)
CFLAGS  += -DGIT_HASH=\"$(GIT_HASH)\" -DGIT_DIRTY=\"$(GIT_DIRTY)\"

```

I'm on my lunch right now. But I'm going to add `-D` for git hash and git dirty status and make a PR tonight. :) 

Please assign this to me if you get the chance.

PR #182, opened 2025-10-27
Add nightly rom releases

ricfehr3:
This allows nightly releases to be pushed to a specific release URL. To complete this PR we need to.

1. Add a release manually, named something with "nightly release" and tagged as `pre-release`
2. Go to the url: https://api.github.com/repos/cellos51/balatro-gba/releases and get the right id's
3. Update the locations in this file.
4. Try to reuse build workflow instead of repeating steps

Note that this file is mostly a copy from the example right now: https://github.com/marketplace/actions/deploy-nightly

@MeirGavish If you're cool with this, I can move it out of draft and quickly do the above two steps and cleanup this file whenever. It's working in my fork: https://github.com/ricfehr3/balatro-gba/releases

Related discussion: https://github.com/cellos51/balatro-gba/discussions/180

### DIFF
--- a/.github/workflows/nightly_release_ci_workflow.yml
+++ b/.github/workflows/nightly_release_ci_workflow.yml
@@ -26,7 +26,7 @@ jobs:
           path: build
 
       - name: Package Release
-        run: mv build/balatro-gba.gba build/balatro-gba_nightly_$(date +'%F')_$(cat build/githash.txt).gba && zip release.zip build/*.gba
+        run: mv build/balatro-gba.gba build/balatro-gba_nightly_$(date +'%+4Y%m%d')_$(cat build/githash.txt).gba && zip release.zip build/*.gba
 
       - name: Deploy release
         uses: WebFreak001/deploy-nightly@v3.2.0


.github/workflows/nightly_release_ci_workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Build the project
+        run: docker compose -f docker-compose.yml run --rm balatro sh -c "make -j$(nproc) && ./scripts/get_memory_map.sh"
+
+      - name: Package Release
+        run: zip ./release.zip ./build/balatro-gba.gba build/balatro-gba.elf
+ 
+      - name: Deploy release
+        uses: WebFreak001/deploy-nightly@v3.2.0
+        with:
+          upload_url: https://uploads.github.com/repos/ricfehr3/balatro-gba/releases/257345890/assets{?name,label} # find out this value by opening https://api.github.com/repos/<owner>/<repo>/releases in your browser and copy the full "upload_url" value including the {?name,label} part
```
ricfehr3: change this

.github/workflows/nightly_release_ci_workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+          release_id: 257345890 # same as above (id can just be taken out the upload_url, it's used to find old releases)
```
ricfehr3: update this

.github/workflows/nightly_release_ci_workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Export UID/GID for Compose
+        run: |
+          echo "UID=$(id -u)" >> "$GITHUB_ENV"
+          echo "GID=$(id -g)" >> "$GITHUB_ENV"
+
+      - name: Run Tests
+        run: cd tests && ./run_tests.sh
+
+      - name: Build the project
+        run: docker compose -f docker-compose.yml run --rm balatro sh -c "make -j$(nproc) && ./scripts/get_memory_map.sh"
```
MeirGavish:
```suggestion
        run: docker compose -f docker-compose.yml run --rm balatro sh -c "make -j$(nproc)
```

I guess there's no need for the memory map script in this one?

MeirGavish (commented):
So finding out the aliexpress scam on discussion #157 got the rom from our release page makes me a little uncomfortable approving this, but I guess the real thought is to make this available to the people who requested so they don't have to go through the trouble of compiling anymore.

GreenNinja2525:
> So finding out the aliexpress scam on discussion #157 got the rom from our release page makes me a little uncomfortable approving this, but I guess the real thought is to make this available to the people who requested so they don't have to go through the trouble of compiling anymore.

@MeirGavish I think this should be left to forks, although the ROM from AliExpress had the splash screen, I still think it was bad exposure. I don't feel comfortable with this pull request either, but I don't have any say here. Honestly, just do what you want, we are already in too deep to turn back now lol

ricfehr3:
I'll drop my two cents here on this whole thing as well.

- **If we want to drop this PR for any reason, that's cool**, I'm not offended. It really is just a copy and paste of the example.

- I don't see how nightly builds would have anything to do with the aliexpress ROM:
  1. If someone has a little skill and a $30 cart flasher, they can build and load ROMs all day without nightly releases. Plus, on our next official release they could do the same thing.
  2. I don't see what "bad exposure" came from the aliexpress ROM. You can't control someone else flashing and selling those in another country. The splash screen that is present on the ROM clearly says, "Hey, if you bought this, you got scammed. This should be free and it is at this website." 
  
- There is one big concern for me with nightly releases and that is, rather than having only a handful of releases to consider when a bug report comes in, there would be seemingly a lot more nightly versions in the wild. That could be a bad thing, I personally see it as a positive. More people playing the ROM means more people can spot bugs. That does put quite the burden on us though to keep up with all the varied versions and bugs, but each ROM now has it's git commit hash in it, so we aren't completely lost.

- This project is a $0 and opensource fan game. I feel that we should focus on the experience of the fans and players who just want to play the game and not on how people can be assholes and scam people with it. The ROMs from this project are shareware, and there should be no effort to gatekeep access to the ROMs for what a scammer _may_ do.

That all being said, I can understand why nightly releases may not be a correct solution due to the disparity in the versions that are released, but we really should keep the focus on making it as easy as possible to pick up and play. Maybe just another release once the joker effects are updated in #160 ? However, I just want to say **I strongly disagree with letting a scammer ruin others access to recent versions out of fear**. I'm more concerned with people like in #180 , who don't have a way to make a new version, but want to play the most recent build.

GreenNinja2525:
@ricfehr3 Those are some very valid reasons. I'm all for nightly releases now after reading your comment. Bug fixing and play testing should be put first as you mentioned. I completely agree that nightly releases would be beneficial for this. Good reasoning @ricfehr3

MeirGavish:
@ricfehr3 Yeah, I agree, we shouldn't let that dictate our decisions. Open source means open to everyone. We can have nightly releases.

ricfehr3:
@MeirGavish @GreenNinja2525 Glad we are on the same page :) . 

For this PR, I also decided to separate the build workflow and call it in separate workflows. 

This PR separates the workflows as:
- Build CI
- Deploy Nightly
- Release On Version Tag (separated from  Build CI workflow)
- Run Tests (this covers #179)

This is currently setup to build nightly and only deploy if it's a new build. It deploys a zipfile with just the `.gba` file to: https://github.com/cellos51/balatro-gba/releases/tag/nightly

I chose to just package the `.gba` file to avoid any confusion for the end user. If we want something else there, it's no biggie. I thought of including the `.elf` file, but I feel that would confuse people. Anyway, **are there any other files we need in the nightly release?** I'm considering putting in the git hash for now, that will match our baked-in git hash from `version.c` from #141 

Speaking of bug reporting. I can make an issue if you approve @MeirGavish , but I think the easiest way to get good bug reporting is to have an info screen for someone to take a screenshot, picture, or easily copy. I'm thinking only at the splash screen we can watch for holding down all the buttons or something, and it instantly drops you into the info screen. With the state rework the @drewjohnson2 did from #155 , I think it wouldn't be too difficult of an addition. I'd rather not rely on a script or anything. I want someone to be able to get us a good bug report from just their phone.

MeirGavish:
And the info screen would include the git hash?

ricfehr3:
> And the info screen would include the git hash?

Exactly. That's the minimum we should have. Definitely can discuss what else can be on that screen. 

I figure it will be a copied and modified splash screen. So we can squeeze that much text for reporting into the screen.

Maybe it should display the most recent tag as well so we can quickly distinguish how old the ROM is.

MeirGavish:
I currently can't think of anything to put in the nightly release other than the rom and git hash. I'll just approve and you can merge this when you feel it's good enough.

README.md
```diff
@@ -1,5 +1,13 @@
 # Balatro-GBA
 
+---
+
+<a href="https://github.com/cellos51/balatro-gba/releases/">
+  <img src="https://img.shields.io/badge/Download_the_ROM_from_github-8A2BE2?&logo=github" alt="Download ROM" width="350">
+</a>
+
+---
```
ricfehr3: @MeirGavish This is one change I wanted to run by you before I merged. I can drop the commit if you want, but I just put a big ol' download button in the README.

Also just going to put this here. I'm going to make a bug report template and add a wiki page on how to file a bug report. The ROMS in the zip are named "balatro-gba_nightly_(date)_(githash).gba", so the name of the file is self-documenting as long as it's not changed. Best we can do before having a screen with the info on it.

ricfehr3: I'm going to add this later if we want, but I'll just drop it for now and merge.

MeirGavish: Yeah, the readme should point to the releases so people understand they don't have to compile the code. It can just be a link to the releases "ROMs can be downloaded in the releases tab"

PR #189, opened 2025-10-29
Changed the order of suits to match Balatro

MeirGavish:
When ordering by suit in the original Balatro, the order is Diamonds, Clubs, Hearts, Spades (from right to left).

image

Our order had Diamonds and Hearts swapped, so I changed it to match Balatro.
I also needed to change their order in the spritesheet since the suit index is used to match the sprite.


ricfehr3 (approved):
Looks great. Very simple change, I can't see any issues with just merging it.

MeirGavish:
Cool, I'll leave it open for the night in case anybody else wants to take a look and merge it in tomorrow

PR #195, opened 2025-11-01
Update issue template for bug reports

ricfehr3:


ricfehr3:
> Approving, you can merge if you want. We can make changes later but it's better than the default.

Awesome, tysm for the quick reply!

PR #196, opened 2025-11-01
Add download button

ricfehr3:
Add a big Download button. I also added something similar to file a bug report on the wiki page. This PR should complete the whole CI -> Release -> Bug Submit cycle going forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,5 @@
 # Balatro-GBA
 
----
-
-<a href="https://github.com/cellos51/balatro-gba/releases/">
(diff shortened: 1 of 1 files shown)


MeirGavish:
That's a big button. I only think it should be after the disclaimer and encouragement to buy the official release. Maybe right before the build instructions?

ricfehr3:
> That's a big button. I only think it should be after the disclaimer and encouragement to buy the official release. Maybe right before the build instructions?

That sounds good to me, definitely want to give all credit to the actual game first.

Issue #201, opened 2025-11-01
Add a licence


gillo04:
To avoid future confusion on the proper use one can make of this project and in what ways it is open, a LICENCE file should be added at the root of the project

ricfehr3:
Good point. Licenses are important. 

My opinion is to make it as permissive as possible. I would first choose the [unlicense](https://choosealicense.com/licenses/unlicense/) next would be [MIT](https://choosealicense.com/licenses/mit/).

Both of these provide those important "AS IS" sections for liability. I prefer **UNLICENSE** because it gives up the code fully to the public domain, and doesn't require other projects that use it to also use the license, whereas the MIT license requires subsequent projects use the MIT license as well. 

Making a PR for UNLICENSE

---

Edit: I made a custom license file that uses UNLICENSE for the code and creative commons v4 for the art: https://creativecommons.org/licenses/by/4.0/

@MathisMartin31 @GreenNinja2525 I would like your opinion on both licenses since you're both very involved. But you've both done a lot of art for the game (please tag anyone else if I'm missing people). I put the first license **"CC BY"** from https://creativecommons.org/share-your-work/cclicenses/ for the art assets, but y'all definitely should be choosing that for the art, I have no say. Just let me know and I'll add it.

MathisMartin31:
Creative Common is fine by me ^^.

I technically made the art I posted here, but I'm not comfortable claiming ownership of it when I just copied Balatro assets.

MeirGavish:
See [comment in associated issue](https://github.com/cellos51/balatro-gba/pull/202#issuecomment-3636715193).

PR #202, opened 2025-11-01
Add Unlicense

ricfehr3:
This is for #201 

Making this PR on my suggestion of declaring this project public domain without any restrictions or liability on our part under [unlicense](https://choosealicense.com/licenses/unlicense/)


GreenNinja2525:
@ricfehr3 I noticed one thing in the license that kinda defeats the purpose of the splash screen/the intentions of this project. It says that anyone is free to SELL this project and use it for commercial means, which is against the project's intentions. That was the whole point of the Aliexpress issue as the project was being sold on a re-pro cart. I know this is an open source project, but it should not be sold as then it gives the impression that it is an official de-make of Balatro, which it is not. Perhaps some other licensing could be used instead? This also isn't technically able to be placed into public domain as it violates the derivative works side of legality. This project is mimicking Balatro, but isn't Balatro (if that makes sense) and since the graphics look similar to the real game, then it is a derivative works violation (but that is why we weren't selling it). That is like somebody drawing a sprite of Mario and making a de-make of a Mario game and trying to sell it, it just isn't legal to sell it, hence why fan games are usually never sold. We cannot give the authority to sell the game or have it be in the public domain as we don't own the IP to Balatro.

GreenNinja2525:
@ricfehr3 What about a creative commons license? This one -Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International (CC BY-NC-SA 4.0)- says that we must be credited for the work if it is remixed or edited in any way outside this repository, but it cannot be sold commercially or used to receive money through sales. It also states that people that distribute this project outside the repository and edit sprites or anything within it, the same license must be present there as well. Please look into this further in case I am mistaking or misunderstanding this license, thanks!

GreenNinja2525:
Also, the license states that the project cannot be taken (unless it is by the IP owners) and converted to closed source or be held for payment as it is still an open source project, but cannot be used for monetary gain.
EDIT:
Here is the license itself, please read through it and make sure I am not misunderstanding anything:
[Attribution-NonCommercial-ShareAlik.txt](https://github.com/user-attachments/files/23286508/Attribution-NonCommercial-ShareAlik.txt)

ricfehr3:
Yeah, I hear you @GreenNinja2525 . I agree too, thinking on it, I'm not comfortable at all with these license terms now. I initially thought, "Just the code should be fine to license as open source", however I see it's pretty freaking important to distinguish someone taking a piece of code vs someone selling the entire product as one ROM. That's my bad. 

The big thing I'm realizing is that even with the creative commons license, there is no right of this project to distribute it under any license because it's a fan game. If there was an engine that gbalatro is built on, then the engine could be the only thing you could license. Similarly, I would assume new art not derived from anything would be allowable too. But the creative commons license clearly states:

```txt
Creative Commons public licenses provide a standard set of terms and
conditions that creators and other rights holders may use to share
original works of authorship and other material subject to copyright
and certain other rights specified in the public license below
```

We are simply not the "creators and rights holders" that this applies to. Unfortunately I think the only way to really have a fan-game licensed is to have official support. Not sure where to go from here.

This is very much related to #171 . It sounds like the only way to preserve the longevity and independence of this project is to have completely separate set of trademark attached things. Art/Names/Style. Sort of like freedoom does: https://freedoom.github.io/ or your average "knock-off" mockbuster kids movie. 

This is legal:

GreenNinja2525:
@ricfehr3 Ah, I see it now. We are the creators of this project, but not the rights holders as you mentioned. We can apply an MIT license to the code, but the art/game design must remain unlicensed then. Applying the MIT license to the code means that the code can be used commercially, but the assets cannot as they are unlicensed. This means if someone sells it, it is illegal (on a technicality). That way the GBA community can use the code and refer to it, but the art remains an unlicensed mess (we are technically violating the derivative works laws)

GreenNinja2525:
@ricfehr3 Leaving the art unlicensed is also good for us if we get a take down as we can argue we never claimed to own the assets that infringe on the original game. That means if this project gets sold as a flash cart, that means we are not at fault as the seller violated the distribution laws by including the infringing assets. That is another point we can use to argue in a takedown event. It seems un-licensing the art makes our project more legal, if that makes sense. Just to be clear, unlicensed does not mean uncredited, the artists can still be credited but it just means we don't own the designs.

ricfehr3:
> > Yeah, I hear you @GreenNinja2525 . I agree too, thinking on it, I'm not comfortable at all with these license terms now. I initially thought, "Just the code should be fine to license as open source", however I see it's pretty freaking important to distinguish someone taking a piece of code vs someone selling the entire product as one ROM. That's my bad.
> > The big thing I'm realizing is that even with the creative commons license, there is no right of this project to distribute it under any license because it's a fan game. If there was an engine separate to balatro, then the engine could be the only thing you could license. Similarly, I would assume new art not derived from anything would be allowable too. But the creative commons license clearly states:
> > ```
> > Creative Commons public licenses provide a standard set of terms and
> > conditions that creators and other rights holders may use to share
> > original works of authorship and other material subject to copyright
> > and certain other rights specified in the public license below
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We are simply not the "creators and rights holders" that this applies to. Unfortunately I think the only way to really have a fan-game licensed is to have official support. Not sure where to go from here.
> 
> Ah, I see it now. We are the creators of this project, but not the rights holders as you mentioned. We can apply an MIT license to the code, but the art/game design must remain unlicensed then. Applying the MIT license to the code means that the code can be used commercially, but the assets cannot as they are unlicensed. This means if someone sells it, it is illegal (on a technicality). That way the GBA community can use the code and refer to it, but the art remains an unlicensed mess (we are technically violating the derivative works laws)

Yeah... and we aren't lawyers either, but it sounds like MIT or similar for the code is the only thing. ~I'm looking for licenses that would cover our terms better for specific non-commercial use.~ Can't do that as it makes it non-free. There's also some direct references to balatro in the code, and that sounds like it be problematic to license :shrug: 


> @ricfehr3 Leaving the art unlicensed is also good for us if we get a take down as we can argue we never claimed to own the assets that infringe on the original game. That means if this project gets sold as a flash cart, that means we are not at fault as the seller violated the distribution laws by including the infringing assets. That is another point we can use to argue in a takedown event. It seems un-licensing the art makes our project more legal, if that makes sense. Just to be clear, unlicensed does not mean uncredited, the artists can still be credited but it just means we don't own the designs.

I agree. I do think if there was a complete art asset redo, the code would definitely still be considered an original work, as it's made from scratch. I think it would be? If every joker was white, the background didn't move, and it was called "GBA Rouge Cards" or something, would it be a completely libre game?

I'm asking that question because if the answer is "no, it's still balatro" then there's no point and we should just ask LocalThunk :shrug:

GreenNinja2525:
>I agree. I do think if there was a complete art asset redo, the code would definitely still be considered an original work, as it's made from scratch. I think it would be? If every joker was white, the background didn't move, and it was called "GBA Rough Cards" or something, would it be a complete libre game?
>

 @ricfehr3 Technically yes, but the values and mechanics for how jokers worked would needed changed. The idea of a poker game with hands that use modifiers is not a copyrighted idea. But the specifics would need changed. Perhaps the shop could be changed, planet cards be reworked, tarot cards be changed, then yes, that would work. That is where something I think Meir mentioned about a cribbage based card game could be fun. We could make our own card game for GBA, that would be awesome. A cribbage based card game that uses modifiers essentially.

ricfehr3:
@GreenNinja2525 Yeah, I agree, it would need to be changed quite a bit. I'm going to leave this PR to cook a bit and have more people chime in. Also going to link this back to that original discussion. 

My opinion right now is we can't license it so don't. Not before a serious code rework and complete asset redo.

GreenNinja2525:
> @GreenNinja2525 Yeah, I agree, it would need to be changed quite a bit. I'm going to leave this PR to cook a bit and have more people chime in. Also going to link this back to that original discussion.
> 
> My opinion right now is we can't license it so don't. Not before a serious code rework and complete asset redo.

Just a small thing with the cribbage idea, perhaps this is the visual style to use. I'll stop commenting here now, just wanted to throw this out there. Nice discussion @ricfehr3 

Cribbage 5

MeirGavish:
I have a few thoughts about this but I've been having a hard time diverting my attention to this issue, I will write them out tomorrow.

MeirGavish:
So I'm not a lawyer and these are just my thoughts and opinions. To get a better idea we need to learn more about the legalities and maybe consult with an actual lawyer (~~I actually have a friend who might be able to give some input~~ He seems to be quite unavailable currently 😅).

I want to separate the mechanics of Balatro and the code. 
The mechanics of the game is the gameplay itself and is more abstract - playing poker hands, jokers etc. We don't own this, we didn't design this game. There can be many implementations to these same abstract mechanics - the game can even be [played in real life with real cards](https://youtu.be/9NlfgQieTaE).
This project is an implementation, specifically on the GBA, and this adds new creative value. The code is made from scratch and built to work on the GBA with all its limitations, and there is a value we create here when making this implementation. The art is made to look like the original but it is also made from scratch, needs to be in a smaller resolution and use a more limited color palette and there is new creative value in that too. 
We can't claim full ownership of this demake but I think remakes and demakes have a creditable creative value as derivative work like cover songs and remixes and we can claim authorship to the implementation.

It seems quite problematic to license this without permission but I do want in some way to prohibit selling this for now. In the unlikely event the original owners want to make this an official port we will need to be able to change that rule and also still claim ownership/authorship to the implementation.

I do love that this is open source and freely available, that's what gave me the chance to be a part of it, but it doesn't seem like we can bind that in a license.

As for Cribbage, right now I want to keep this project Balatro focused, that was the original idea, make Balatro work on the GBA, and personally I want to continue working on that. However I still think it's a nice idea to make a Cribbage focused game in a fork, or hell even as a game for modern platforms that can be sold if you want, but I don't know Cribbage and haven't given it much thought.

MeirGavish:
Okay, so from looking things up a bit it seems like game mechanics are not copy-rightable but everything else including names, imagery, UI, and music is. I would argue that this is transformative work but this is definitely a gray area (and we should comply with a DMCA) since we are explicitly basing this on Balatro. So we really can't license this and can't explicitly allow anyone to use it so I'm closing this and the associated issue.

PR #203, opened 2025-11-02
Junk directories on release zip

ricfehr3:
The zipfiles from the release workflow have the `build/` directory still present. 

I'm adding a `-j` (`--junk-paths`) and also removing the `.zip` from the command. 

This isn't fixing anything broken, it's just annoying me, lol.

### DIFF
--- a/.github/workflows/nightly_release_ci_workflow.yml
+++ b/.github/workflows/nightly_release_ci_workflow.yml
@@ -29,7 +29,7 @@ jobs:
           path: build
 
       - name: Package Release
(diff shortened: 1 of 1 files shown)


MeirGavish (commented):
I approved without looking but now that I'm looking, this doesn't remove the .zip extension from the released file, right? Because otherwise it could be confusing people how they should open the file.

ricfehr3:
> I approved without looking but now that I'm looking, this doesn't remove the .zip extension from the released file, right? Because otherwise it could be confusing people how they should open the file.

No worries. It still adds the `.zip` automatically, so no difference in the name of the archive. I removed it to better follow the examples in the man page.

MeirGavish:
Cool, merging

PR #205, opened 2025-11-02
Implement cards held in hand effects

MathisMartin31:
Picking up from where I left my last PR, this will add the ability for cards held in hand to activate the associated Jokers like Raised Fist, Shoot the Moon or Mime.

As of creating this PR, only Raised Fist has been updated to show this works:

https://github.com/user-attachments/assets/87255aa9-6135-4cae-a52f-a6a9c312713a

### DIFF
--- a/include/joker.h
+++ b/include/joker.h
@@ -49,11 +49,11 @@
 // Joker per Joker basis (see if it's there, then do something, e.g. Pareidolia, Baseball Card)
 enum JokerEvent
 {
-    JOKER_EVENT_ON_HAND_PLAYED,     // Triggers only once when the hand is played
+    JOKER_EVENT_ON_JOKER_CREATED,   // Triggers only once when the Joker is created, mainly used for data initialization
+    JOKER_EVENT_ON_HAND_PLAYED,     // Triggers only once when the hand is played and before the cards are scored
     JOKER_EVENT_ON_CARD_SCORED,     // Triggers when a played card scores (e.g. Walkie Talkie, Fibonnacci...)
     JOKER_EVENT_ON_CARD_SCORED_END, // Triggers after the card has finishd scoring (e.g. retrigger Jokers)
     JOKER_EVENT_ON_CARD_HELD,       // Triggers when going through held cards
-    //JOKER_EVENT_ON_CARD_HELD_END,   // Triggers after when going through held cards (e.g. retrigger)
     JOKER_EVENT_INDEPENDENT,        // Joker will trigger normally, when Jokers are scored (e.g. base Joker)
     JOKER_EVENT_ON_HAND_SCORED_END, // Triggers when entire hand has finished scoring (e.g. food Jokers)
     JOKER_EVENT_ON_HAND_DISCARDED,  // Triggers when discarding a hand
@@ -107,12 +107,10 @@ typedef struct  // These jokers are triggered after the played hand has finished
 } JokerEffect;
 
 typedef JokerEffect (*JokerEffectFunc)(Joker *joker, Card *scored_card, enum JokerEvent joker_event);
-typedef void (*JokerCallbackOnCreated)(Joker *joker);
 typedef struct {
     u8 rarity;
     u8 base_value;
     JokerEffectFunc joker_effect;
-    JokerCallbackOnCreated on_joker_created;
 } JokerInfo;
 const JokerInfo* get_joker_registry_entry(int joker_id);
 size_t get_joker_registry_size(void);
--- a/source/game.c
+++ b/source/game.c
@@ -2232,8 +2232,9 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
 
                         if (*played_selections == 0)
                         {
-                            play_state = PLAY_BEFORE_SCORING;
+                            _joker_scored_itr = list_itr_create(&_owned_jokers_list);
                             timer = TM_ZERO;
+                            play_state = PLAY_BEFORE_SCORING;
                         }
                     }
 
@@ -2250,14 +2251,13 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
                     {
                         return;
                     }
-
-                    _joker_scored_itr = list_itr_create(&_owned_jokers_list);
-                    play_state = PLAY_SCORING_CARDS;
                     
                     if (card_object_is_selected(played[i]) && played_top - i >= *played_selections)
                     {
                         played_y -= int2fx(10);
                     }
+
+                    play_state = PLAY_SCORING_CARDS;
                     break;
 
                 case PLAY_SCORING_CARDS:
@@ -2327,8 +2327,7 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
                             }
                         }
 
-                        // advance state after going past the last card (exited the loop without returning)
-                        play_state = PLAY_SCORING_JOKERS;
+                        play_state = PLAY_SCORING_HELD;
                         // reuse these variables for held cards
                         _joker_scored_itr = list_itr_create(&_owned_jokers_list);
                         scored_card_index = hand_top;
--- a/source/joker.c
+++ b/source/joker.c
@@ -15,6 +15,7 @@
 
 #define JOKER_SCORE_TEXT_Y 48
 #define HELD_CARD_SCORE_TEXT_Y 108
+#define MAX_CARD_SCORE_STR_LEN 2
 #define NUM_JOKERS_PER_SPRITESHEET 2
 
 static const unsigned int *joker_gfxTiles[] =
@@ -145,7 +146,9 @@ Joker *joker_new(u8 id)
     joker->value = jinfo->base_value + edition_price_lut[joker->modifier];
     joker->rarity = jinfo->rarity;
     joker->data = 0;
-    jinfo->on_joker_created(joker);
+
+    // initialize persistent Joker data if needed
+    jinfo->joker_effect(joker, NULL, JOKER_EVENT_ON_JOKER_CREATED);
 
     return joker;
 }
@@ -159,7 +162,7 @@ void joker_destroy(Joker **joker)
 JokerEffect joker_get_score_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
 {
     const JokerInfo *jinfo = get_joker_registry_entry(joker->id);
-    if (!jinfo || jinfo->joker_effect == NULL) return (JokerEffect){0};
+    if (!jinfo) return (JokerEffect){0};
 
     return jinfo->joker_effect(joker, scored_card, joker_event);
 }
@@ -256,7 +259,7 @@ void set_and_shift_text(char* str, int* cursor_pos_x, int* cursor_pos_y, int col
     tte_write(str);
 
     // + 1 For space
-    const int joker_score_display_offset_px = (strlen(str) + 1)*TTE_CHAR_SIZE;
+    const int joker_score_display_offset_px = (MAX_CARD_SCORE_STR_LEN + 1) * TTE_CHAR_SIZE;
     *cursor_pos_x += joker_score_display_offset_px;
 }
 
@@ -281,19 +284,19 @@ bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum
     *retrigger = joker_effect.retrigger;
     // joker_effect.message will have been set if the Joker had anything custom to say
 
-    int cursorPosX = 8; // Offset of one tile to better center the text on the card
+    int cursorPosX = TILE_SIZE; // Offset of one tile to better center the text on the card
     int cursorPosY = 0;
-    if (joker_event != JOKER_EVENT_ON_CARD_HELD)
-    {
-        cursorPosX += fx2int(joker_object->sprite_object->x);
-        cursorPosY = JOKER_SCORE_TEXT_Y;
-    }
-    else
+    if (joker_event == JOKER_EVENT_ON_CARD_HELD)
     {
         // display the text on top of the card instead of below the Joker for Held Cards effects
         // scored_card cannot be NULL here because of the joker event
         cursorPosX += fx2int(card_object->sprite_object->x);
         cursorPosY = HELD_CARD_SCORE_TEXT_Y;
+    }
+    else
+    {
+        cursorPosX += fx2int(joker_object->sprite_object->x);
+        cursorPosY = JOKER_SCORE_TEXT_Y;
     }
 
     if (joker_effect.chips > 0)
@@ -321,7 +324,7 @@ bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum
         set_and_shift_text(score_buffer, &cursorPosX, &cursorPosY, TTE_YELLOW_PB);
     }
     // custom message for Jokers (including retriggers where Jokers will say "Again!")
-    if (joker_effect.message[0] != '\0') // Message is not empty
+    if (joker_effect.message != NULL && joker_effect.message[0] != '\0') // Message is not empty
     {
         set_and_shift_text(joker_effect.message, &cursorPosX, &cursorPosY, TTE_WHITE_PB);
     }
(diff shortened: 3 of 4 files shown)


MeirGavish:
The video looks good! I assume it also means that Raised Fist will activate after say, Half Joker no matter their order in held jokers.

MathisMartin31:
Thanks ^^

Yeah I should test with Joker triggered by other events, but "on card held" Jokers will be triggered between "on card scored" and "independent" Jokers (just as in Balatro), meaning Raised Fist would activate before Half Joker, but after something like Even Steven, and according to the order of both Cards in hand and owned Jokers (from left to right)

MeirGavish:
Right, it should be before, had a mistake in my previous comment

MathisMartin31:
I tested it, I think this works as intended ^^

https://github.com/user-attachments/assets/26a63dfe-f80b-43c1-8c11-ee5d46967add

MeirGavish:
Yep, looks good

MathisMartin31:
I tested the currently implemented Jokers and they should work fine (except Shoot the Moon because we still can't spawn it in the game).

I also experimented with putting the effect text related to the held cards above them (how it's done in Balatro, the same way scored cards and Jokers have their text next to them) but it may be a bit cramped down there, what do you think?

https://github.com/user-attachments/assets/11b2c72d-e179-49e9-bffc-0d0b5dc38dd7

source/game.c
```diff
@@ -2225,18 +2242,46 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
-                            
+                
+                case PLAY_SCORING_HELD:
+
+                    if (i == 0 && (timer % FRAMES(30) == 0) && timer > FRAMES(40))
+                    {
+                        tte_erase_rect_wrapper(HELD_CARDS_SCORES_RECT);
+
+                        // Go through all held cards and see if they activate Jokers
+                        for ( ; scored_card_index >= 0; scored_card_index--)
```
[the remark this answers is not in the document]
MathisMartin31: Soooo, I guess it's all good then

source/joker_effects.c
```diff
@@ -400,23 +400,40 @@ static JokerEffect blue_joker_effect(Joker *joker, Card *scored_card, enum Joker
...
+            *p_lowest_value_index = 0;
+            CardObject** hand = get_hand_array();
+            int hand_size = hand_get_size();
+            for (int i = 0; i < hand_size; i++ )
+            {
+                u8 value = card_get_value(hand[i]->card);
+                u8 lowest_value = card_get_value(hand[*p_lowest_value_index]->card);
+                if (lowest_value > value)
+                {
+                    *p_lowest_value_index = i;
```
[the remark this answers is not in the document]
MathisMartin31: I tweaked Copilot's proposition a bit, as it would have segfaulted with a hand size of 0, will test it a bit later I'm at work rn ><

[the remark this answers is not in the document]
MathisMartin31: @MeirGavish I just tested Raised Fist, all seems to be working as intended ^^

source/joker.c
```diff
@@ -281,35 +282,49 @@ bool joker_object_score(JokerObject *joker_object, Card* scored_card, enum Joker
     *retrigger = joker_effect.retrigger;
     // joker_effect.message will have been set if the Joker had anything custom to say
 
-    int cursorPosX = fx2int(joker_object->sprite_object->x) + 8; // Offset of 16 pixels to center the text on the card
+    // display the text on top of the card instead of below the Joker for Held Cards effects
+    int cursorPosX = 8; // Offset of 16 pixels to center the text on the card
```
ricfehr3: Does this comment make sense anymore? If so, I don't really understand it. It sounds like 16 pixels should center it, but we use 8?

source/joker_effects.c
```diff
@@ -425,17 +443,11 @@ static JokerEffect reserved_parking_joker_effect(Joker *joker, Card *scored_card
...
-    CardObject** hand = get_hand_array();
-    int hand_size = hand_get_size();
-    for (int i = 0; i < hand_size; i++ )
+    if ((random() % 2 == 0) && card_is_face(scored_card))
     {
-        if ((random() % 2 == 0) && card_is_face(hand[i]->card))
-        {
-            effect.money += 1;
-        }
+        effect.money += 1;
```
ricfehr3: nitpik `++` instead of `+= 1`

source/joker.c
```diff
@@ -281,35 +282,49 @@ bool joker_object_score(JokerObject *joker_object, Card* scored_card, enum Joker
...
         snprintf(score_buffer, sizeof(score_buffer), "X%d", joker_effect.xmult);
-        set_and_shift_text(score_buffer, &cursorPosX, TTE_RED_PB);
+        set_and_shift_text(score_buffer, &cursorPosX, &cursorPosY, TTE_RED_PB);
     }
     if (joker_effect.money > 0)
     {
         char score_buffer[INT_MAX_DIGITS + 2];
-        snprintf(score_buffer, sizeof(score_buffer), "+%d", joker_effect.money);
-        set_and_shift_text(score_buffer, &cursorPosX, TTE_YELLOW_PB);
+        snprintf(score_buffer, sizeof(score_buffer), "%d$", joker_effect.money);
```
ricfehr3, line 323: Not sure if removing the `+` was intentional or not. If it was, ignore me.

source/joker.c
MathisMartin31: Yeah I kept it there but now I don't know if it should stay. It does look weird though ^^'

source/joker.c
MathisMartin31: It was ^^
I looked at Balatro and it doesn't say "+X" when gaining money, but "X$"

source/joker.c
MathisMartin31: I changed it a bit, I think it's better what do you think?

MeirGavish: I don't quite understand this either. See [related comment](https://github.com/cellos51/balatro-gba/pull/95#discussion_r2280971597).

MathisMartin31: Oh I see.

The text doesn't seem to be perfectly centered, so what I think that they tried to do was shift the text by a tile to ensure it was approximatively displayed above the card, not too much to the left. 

It would be better, as you said un your other comment, to go the extra mile and actually compute the best tile to put the text on to center it above the cards/below the Joker

MathisMartin31: If we want to center the text begin written when scoring Cards/Jokers we would need to change slightly how we check for Joker effects and print messages, because we would need to know the length of the full message.

Right now we check if we have Chips, print the Chips, then go onto the Mult, xMult etc. This means that, for Walkie Talkie for example that gives Chips and Mult, we can't take the Mult into account when starting with printing the Chips.

We would need to create all the necessary buffers, compute the total length of all the messages combined, center it around the Card/Joker then snap it back int position if it goes too far to the right or left of the screen.

But maybe this could be something for another PR? This seems to be developing into a bigger issue to solve with scoring logic to change and all

MeirGavish: Definitely for another PR

source/game.c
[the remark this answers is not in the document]
MeirGavish: What the hell Copilot

source/joker.c
```diff
@@ -281,35 +282,49 @@ bool joker_object_score(JokerObject *joker_object, Card* scored_card, enum Joker
...
+        set_and_shift_text(score_buffer, &cursorPosX, &cursorPosY, TTE_YELLOW_PB);
     }
     // custom message for Jokers (including retriggers where Jokers will say "Again!")
     if (joker_effect.message[0] != '\0') // Message is not empty
```
```suggestion
    if (joker_effect.message != NULL && joker_effect.message[0] != '\0') // Message is not empty
```

I missed this in the retrigger PR, this needs a NULL-check

source/joker.c
```diff
@@ -281,35 +282,49 @@ bool joker_object_score(JokerObject *joker_object, Card* scored_card, enum Joker
...
+        cursorPosX += fx2int(joker_object->sprite_object->x);
+        cursorPosY = JOKER_SCORE_TEXT_Y;
+    }
+    else
+    {
+        // display the text on top of the card instead of below the Joker for Held Cards effects
+        // scored_card cannot be NULL here because of the joker event
+        cursorPosX += fx2int(card_object->sprite_object->x);
+        cursorPosY = HELD_CARD_SCORE_TEXT_Y;
+    }
```
```suggestion
    if (joker_event == JOKER_EVENT_ON_CARD_HELD)
    {
         // display the text on top of the card instead of below the Joker for Held Cards effects
        // scored_card cannot be NULL here because of the joker event
        cursorPosX += fx2int(card_object->sprite_object->x);
        cursorPosY = HELD_CARD_SCORE_TEXT_Y;
    }
    else
    {
        cursorPosX += fx2int(joker_object->sprite_object->x);
        cursorPosY = JOKER_SCORE_TEXT_Y;
    }
```

Just switch the cases since the double negative here makes it confusing to read

source/game.c
```diff
@@ -2156,11 +2157,27 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
+                    }
+
+                    joker_scored_index = 0;
+                    play_state = PLAY_SCORING_CARDS;
+                    
+                    if (card_object_is_selected(played[i]) && played_top - i >= *played_selections)
+                    {
+                        played_y -= int2fx(10);
+                    }
+                    break;
```
```suggestion
                    
                    if (card_object_is_selected(played[i]) && played_top - i >= *played_selections)
                    {
                        played_y -= int2fx(10);
                    }
                     
                    joker_scored_index = 0;
                    play_state = PLAY_SCORING_CARDS;
                    break;
```

Move the state transition to the end

include/joker.h
```diff
@@ -50,6 +50,8 @@ enum JokerEvent
     JOKER_EVENT_ON_HAND_PLAYED,     // Triggers only once when the hand is played
     JOKER_EVENT_ON_CARD_SCORED,     // Triggers when a played card scores (e.g. Walkie Talkie, Fibonnacci...)
     JOKER_EVENT_ON_CARD_SCORED_END, // Triggers after the card has finishd scoring (e.g. retrigger Jokers)
+    JOKER_EVENT_ON_CARD_HELD,       // Triggers when going through held cards
+    //JOKER_EVENT_ON_CARD_HELD_END,   // Triggers after when going through held cards (e.g. retrigger)
```
MeirGavish: So, this state is just for Mime? If so it can be removed for now rather than commented and re-added when Mime is.

include/joker.h
```diff
@@ -50,6 +50,8 @@ enum JokerEvent
     JOKER_EVENT_ON_HAND_PLAYED,     // Triggers only once when the hand is played
```
MeirGavish:
```suggestion
    JOKER_EVENT_ON_HAND_PLAYED,     // Triggers only once when the hand is played before the cards are scored
```

To make the timing of this event clearer.

source/joker.c
```diff
@@ -249,25 +250,25 @@ void joker_object_shake(JokerObject *joker_object, mm_word sound_id)
...
+void set_and_shift_text(char* str, int* cursor_pos_x, int* cursor_pos_y, int color_pb)
 {
-    tte_set_pos(*cursor_pos_x, JOKER_SCORE_TEXT_Y);
+    tte_set_pos(*cursor_pos_x, *cursor_pos_y);
     tte_set_special(color_pb * TTE_SPECIAL_PB_MULT_OFFSET);
     tte_write(str);
 
     // + 1 For space
-    const int joker_score_display_offset_px = (MAX_CARD_SCORE_STR_LEN + 1)*TTE_CHAR_SIZE;
+    const int joker_score_display_offset_px = (strlen(str) + 1)*TTE_CHAR_SIZE;
```
MeirGavish: The call to `strlen()` adds performance cost for each call, does it have to be dynamic? I thought the text is only actually shifted when displaying score, I think the only existing cases are Walkie Talkie and Scholar?

source/game.c
```diff
@@ -2156,11 +2157,27 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
                     }
                     break;
                 
+                case PLAY_BEFORE_SCORING:
+
+                    // Activate Jokers with an effect just before the hand is scored
+                    if (check_and_score_joker_for_event(&joker_scored_index, NULL, JOKER_EVENT_ON_HAND_PLAYED))
```
MeirGavish: It seems like you can now use this state/event (`PLAY_BEFORE_SCORING`/`JOKER_EVENT_ON_HAND_PLAYED`) to initialize the data for all the existing jokers that need it - moving the initialization logic from `on_joker_created()` as well as the reset logic from `JOKER_EVENT_ON_CARD_SCORED_END` to one place. I think then you can remove both `on_joker_created()` and the `JOKER_EVENT_ON_CARD_SCORED_END`. 

Actually after giving it some thought I think the joker callbacks I originally thought of can be replaced with joker events (such as JOKER_EVENT_ON_JOKER_CREATED which could be added when needed) and then everything would call the same function which would be a little awkward (especially because many events are not actually scoring but more notifying and there's always that card parameter) but also consistent.

source/game.c
```diff
@@ -2156,11 +2157,27 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
                     }
                     break;
                 
+                case PLAY_BEFORE_SCORING:
```
MeirGavish, line 2247: Also really waiting for #220 so we can get rid of all these levels of indentation and length of the switch cases.

source/game.c
```diff
@@ -2225,18 +2242,46 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
+                        scored_card_index = 0;
+
+                        play_state = PLAY_SCORING_JOKERS;
+                    }
+
+                    if (card_object_is_selected(played[i]))
+                    {
+                        played_y -= int2fx(10);
+                    }
+                    break;
```
MeirGavish, line 2368: Why is this needed on every state?

source/joker.c
```diff
@@ -281,35 +282,49 @@ bool joker_object_score(JokerObject *joker_object, Card* scored_card, enum Joker
     *retrigger = joker_effect.retrigger;
     // joker_effect.message will have been set if the Joker had anything custom to say
 
-    int cursorPosX = fx2int(joker_object->sprite_object->x) + 8; // Offset of 16 pixels to center the text on the card
+    int cursorPosX = 8; // Offset of one tile to better center the text on the card
```
```suggestion
    int cursorPosX = TTE_CHAR_SIZE; // Offset of one tile to better center the text on the card
```

Same comment from #218

source/joker.c
MathisMartin31: Yeah that one's my bad, I tried to anticipate too much when we talked about centering the effect text above the scoring card/joker, and didn't realize that the only case this was needed was with these two, and that the chips amount was always 2 digits long

source/game.c
MathisMartin31: I tested it, and it's done to maintain the cards where they should be (scoring ones above, non scoring ones below), and play animations correctly when the move up and down.

We need to update these animations not by returning, but by breaking to go to the code that updates the card position and scale.

I haven't looked into why it's done this way and why it works but if I touch any of these blocks the animations are borked

source/joker.c
MathisMartin31: Weird that it worked until now

source/game.c
MathisMartin31: I took out the `on_created` callback and added the `JOKER_EVENT_ON_JOKER_CREATED` event. I also took the opportunity to replace all NULL joker effect functions by a new `joker_effect_noop`. This way we don't need to check if `joker_effect` is NULL

MeirGavish:
Same here as in [the comment on #218](https://github.com/cellos51/balatro-gba/pull/218#issuecomment-3520378867). Since you added more commits after the merge here, maybe you can cherry-pick them from the backup branch to apply them after the reset if needed.

MathisMartin31:
The branch is fixed ^^'

source/game.c
MeirGavish: Good job!

PR #206, opened 2025-11-02
New Run Button

### LINKED ISSUE #185: New Run Button
Restarting functionality was wonderfully added in PR #155, now pressing any key on game over restarts the run.
The only thing that's left now is to make it clear to the user by adding a "New Run" button at the game over menu. We can just add the single button and leave it highlighted, and instead of using any button to restart, it can be done with the "A" selection button to feel like you selected the highlighted button.
In the future we can decide to add a second "Main Menu" button and make this functionality more complex but for now we can keep it simple.

The game over menu is created by inflating an empty menu rectangle in place of the round end menu with `main_bg_se_copy_expand_3x3_rect()`, adding the button would mean keeping it in the screenblock outside the visible screen and copying it to the menu after it's created so it can pop out along with it. The menu may need to be enlarged a bit to allow space for the button, make sure the menu is still well centered and the text is still well centered within it.
Messing with the background graphics could potentially change the palette mapping done by grit and mess up hardcoded palette indices for button highlights etc. but I hope this doesn't happen if no new colors are added.

MeirGavish:
closes #185 

I ended up opting for red for both since it was much simpler with the palettes.

Quoting from the issue
>Messing with the background graphics could potentially change the palette mapping done by grit and mess up hardcoded palette indices for button highlights etc. but I hope this doesn't happen if no new colors are added.

This did happen, and I needed to change some palette indices to compensate.

@drewjohnson2 Also want to thank you since your refactors made this much simpler to implement.

game_over_new_run_large
you_win_new_run_large

Edit: this also included a refactor to replace the multiple copy calls for the cash out button with a single call to `main_bg_se_copy_expand_3x3_rect()` since I changed that rect too so I thought I might as well.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -1472,7 +1472,6 @@ static void game_playing_process_hand_select_input()
         }
         else // Discard button logic
         {
-			// 7 is score and play hand button color
             memcpy16(&pal_bg_mem[PLAY_HAND_BTN_BORDER_PID], &pal_bg_mem[PLAY_HAND_BTN_PID], 1);
             memset16(&pal_bg_mem[DISCARD_BTN_BORDER_PID], HIGHLIGHT_COLOR, 1);
 


MeirGavish:
I'm also wondering whether I should make the dialog a bit larger like this

new_run_larger_menu_large

Edit: Yes, I think so, gonna change it.

source/game.c
```diff
@@ -1468,8 +1473,8 @@ static void game_playing_process_hand_select_input()
         else // Discard button logic
         {
 			// 7 is score and play hand button color
```
ricfehr3: Comment not needed, and is also incorrect.

source/game.c
```diff
@@ -2510,30 +2515,12 @@ static void game_round_end_display_cashout()
...
-        main_bg_se_fill_rect_with_se(main_bg_se_get_se(top_point), top_rect);
-    
-        Rect middle_rect = {11, 9, 22, 9};
-        BG_POINT middle_point = {6, 30};
-        main_bg_se_fill_rect_with_se(main_bg_se_get_se(middle_point), middle_rect);
-    
-        Rect bottom_rect = {11, 10, 22, 10};
-        BG_POINT bottom_point = {6, 31};
-        main_bg_se_fill_rect_with_se(main_bg_se_get_se(bottom_point), bottom_rect);
+        main_bg_se_copy_expand_3x3_rect(CASHOUT_DEST_RECT, CASHOUT_SRC_3X3_RECT_POS);
```
ricfehr3, line 2517: Very big fan of this to clean it up. nice!

source/game.c
```suggestion
```

MeirGavish: Nice catch! Thanks!

source/game.c
MeirGavish: Thanks! Yeah, I've been wanting to do that ever since I added this function, this was a good opportunity.

MeirGavish:
@ricfehr3 Thanks a lot! Merging

PR #207, opened 2025-11-02
Update Hanging Chad art

MeirGavish:
I tried to make some improvements to the art for Hanging Chad.
I figured that since we don't have enough room to spell out "JOKER" anyway we could just reduce to the ballot card to 4 options instead of 5 to give the text some room. I also wanted to change the chad so it feels more hanging.

Tagging artists for their opinions: @MathisMartin31 @GreenNinja2525 @NoWhammies10

balatro-gba-4

In-game

hanging_chad_edited_large

Enlarged

And adding the .ase file (zipped) where I separated the chad to another layer (I almost feel like an actual artist doing that)
[hanging_chad_edited.zip](https://github.com/user-attachments/files/23292005/hanging_chad_edited.zip)


MathisMartin31:
Nice ^^, I was thinking about taking a look at this one the last few days lol

I think putting the "Ballot" one pixel up could allow you to make bigger rectangles to fit the original better? That one is up to you though, it's already a big improvement with the hanging thingy

MeirGavish:
> Nice ^^, I was thinking about taking a look at this one the last few days lol
> 
> I think putting the "Ballot" one pixel up could allow you to make bigger rectangles to fit the original better? That one is up to you though, it's already a big improvement with the hanging thingy

Thanks! Good point. Fitting letters in there would be problematic since they would need to be 3x3 at least but maybe it would still look better with bigger rectangles, I will give it a try later.
Yeah, seeing it in-game made me want to try improving it, feel free to take it on and make changes to what I did.

MathisMartin31:
I couldn't get the letters to fit, but I made the rectangles bigger and added a little shadow. What do you think? (I put Hack next to it because I tweaked a color or two to make Hanging Chad color a bit lighter)

hanging_chad

MeirGavish:
@MathisMartin31 Good job! Looks great! The only thing is the dots on the left and right, now that the rectangles are bigger, they should be moved to align with them. I think I prefer to merge this PR for now and can you open a new PR for your edits so you get full credit for the contribution?

MathisMartin31:
Oh I wouldn't have minded don't worry, but I'll create a new PR ^^

ricfehr3:
Just wanted to say it looks awesome :)

MeirGavish:
Next PR #209

MathisMartin31:
> Just wanted to say it looks awesome :)

Thanks ^^

PR #210, opened 2025-11-03
Add flaming effect when exceeding score requirements

MathisMartin31:
Here's a quick video of it working, plus some gameplay to confirm the color swapping is correct

https://github.com/user-attachments/assets/8ecc6a5f-a0b9-4c53-9b3e-7737664ad9d6


source/game.c
```diff
@@ -2351,6 +2374,61 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
+
+static void process_flaming_score()
+{
+    static u8 flame_score_chips_frame = 0;
+    static u8 flame_score_mult_frame  = 4;
+    
+    // animate flames at 10FPS = change sprites every 6 frames
+    if (score_flames_active)
+    {
+        if (timer % 5 == 0)
```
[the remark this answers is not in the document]
MathisMartin31: My bad, I increased the animation rate to 12 fps but forgot to update the comment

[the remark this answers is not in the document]
MeirGavish: Feels like a magic number too...

MeirGavish:
That was quick! Nice! Will the flames also display when selecting cards in hand? An easy way to test this is to disable deck shuffling and then you can get a Straight Flush on the first round with a score of 100x8.

source/game.c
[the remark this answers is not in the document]
MathisMartin31: Yeah, maybe I can define them at the top of the file?

MathisMartin31:
> Will the flames also display when selecting cards in hand?

That's a good question! Normally yes, since I toggle the flames from the `display_chips/mult` functions but I'll check

@MeirGavish I just tested it, it all works as intended

source/game.c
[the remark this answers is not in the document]
MeirGavish: It can be above the function too

source/game.c
```diff
@@ -2372,6 +2451,9 @@ static void game_playing_ui_text_update()
         last_hand_size = hand_get_size();
         last_deck_size = deck_get_size();
     }
+
+    // animate score flames if we exceed the score requirement
+    process_flaming_score();
```
MeirGavish: This isn't really a part of the UI text, move this call to game_playing_on_update()

source/game.c
```diff
@@ -2351,6 +2400,36 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
+            // mult flame (red)
+            main_bg_se_copy_rect(score_flame_mult_frames [flame_score_mult_frame  + 1], score_flame_mult);
+        }
+    }
+    else
+    {
+		// clear flames if the condition is not met
+        main_bg_se_copy_rect(score_flame_chips_frames[0], score_flame_chips);
+        main_bg_se_copy_rect(score_flame_mult_frames [0], score_flame_mult);
+    }
```
MeirGavish: This is a redundant copy called on every frame, move it to `check_flaming_score()` then you can use the toggle

source/game.c
```diff
@@ -2351,6 +2400,36 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
     }
 }
 
+// update flames animation every 5 frames for 12FPS
+#define SCORE_FLAMES_ANIM_FREQ 5
+
+static void process_flaming_score()
```
MeirGavish: This function name should be prefixed with `game_playing_` so it's clear it belongs to that state's on_update calls and consistent with most other functions from there.

source/game.c
```diff
@@ -2351,6 +2400,36 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
+{
+    static u8 flame_score_chips_frame = 0;
+    static u8 flame_score_mult_frame  = NB_SCORE_FLAMES_FRAMES / 2; // offset mult flames animation by half the number of frames for them to not play in sync
+    
+    if (score_flames_active)
+    {
+		// animate flames at 12FPS = change sprites every 5 frames
+        if (timer % SCORE_FLAMES_ANIM_FREQ == 0)
+        {
+            // cycling through 8 frames total
```
```suggestion
```
Redundant comment, the number of frames is already in the define

source/game.c
```diff
@@ -2351,6 +2400,36 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
+#define SCORE_FLAMES_ANIM_FREQ 5
+
+static void process_flaming_score()
+{
+    static u8 flame_score_chips_frame = 0;
+    static u8 flame_score_mult_frame  = NB_SCORE_FLAMES_FRAMES / 2; // offset mult flames animation by half the number of frames for them to not play in sync
+    
+    if (score_flames_active)
+    {
+		// animate flames at 12FPS = change sprites every 5 frames
```
```suggestion
```

Redundant comment, this is already stated in the define

source/game.c
```diff
@@ -483,6 +484,33 @@ static const Rect GAME_OVER_ANIM_RECT       = {11,      8,       23,     28};
...
+                                              {29,      20,      31,     20},
+                                              {29,      21,      31,     21},
+                                              {29,      22,      31,     22},
+                                              {29,      23,      31,     23},
+                                              {29,      24,      31,     24},
+                                              {29,      25,      31,     25},
+                                              {29,      26,      31,     26},
+                                              {29,      27,      31,     27},
+                                              {29,      28,      31,     28},
+};
```
MeirGavish: There's no point in keeping so many consecutive rects in memory, would be better to just define a single rect for each of the chips and mult and use the current frame index to just offset the top and bottom, it will save up some memory.
Also the 0th index should be the first frame of the animation and the no animation should be in a separate rect, it avoids the unnecessary +1's and would make it clear what it's meant for in the rect name.

Also use uppercase for the constant names.

See my [suggestion](https://github.com/cellos51/balatro-gba/pull/210#discussion_r2488737881)

source/game.c
```diff
@@ -2351,6 +2400,36 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
+        if (timer % SCORE_FLAMES_ANIM_FREQ == 0)
+        {
+            // cycling through 8 frames total
+            flame_score_chips_frame = (flame_score_chips_frame + 1) % NB_SCORE_FLAMES_FRAMES;
+            flame_score_mult_frame  = (flame_score_mult_frame  + 1) % NB_SCORE_FLAMES_FRAMES;
+            // chips flame (blue)
+            main_bg_se_copy_rect(score_flame_chips_frames[flame_score_chips_frame + 1], score_flame_chips);
+            // mult flame (red)
+            main_bg_se_copy_rect(score_flame_mult_frames [flame_score_mult_frame  + 1], score_flame_mult);
+        }
```
```suggestion
            Rect curr_frame_chips_rect = SCORE_FLAMES_CHIPS_FRAME_RECT;
            flame_score_chips_frame_idx = (flame_score_chips_frame + 1) % NB_SCORE_FLAMES_FRAMES;
            flame_score_mult_frame_idx  = (flame_score_mult_frame  + 1) % NB_SCORE_FLAMES_FRAMES;
            
            curr_frame_chips_rect.top += flame_score_chips_frame_idx;
            curr_frame_chips_rect.bottom += flame_score_chips_frame_idx;
            
            // chips flame (blue)
            main_bg_se_copy_rect(curr_frame_chips_rect, score_flame_chips);
            // mult flame (red)
            main_bg_se_copy_rect(score_flame_mult_frames [flame_score_mult_frame  + 1], score_flame_mult);
        }
```
Regarding [this comment](https://github.com/cellos51/balatro-gba/pull/210#discussion_r2488698703)
This suggestion is not a full rewrite, more of a sketch of how it would look.

source/game.c
```diff
@@ -483,6 +484,33 @@ static const Rect GAME_OVER_ANIM_RECT       = {11,      8,       23,     28};
...
+                                              {29,      22,      31,     22},
+                                              {29,      23,      31,     23},
+                                              {29,      24,      31,     24},
+                                              {29,      25,      31,     25},
+                                              {29,      26,      31,     26},
+                                              {29,      27,      31,     27},
+                                              {29,      28,      31,     28},
+};
+static const BG_POINT score_flame_chips     = {1,       9};
+static const BG_POINT score_flame_mult      = {5,       9};
```
```suggestion
static const BG_POINT SCORE_FLAMES_CHIPS_POS     = {1,       9};
static const BG_POINT SCORE_FLAMES_MULT_POS      = {5,       9};
```

Add "_POS" to the name to make the name specific and not ambiguous and use uppercase for the constant to be consistent with other rects and positions.

source/game.c
```diff
@@ -483,6 +484,33 @@ static const Rect GAME_OVER_ANIM_RECT       = {11,      8,       23,     28};
 static const BG_POINT NEW_RUN_BTN_DEST_POS  = {15,      26};
 static const Rect NEW_RUN_BTN_SRC_RECT      = {0,       30,      4,      31};
 
+// Flaming score animation frames
+#define NB_SCORE_FLAMES_FRAMES 8
```
```suggestion
#define NUM_SCORE_FLAMES_FRAMES 8
```

Just a more consistent name to be clearer.

MeirGavish (changes requested):
Good work! It looks great and really adds a lot.
Just a few changes, see my comments.

source/game.c
MathisMartin31: I did basically this, but reused the frame rect variable by shifting it by 3 tiles to the right to fetch the red flames. Also shifted the frames themselves by 4 so I can use the same frame index for both animations. This all condenses the needed values to 2 Rects and 2 Points

source/game.c
MeirGavish, line 1: Should have told you to wait before my PR is merged, sorry.

MathisMartin31: Sorry about what ? 🤔

MathisMartin31:
> I just want to note that I see that in Balatro the flames persist for a bit until the score is fully accumulated at least

I had actually thought about it before deciding against trying it, as I think this would have added a whole lot more tiles to animate the fade in/out of the flames

MeirGavish:
> > I just want to note that I see that in Balatro the flames persist for a bit until the score is fully accumulated at least
> 
> I had actually thought about it before deciding against trying it, as I think this would have added a whole lot more tiles to animate the fade in/out of the flames

Of course, I didn't mean a full fade-out, just to let the flames persist for a little longer.

MathisMartin31:
Oh I see what you mean now ><

Issue #200, opened 2025-11-01
[BUG] If you get a high enough score it overflows and the text persists


MeirGavish:
**Describe the bug**
If you get a high enough score it overflows its rect and the overflown digit text is not erased when the text is reset.
This would happen in several text boxes 
* The current hand total score - 7 character tiles
* The current total score of the player for the round - 5 character tiles
* Both the chips and mult boxes - 3 character tiles each
* Money - 7 character tiles. 
This doesn't seem to be possible there currently but it also needs to be addressed. This also needs to include the $ sign.

**To Reproduce**
Get a high enough score so the digits overflow the text rect.

**Expected behavior**
Text should be erased

**Screenshots**

Screenshot from discussion #148
<img width="3055" height="2026" alt="Image" src="https://github.com/user-attachments/assets/9413d782-d96d-4b2e-a8af-0fd8a5d02469" />

**What version**
Currently this is still happening in the main branch...

**What Emulator or Platform were you playing on**
 - mgba 

**Additional context**
So, I think the correct solution here is actually constrain all the score text to its rects and use K/M when it's too big to display, this is currently not done very well, with some duplicated code, it should be in a general function that receives the max number of characters

MathisMartin31:
Since we're dealing with 32 bits, going up to Billions abbreviated with B would be enough, with a `#define NANEINF 0xFFFFFFFF` and a check when calculating the score.

Or we could implement a way to display numbers in scientific notation as it's done in Balatro, with 16500 being displayed as 1.65e4. This could also help solve issue #177

MathisMartin31:
So just to be sure, the position of text isn't tied to the tiles, it can be any point on the screen right, and the characters are 8x8 tiles with the characters fully contained within a 7x7 square?

Because if yes, we could do something like this where the gray squares represent the critical area actually taken up by the characters:

- from 0 to 99,999, just print the score directly like this, centered
<img width="1688" height="387" alt="Image" src="https://github.com/user-attachments/assets/a678b679-0564-4ef3-8f5c-4c7f0bc4520f" />

- from 10^6 and up, print text like this, broken up between integer part, decimal part and exponent
<img width="2436" height="548" alt="Image" src="https://github.com/user-attachments/assets/aac61fd2-3350-4267-8358-b45dad30f810" />

[...]

PR #212, opened 2025-11-05
Make Chips, Mult and Score unsigned and protect against overflow

MathisMartin31:
The goal of this PR is to allow for bigger scores by making the values unsigned instead of signed.

I also added macros in `utils.h` to make additions/multiplications safe in regards to integer overflow by adding quick checks for whether or not the result of said operation would be larger than `UINT32_MAX`. If yes, return `UINT32_MAX`, else the operation result.

The above check is heavier for multiplications as it requires an integer division, but it is currently required in only 2 places: when a Joker scores xMult, and when computing Chips*Mult at the end of each Hand, so I believe it's acceptable to run this check.

Let me know if I missed places where overflow safety would be needed. ^^

I had mentioned doing this in the discussion for this issue, but this request _**does not**_ solve #200, although I believe it's a first step towards making the scoring system better and more reliable.

### DIFF
--- a/source/util.c
+++ b/source/util.c
@@ -17,25 +17,23 @@ int int_arr_max(int int_arr[], int size)
 
 // Avoid uint overflow when add/multiplying score
 
-#define UINT32_MAX 4294967295
-#define UINT16_MAX 65535
-
-unsigned long u32_protected_add(unsigned long a, unsigned long b)
+uint32_t u32_protected_add(uint32_t a, uint32_t b)
 {
     return (a > (UINT32_MAX - b)) ? UINT32_MAX : (a + b);
 }
     
-unsigned int u16_protected_add(unsigned int a, unsigned int b)
+uint16_t u16_protected_add(uint16_t a, uint16_t b)
 {
     return (a > (UINT16_MAX - b)) ? UINT16_MAX : (a + b);
 }
 
-unsigned long u32_protected_mult(unsigned long a, unsigned long b)
+uint32_t u32_protected_mult(uint32_t a, uint32_t b)
 {
     return (a == 0 || b == 0) ? 0 : (a > (UINT32_MAX / b) ? UINT32_MAX : a * b);
 }
 
-unsigned int u16_protected_mult(unsigned int a, unsigned int b)
+uint16_t u16_protected_mult(uint16_t a, uint16_t b)
 {
     return (a == 0 || b == 0) ? 0 : (a > (UINT16_MAX / b) ? UINT16_MAX : a * b);
+
 }


include/util.h
```diff
@@ -33,11 +33,40 @@ static inline int get_digits_even(int n)
...
+        )                           \
+)
+
 #define UNDEFINED -1
 
 #define NUM_ELEM_IN_ARR(arr) (sizeof(arr) / sizeof((arr)[0]))
 
-#define INT_MAX_DIGITS 11 // strlen(str(INT_MAX)) = strlen("-2147483647")
+#define INT_MAX_DIGITS  11 // strlen(str(INT_MAX)) = strlen("-2147483647")
+#define UINT_MAX_DIGITS 10 // strlen(str(UINT32_MAX)) = strlen("4294967295")
```
ricfehr3, line 43: I don't think this would work in `sprintf`. I'm one to talk after being completely wrong the last time we talked about it, haha. But, if you recall, we were testing why `sprintf` had errors on the maximum digits, and we found that we needed `11` for an integer. In testing that, I tried to change the `sprintf` stuff to use unsigned integers as you do here and it thought the unsigned int should still have `11`, producing that warning. 

You're not technically wrong here, but I think we should have it at 11 as well because it's how the compiler sees it.

MathisMartin31: I remember testing the code by forcing a pair to give me UINT32_MAX Chips and it didn't crash so I think it's okay to have it at 10?

source/joker.c
```diff
@@ -274,30 +274,32 @@ bool joker_object_score(JokerObject *joker_object, Card* scored_card, enum Joker
         return false;
     }
 
-    *chips    += joker_effect.chips;
-    *mult     += joker_effect.mult;
-    *mult     *= joker_effect.xmult > 0 ? joker_effect.xmult : 1; // if xmult is zero, DO NOT multiply by it
+    // protect chips and mult against overflow
+    *chips = U32_PROTECTED_ADD (joker_effect.chips, *chips);
```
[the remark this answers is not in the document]
MathisMartin31: I actually liked that it was all aligned though... Maybe I'm going overboard with it sometimes but I like when similar lines are neatly aligned ><

include/util.h
ricfehr3: I guess I just really wanted to triple down on being wrong about this 😂 . This is good, what I was seeing was this:

```c
// works
snprintf(score_buffer, sizeof(score_buffer), "+%lu", (u32)card_value);
// doesn't work
snprintf(score_buffer, sizeof(score_buffer), "+%lu", card_value);
```
It's because the variable it was formatting was an `int`, the destination of `%lu` doesn't matter.

My bad again. Next time we change it I'll be ready with another wrong take o7

source/game.c
```diff
@@ -2246,16 +2248,20 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
                                 tte_set_pos(fx2int(played[j]->sprite_object->x) + 8, SCORED_CARD_TEXT_Y); // Offset of 16 pixels to center the text on the card
                                 tte_set_special(TTE_BLUE_PB * TTE_SPECIAL_PB_MULT_OFFSET); // Set text color to blue from background memory
 
+                                u8 card_value = card_get_value(played[j]->card);
+
                                 // Write the score to a character buffer variable
-                                char score_buffer[INT_MAX_DIGITS + 2]; // for '+' and null terminator
-                                snprintf(score_buffer, sizeof(score_buffer), "+%d", card_get_value(played[j]->card));
+                                char score_buffer[UINT_MAX_DIGITS + 2]; // for '+' and null terminator
+                                snprintf(score_buffer, sizeof(score_buffer), "+%d", card_value);
```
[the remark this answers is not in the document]
MathisMartin31: I actually added a `UINT8_MAX_DIGITS` for this since the card value is a `u8` and not a `u32`, and the specifier for uint8 is `%hhu`

include/util.h
MathisMartin31: Well I just changed it again ^^'

Copilot's review made me realize some things were wrong about it. I went the extra mile and made everything match the fact that the card value is an `u8` by adding the constant `UINT8_MAX_DIGITS` and using `%hhu` in the snprintf

I tested it and it should work correctly.

ricfehr3: This is cool. I didn't know about that format specifier. From the description of in the c99 standard, I sounds like it's exactly what to put here, an integer form of an unsigned char. I would have used a `%u` and never thought about it. Nice one.

source/game.c
```diff
@@ -2307,15 +2308,17 @@ static void played_cards_update_loop(bool* discarded_card, bool* sound_played)
                                 tte_set_pos(fx2int(scored_card_object->sprite_object->x) + TILE_SIZE, SCORED_CARD_TEXT_Y); // Offset of 1 tile to keep the text on the card
                                 tte_set_special(TTE_BLUE_PB * TTE_SPECIAL_PB_MULT_OFFSET); // Set text color to blue from background memory
 
+                                u8 card_value = card_get_value(scored_card_object->card);
+
                                 // Write the score to a character buffer variable
-                                char score_buffer[INT_MAX_DIGITS + 2]; // for '+' and null terminator
-                                snprintf(score_buffer, sizeof(score_buffer), "+%d", card_get_value(scored_card_object->card));
+                                char score_buffer[UINT8_MAX_DIGITS + 2]; // for '+' and null terminator
+                                snprintf(score_buffer, sizeof(score_buffer), "+%hhu", card_value);
```
ricfehr3: I did have to look this one up. It's neat. Seems like a lot of online guides miss it too, I had to check the actual standard, haha.

ricfehr3 (approved):

include/util.h
```diff
@@ -45,9 +43,9 @@ static inline int get_digits_even(int n)
...
 int int_arr_max(int int_arr[], int size);
 
-u32 u32_protected_add(u32 a, u32 b);
-u16 u16_protected_add(u16 a, u16 b);
-u32 u32_protected_mult(u32 a, u32 b);
-u16 u16_protected_mult(u16 a, u16 b);
+unsigned long u32_protected_add (unsigned long a, unsigned long b);
+unsigned int  u16_protected_add (unsigned int  a, unsigned int  b);
+unsigned long u32_protected_mult(unsigned long a, unsigned long b);
+unsigned int  u16_protected_mult(unsigned int  a, unsigned int  b);
```
MeirGavish: include stdint.h and use uint32_t and uint16_t instead of unsigned long and unsigned int

source/util.c
```diff
@@ -16,22 +16,26 @@ int int_arr_max(int int_arr[], int size)
 }
 
 // Avoid uint overflow when add/multiplying score
-u32 u32_protected_add(u32 a, u32 b)
+
+#define UINT32_MAX 4294967295
+#define UINT16_MAX 65535
```
MeirGavish: Also use the constants from stdint.h

MeirGavish (changes requested):
I just couldn't help but notice the changes in the last commit so I'm adding my comments already.

include/util.h
MathisMartin31: Should be fixed now ^^

MeirGavish (approved):
Great!

Issue #191, opened 2025-10-30
Improve efficiency of "whole hand" condition checks


MathisMartin31:
As @ricfehr3 suggested in [#160](https://github.com/cellos51/balatro-gba/pull/160#discussion_r2476265327) , I'm creating an issue to remember to optimize how we check for "whole hand" conditions at some point.

For instance, The Duo, Trio, Family, Tribe and Order check for their condition (there being a Pair, 3 OAK, 4 OAK, Flush or Straight in the played hand) every time they are scored which, with Joker retriggers, could happen several times per hand played while the hand itself doesn't change.

This is inefficient, and we should find a way to check these conditions only once just before scoring the hand and store this information, instead of checking it when scoring the Jokers.

An idea I had was changing the way the `HandType/hand_type` enum/variable and work: when calling the `hand_get_type()` function, instead of attributing the highest value hand to `hand_type`, the variable would act as a bitfield of size 13 (10 regular + 3 secret hands) and store all hands contained within the played hands, which would be flags represented by the values of the `HandType` enum.

Getting the hand that actually counts would mean retrieving the first leftmost bit set to 1, and checking if the played hand contains a specific hand would mean checking if `(hand_type >> HAND_TYPE_FLAG) && 1` is true

Edit: that or just use a proper bitfield doing this, whichever one looks cleaner:

```
struct {
    u16 high_card : 1;
    u16 pair : 1;
    ...
} hand_type;
```

What do you think?

MeirGavish:
The final scored hand type should stay in a single variable using the existing `enum HandType`, it should be computed once per played hand including the hand type prioritizing. That said we can have a global `contained_hands` struct that caches the contained hands while they are computed only once, the fields can be bools in my opinion but we can save on memory by using bitfields so long as it has a simple interface. The `hand_contains_*` functions can access this table and we can move their computation logic to `compute_hand_contains_*` functions.

Also this is another sensitive place to refactor, we don't want to break hand detection.

Finally, PR #95 touches that and further complicates the hand type detection and I mean to integrate it soon so keep that in mind.

MathisMartin31:
Indeed, we should wait for this PR to be merged and see what can be done afterwards so we aren't both changing the same code at the same time ^^'

Issue #174, opened 2025-10-19
Add "Hand" Structure and move into it's own file


ricfehr3:
Currently, we use an array of pointers to **CardObject**'s to represent a hand. This is a static variable held in `game.c`. It would be useful to move that **CardObject** array into it's own **Hand** structure that also contains variables relating to the hand, see: https://github.com/cellos51/balatro-gba/issues/156#issuecomment-3350479772 , thanks @MathisMartin31  :) .

All functions relating to modifying and analyzing the hand should be moved into their own implementation files. We currently have a `hand_analysis.h/c` file, but that may need to be combined with the concept of a `Hand` into one file, up to whoever implements it. 

These functions include all the ones located in `game.c`:

```c
static const HandValues hand_base_values[];
int get_hand_top(void)
int hand_get_size(void)
void sort_hand_by_suit()
void sort_hand_by_rank()
void sort_cards()
enum HandType hand_get_type()
static void print_hand_type(const char* hand_type_str)
void set_hand()
void hand_set_focus(int index)
void hand_deselect_all_cards()
void hand_change_sort()
int hand_get_max_size()
bool hand_discard()
bool hand_play()
....
Whatever else you find...
```

Each of the functions should either be moved to the new hand file and sit behind a `static` scope within the file, or if it's a public facing function, it needs to take in a pointer to the "**Hand**" struct it intends to modify.

This does beg the question of how a "hand" should be designed. Should it be generic to encompass the a deck, joker row, selection cards? With variables for things like face-up and face-down per card? I think so. But moving the functionality to it's own file is a good first step imo.

MathisMartin31:
@ricfehr3 Since this is heavily used by the Round state, I think we should revive this issue and resolve it before addressing #465

MathisMartin31:
Also yes, I think hand.h and hand_analysis.h should be merged into a single file

MeirGavish:
I think separating the hand to its own entity is a good idea, allowing a hand to exist independently of the round would be good. That way we can create a hand for when opening a Tarot/Spectral card or create a hand separately of the deck and round to unit test hand type detection and jokers for example.

>This does beg the question of how a "hand" should be designed. Should it be generic to encompass the a deck, joker row, selection cards? With variables for things like face-up and face-down per card? I think so. But moving the functionality to it's own file is a good first step imo.

[...]

PR #213, opened 2025-11-06
Optimize "whole hand" conditions checks for Joker effects

MathisMartin31:
@ricfehr3 @MeirGavish
From the discussion in issue #191

I don't really have a way of measuring if we really save cycles with this solution but here are my main takeaways:

- Since the goal was to store all hand types contained within the played hand, the `hand_get_type` function now doesn't return early which means it will always check for a straight, flush, etc. This could be bad, but it was doing that most of the time already since we are at High Card almost every time we select a card, until suddenly we make a hand, and it is most likely not a 5 of a Kind that would have returned early.

- In return, thee checks will be done only once each time this function is called, because their result is immediately stored within the `contained_hands` bitfield.

- These checks will not be called from Joker Effect functions anymore, because they now only need to access the bitfield and check the bit corresponding to the hand they give a bonus to. This affects the rare Duo, Trio, Family, Order and Tribe Jokers, but also more common ones like the Zany Joker.

All in all I'm not sure we're gaining much in the end. Maybe the `hand_get_type` function can be further optimized by factorizing some loops from the functions in `hand_analysis.c`, but we should be losing cycles only when making big, rare hands

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -1359,7 +1359,8 @@ enum HandType compute_hand_type(struct ContainedHandTypes contained_types)
         // Shift the bit we want to check to the front and mask it with 1 to keep only that
         // Since the ContainedHandTypes is ordered the same way as the HandType enum, we
         // can shift right by ret-1 to have the bit we want at the front
-        if ((contained_types.value >> (ret-1)) & 0x1) {
+        if ((contained_types.value >> (ret - 1)) & 0x1)
+        {
             break;
         }
     }


include/game.h
```diff
@@ -78,6 +78,25 @@ enum HandType
     FLUSH_FIVE
 };
 
+// Store all contained hands to optimize "whole hand condition" Jokers
+typedef struct ContainedHandTypes
```
ricfehr3, line 100: 2 things here.

1. I have no problem with adding `PADDING : 3` but just to throw it out there as an option, you can also use the gcc attribute: `__attribute__ ((aligned (4)))` when declaring the struct: https://gcc.gnu.org/onlinedocs/gcc-3.3/gcc/Type-Attributes.html
2. I think you should us tonc types here: `u16` for consistency.

source/game.c
```diff
@@ -169,6 +169,7 @@ static enum HandState hand_state = HAND_DRAW;
 static enum PlayState play_state = PLAY_PLAYING;
 
 static enum HandType hand_type = NONE;
+static ContainedHandTypes contained_hands = {0};
```
ricfehr3: nitpick: Prefix this with an `_`.

include/game.h
MathisMartin31: I thought using padding was a bit more obvious than gcc attributes, but maybe not. 🤔 

tonc types were not available in `game.h` since tonclib is included in `game.c` and I didn't know if it was a good idea to move this include

ricfehr3: The `PADDING` is good, no worries. Another (cleaner imho) option would be to do an unnamed bitfield:

```c
{
    unsigned short int HIGH_CARD       : 1;
    unsigned short int PAIR            : 1;
    unsigned short int TWO_PAIR        : 1;
    unsigned short int THREE_OF_A_KIND : 1;
    unsigned short int FOUR_OF_A_KIND  : 1;
    unsigned short int STRAIGHT        : 1;
    unsigned short int FLUSH           : 1;
    unsigned short int FULL_HOUSE      : 1;
    unsigned short int STRAIGHT_FLUSH  : 1;
    unsigned short int ROYAL_FLUSH     : 1;
    unsigned short int FIVE_OF_A_KIND  : 1;
    unsigned short int FLUSH_HOUSE     : 1;
    unsigned short int FLUSH_FIVE      : 1;
    unsigned short int                 : 3;
    // this works too, the '0' auto-pads it to the last variables boundary
    // unsigned short int                 : 0;
} ContainedHandTypes;
```

I say it's cleaner because you can't even access the data in `PADDING`, but up to you.

MathisMartin31: I'll do that, thanks ^^

MeirGavish: >tonc types were not available in game.h since tonclib is included in game.c and I didn't know if it was a good idea to move this include

@MathisMartin31 
You can include tonc.h in game.h.

MathisMartin31:
Branch rebased and tested, ready for review ^^

MeirGavish:
Sorry for the delay with this PR, since this makes changes in the hand detection code I want to merge this after making the release mentioned in #307.

MathisMartin31:
No no don't worry, it's a very small optimization so it's only natural this comes after features and functional stuff ^^

MathisMartin31:
I finally found that stupid whitespace, AAAAAAAAH (not on my PC so I couldn't do clang format automatically T_T)

ricfehr3:
> I finally found that stupid whitespace, AAAAAAAAH (not on my PC so I couldn't do clang format automatically T_T)

I am so... so.. sorry. That sounds like an absolute nightmare :astonished: .

MathisMartin31:
> I am so... so.. sorry. That sounds like an absolute nightmare 😲 .

I'm exaggerating because I'm tired from work, don't worry it's not that bad ^^'

MeirGavish:
> > I am so... so.. sorry. That sounds like an absolute nightmare 😲 .
> 
> I'm exaggerating because I'm tired from work, don't worry it's not that bad ^^'

@ricfehr3 lol Honestly, maybe it's worth considering switching to something a little more forgiving that will make it a little easier on developers. I haven't checked but maybe `clang-tidy` can give us similar format checking that would not be so picky about whitespace etc.

ricfehr3:
> > > I am so... so.. sorry. That sounds like an absolute nightmare 😲 .
> > 
> > 
> > I'm exaggerating because I'm tired from work, don't worry it's not that bad ^^'
> 
> @ricfehr3 lol Honestly, maybe it's worth considering switching to something a little more forgiving that will make it a little easier on developers. I haven't checked but maybe `clang-tidy` can give us similar format checking that would not be so picky about whitespace etc.

I don't think we should get rid of `clang-format`. Besides looking nice, It's a pretty standard tool and It helps with PR reviews as it's both consistent and points formatting issues we would otherwise have to find and comment on. However, it shouldn't be this much of a pain in the ass, lol. `clang-format` as a standalone executable should be sufficient to not have to manually find and fix these issues:

```sh
clang-format -i source/blind.c
```

@MathisMartin31 I think something may be configured incorrectly in your IDE. I remember you mentioned running VSCodium. I downloaded it, installed the [extension](https://marketplace.visualstudio.com/items?itemName=xaver.clang-format) (recommended [here](https://clang.llvm.org/docs/ClangFormat.html#visual-studio-code-integration)), then hit `ctrl-shift-I` (I got it from [here](https://code.visualstudio.com/docs/cpp/cpp-ide#_code-formatting)) and it was able to format the file in-place. I thought this was an automatic check in vscode, but I should have tried and verified it earlier, that's my bad.  

As for `clang-tidy`, I did mess with it earlier, and I don't think we should use it. It's not a code formatter, but a linter and static analyzer. Which, would be really cool, but I think that would be a much bigger task with a lot of future upkeep (I think `clang-format` and -Wall is sufficient). It also wouldn't work as a formatter like with `clang-format.`

All this said, I don't want to make this project difficult to submit PRs to. I  feel like I should do a few things to make it easier to approach this. 
1. Create a clear developer guide (I know I said I would stop writing docs but I guess I'm addicted) that lays out:
	1. Our PR requirements, and our CI checks.
	2. Our Coding Style/Standards
	3. How to build and open docs
	4. How to use tools
		1. clang-format
			1. vscode
			2. manually
		2. Scripts in the `.scripts` dir
	5. our usual debug cycle: Mgba <-> gdb <-> ide (doesn't have to be detailed, but I think it would be nice to let people what the usual process is.)
2. A very obvious "Contribute" button that links to that document
   
Most of these topics are covered already across various PR comments, the wiki, but should be organized into a one-stop shop for developers. It would be safe to assume that people would read that before contributing. 

Here, I just made a pr for it: https://github.com/GBALATRO/balatro-gba/pull/334

MeirGavish:
> > > > I am so... so.. sorry. That sounds like an absolute nightmare 😲 .
> > > 
> > > 
> > > I'm exaggerating because I'm tired from work, don't worry it's not that bad ^^'
> > 
> > 
> > @ricfehr3 lol Honestly, maybe it's worth considering switching to something a little more forgiving that will make it a little easier on developers. I haven't checked but maybe `clang-tidy` can give us similar format checking that would not be so picky about whitespace etc.
> 
> I don't think we should get rid of `clang-format`. Besides looking nice, It's a pretty standard tool and It helps with PR reviews as it's both consistent and points formatting issues we would otherwise have to find and comment on. However, it shouldn't be this much of a pain in the ass, lol. `clang-format` as a standalone executable should be sufficient to not have to manually find and fix these issues:
> 
> ```shell
> clang-format -i source/blind.c
> ```
> 
> @MathisMartin31 I think something may be configured incorrectly in your IDE. I remember you mentioned running VSCodium. I downloaded it, installed the [extension](https://marketplace.visualstudio.com/items?itemName=xaver.clang-format) (recommended [here](https://clang.llvm.org/docs/ClangFormat.html#visual-studio-code-integration)), then hit `ctrl-shift-I` (I got it from [here](https://code.visualstudio.com/docs/cpp/cpp-ide#_code-formatting)) and it was able to format the file in-place. I thought this was an automatic check in vscode, but I should have tried and verified it earlier, that's my bad.
> 
> As for `clang-tidy`, I did mess with it earlier, and I don't think we should use it. It's not a code formatter, but a linter and static analyzer. Which, would be really cool, but I think that would be a much bigger task with a lot of future upkeep (I think `clang-format` and -Wall is sufficient). It also wouldn't work as a formatter like with `clang-format.`
> 

Alright, clang-format it is. I agree that it's good to have an automatic formatting tool, I just wish it was a little less strict about some things such as trailing whitespace. I guess it would be better if we could manage to get in-editor warnings and then there are no surprises in the CI.

ricfehr3:
I've already approved this, but I went back and took another look after your changes, and I think this one is good. :100: ! Good job.

Mega-Evolved-Ziquinho:
@MeirGavish  Just reminding you that a lot of these pull requests need your approval but are otherwise done

MeirGavish:
> @MeirGavish Just reminding you that a lot of these pull requests need your approval but are otherwise done

I will give this a look, but as I said this will go in after the release.

MathisMartin31:
> The only thing about this is I think it's better to not set the global variable inside the function and also not to set both of the variables in the same function. I think it would be better to separate it into 2 functions and have them each return values instead.
> This will make the functions a little more generic and less reliant on the global variables.

@MeirGavish I see your point, it does require iterating on all (in the worst case i.e. a High Card) possible contained hand types to determine what the current hand type is, but it doesn't look too bad compared to how it was done before.

However, I do personally prefer setting both global vars in the same function like I do currently, because I can't, on the top of my head, think of a situation where we would compute the current hand type but not want to update the contained hand types, and there is no point updating the contained hand types if the current hand type hasn't changed.

MeirGavish:
> > The only thing about this is I think it's better to not set the global variable inside the function and also not to set both of the variables in the same function. I think it would be better to separate it into 2 functions and have them each return values instead.
> > This will make the functions a little more generic and less reliant on the global variables.
> 
> @MeirGavish I see your point, it does require iterating on all (in the worst case i.e. a High Card) possible contained hand types to determine what the current hand type is, but it doesn't look too bad compared to how it was done before.
> 
> However, I do personally prefer setting both global vars in the same function like I do currently, because I can't, on the top of my head, think of a situation where we would compute the current hand type but not want to update the contained hand types, and there is no point updating the contained hand types if the current hand type hasn't changed.

It's more of a logical design consideration than a performance one, to make the functions more generic and independent and rely less on global state. It makes sense that `get_hand_type` will need to rely on the result of `hand_get_contained_types` and not have to duplicate this calculation. While we don't have to optimize for a case where the hand type is needed but not the contained hands, it could make sense to do it the other way around. The jokers rely on the cached global variable of contained hands, but they should be able to also calculate it independently if for example the cache is invalid for some reason and we don't want to rely on it.
It also makes sense for testing, we could test the hand prioritizing logic independently if it's separated to a function `get_hand_type()` that just receives the `ContainedHandTypes` and doesn't have or rely on any global variable state.
I also hope that `hand_get_contained_types()` could avoid relying on the hand global variable and could just receive a hand struct/list/array (probably best to have a struct that abstracts the array/list implementation) so it can also be tested but this is for another refactor related to #174.

MathisMartin31:
> > > The only thing about this is I think it's better to not set the global variable inside the function and also not to set both of the variables in the same function. I think it would be better to separate it into 2 functions and have them each return values instead.
> > > This will make the functions a little more generic and less reliant on the global variables.
> > 
> > @MeirGavish I see your point, it does require iterating on all (in the worst case i.e. a High Card) possible contained hand types to determine what the current hand type is, but it doesn't look too bad compared to how it was done before.
> > 
> > However, I do personally prefer setting both global vars in the same function like I do currently, because I can't, on the top of my head, think of a situation where we would compute the current hand type but not want to update the contained hand types, and there is no point updating the contained hand types if the current hand type hasn't changed.
> 
> It's more of a logical design consideration than a performance one, to make the functions more generic and independent and rely less on global state. It makes sense that `get_hand_type` will need to rely on the result of `hand_get_contained_types` and not have to duplicate this calculation. While we don't have to optimize for a case where the hand type is needed but not the contained hands, it could make sense to do it the other way around. The jokers rely on the cached global variable of contained hands, but they should be able to also calculate it independently if for example the cache is invalid for some reason and we don't want to rely on it.
> It also makes sense for testing, we could test the hand prioritizing logic independently if it's separated to a function `get_hand_type()` that just receives the `ContainedHandTypes` and doesn't have or rely on any global variable state.
> I also hope that `hand_get_contained_types()` could avoid relying on the hand global variable and could just receive a hand struct/list/array (probably best to have a struct that abstracts the array/list implementation) so it can also be tested but this is for another refactor related to #174.

That's a whole lot of good reasons you got there, I'll get to it ^^

MathisMartin31:
@MeirGavish This is my proposed solution:

- Have separate `compute` and `get` functions for both the `hand_type` and `_contained_hand_types` variables.
- Instead of simply iterating through the `_contained_hand_types` bitfield starting from the end, I ensure the bits are in the right order (4 OAK was in the wrong place after 3 OAK) and do a binary search by hand for the leftmost bit set to 1. This ensures that even in the worst case scenario of a low level hand, we find the scoring hand type in less than 4 comparisons instead of 13.

I had to make the ContainedHandTypes struct a union with a `value` field to get the total value of all the fields together and do comparisons with that.

I did some tests by playing a few hands and having Jokers that check against contained hands like the Duo, Trio, Family, Order and Tribe Jokers which all worked fine.

source/game.c
```diff
@@ -1277,92 +1268,195 @@ void hand_get_type(void)
...
+                    return HIGH_CARD;
+                }
+            }
+            else
+            {
+                return NONE;
+            }
         }
     }
 }
```
MeirGavish, line 1374: I don't think this hardcoded binary search with magic numbers is warranted. It's very difficult to read, verify, and debug. It's not really on any critical path to require such aggressive optimization. I think it's better to just iterate through. Actually, thinking about it, `contained_types` could be a bitset, but it would need reverse iteration, because you want to return the highest hand contained. But that' just a thought, you can forego the bitset implementation and just do a simple if-else here in the order of hand type values from most valuable to least.

MathisMartin31: Yeaaaaah, you're right. I don't know why I thought this was necessary. Maybe I just don't picture correctly in my head what kind of time it takes for the GBA to do computations, plus this PR was for optimization purposes so I thought maybe going the extra mile was necessary.

I'll just do a simple loop instead ^^

MathisMartin31: @MeirGavish my latest commit should address this

> Actually, thinking about it, contained_types could be a bitset, but it would need reverse iteration

That's what I did, iterating from `FLUSH_FIVE` down to `NONE` (making sure we don't underflow lol, I had that happen to me at work) and shifting/masking the value of `contained_types` to only keep the bit we want to check

Issue #217, opened 2025-11-08
[BUG] Incorrect Jokers/Cards activated when retriggering a card


MathisMartin31:
**Describe the bug**
When retriggering a Card that has been scored, the card to its left will activate "On Card Scored" Jokers.

**To Reproduce**
With Sock and Buskin, play a low normal card, and a face card. The first played card will not be scored, then the face card will score, S&B will activate, prompting a retrigger. The first, non scored card will then activate the Jokers it normally would when scored.

**Expected behavior**
The retrigger should make the previously scored card score again, without having a previous, non scored card, activating its Jokers.

**Videos**

https://github.com/user-attachments/assets/4c4e7d63-31f8-4ae8-b048-e13887167a62

**What version**
Main branch

**What Emulator or Platform were you playing on**
 mgba

MathisMartin31:
I already have a fix though, I just wanted to expose the bug, clean up my code, and ask everyone if I should create a new PR for this or merge it along with #205 

https://github.com/user-attachments/assets/24924eb4-418c-4ecc-aa24-28847393cac7

ricfehr3:
Nice work and thanks for taking the time to document this in an issue! If fixing the bug depends on any work in #205 , I would put the fix in there. Otherwise, I would say do a separate PR.

MathisMartin31:
Yeah it's more related to #160 as it was this PR that introduced retriggers. I'll open a new one

MeirGavish:
Good job catching this and opening an issue!

PR #218, opened 2025-11-08
Fix bug occurring when a retriggered card has a non-scoring card to the left

MathisMartin31:
Fixing the bug introduced with PR #160 and documented in issue #217 

I tested different retrigger Jokers to make sure the issue was not with how some Jokers were implemented, but rather that it was with how we handle the retrigger when scoring the cards/Jokers.

The corrected version now properly keeps track of what card was scored before, to go back to this index, instead of the current index -1 which is a potentially non scoring card.

Also simplified the use of `played_selections` and `scored_card_index` to avoid confusion when scoring cards. The two variables were used to the same goal in this part of the code, but they have two wildly different names that make sense in their respective context, so I don't know if one should be entirely dropped in favor of the other, or if keeping both would be better for readability.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -2255,7 +2255,7 @@ static void played_cards_update_loop(bool* discarded_card, bool* sound_played)
                         return;
                     }
                     
-                    if (card_object_is_selected(played[i]) && played_top - i >= *played_selections)
+                    if (card_object_is_selected(played[i]) && played_top - i >= scored_card_index)
                     {
                         played_y -= int2fx(10);
                     }
@@ -2283,25 +2283,14 @@ static void played_cards_update_loop(bool* discarded_card, bool* sound_played)
 
                         if (scored_card_index > 0 && card_object_is_selected(played[previous_scored_card_index]))
                         {
-                            if (check_and_score_joker_for_event(&_joker_scored_itr, played[previous_scored_card_index]->card, JOKER_EVENT_ON_CARD_SCORED))
+                            if (check_and_score_joker_for_event(&_joker_scored_itr, played[previous_scored_card_index], JOKER_EVENT_ON_CARD_SCORED))
                             {
-                                if (check_and_score_joker_for_event(&_joker_scored_itr, played[*played_selections - 1], JOKER_EVENT_ON_CARD_SCORED))
-                                {
-                                    return;
-                                }
-                            
-                                // Trigger all Jokers that have an effect when a card finishes scoring
-                                // (e.g. retriggers) after activating all the other scored_card Jokers normally
-                                _joker_scored_itr = list_itr_create(&_owned_jokers_list);
-                                if (check_and_score_joker_for_event(&_joker_scored_itr, played[*played_selections - 1], JOKER_EVENT_ON_CARD_SCORED_END))
-                                {
-                                    return;
-                                }
+                                return;
                             }
 
                             // Trigger all Jokers that have an effect when a card finishes scoring
                             // (e.g. retriggers) after activating all the other scored_card Jokers normally
-                            if (check_and_score_joker_for_event(&_joker_card_scored_end_itr, played[previous_scored_card_index]->card, JOKER_EVENT_ON_CARD_SCORED_END))
+                            if (check_and_score_joker_for_event(&_joker_card_scored_end_itr, played[previous_scored_card_index], JOKER_EVENT_ON_CARD_SCORED_END))
                             {
                                 return;
                             }


MeirGavish:
Ah, and our `played_selections` worries manifest. I was so wrong to think of retriggers as not so complicated. This code is quite a spaghetti that needs to be untangled.

source/game.c
```diff
@@ -2210,59 +2213,64 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
 
-                                return;
+                            // store last played card for next Jokers and possible retriggers
+                            // and search the next scoring card
+                            previous_scored_card_index = scored_card_index;
+                            do
+                            {
+                                scored_card_index++;
                             }
+                            while (!card_object_is_selected(played[scored_card_index]));
```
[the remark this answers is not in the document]
MathisMartin31: For both comments, although there seemed to be issues with out of bounds array accesses, the code looked like it was working fine. I still added both suggested checks and tested the code

source/game.c
```diff
@@ -2210,59 +2213,69 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
-                            scored_card_index++; // Count the number of cards that have been scored
+                        if (scored_card_index <= played_top)
+                        {
+                            CardObject* scored_card_object = played[scored_card_index];
 
-                            if (j <= played_top && card_object_is_selected(played[j]))
+                            if (card_object_is_selected(scored_card_object))
                             {
-                                tte_set_pos(fx2int(played[j]->sprite_object->x) + 8, SCORED_CARD_TEXT_Y); // Offset of 16 pixels to center the text on the card
+                                tte_set_pos(fx2int(scored_card_object->sprite_object->x) + 8, SCORED_CARD_TEXT_Y); // Offset of 1 tile to keep the text on the card
```
```suggestion
                                tte_set_pos(fx2int(scored_card_object->sprite_object->x) + TTE_CHAR_SIZE, SCORED_CARD_TEXT_Y); // Offset to keep the text centered on the card
```

If you are going to change this unclear documentation, remove the magic number.

source/game.c
```diff
@@ -2210,59 +2213,64 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
+                            // (e.g. retriggers) after activating all the other scored_card Jokers normally
+                            if (check_and_score_joker_for_event(&joker_card_scored_end_index, played[previous_scored_card_index]->card, JOKER_EVENT_ON_CARD_SCORED_END))
+                            {
+                                return;
+                            }
+                        }
 
-                            scored_card_index++; // Count the number of cards that have been scored
+                        // Score card
+                        CardObject* scored_card_object = played[scored_card_index];
```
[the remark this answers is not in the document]
MeirGavish: And this adds yet another level of indentation 😭
But I think it's better to have this guard, can't say for sure.

source/game.c
```diff
@@ -2210,59 +2213,69 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
                         if (retrigger)
                         {
                             retrigger = false;
-                            scored_card_index--;
-                            (*played_selections)--;
-                            joker_scored_index = 0;
+                            scored_card_index = previous_scored_card_index;
                         }
```
MeirGavish, line 2278: The comment above this block is now inaccurate

source/game.c
```diff
@@ -2299,7 +2312,7 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
 
                         play_state = PLAY_ENDING;
                         timer = TM_ZERO;
-                        *played_selections = played_top + 1; // Reset the played selections to the top of the played stack
+                        scored_card_index = played_top + 1; // Reset the played selections to the top of the played stack
```
```suggestion
                        scored_card_index = played_top + 1; // Reset the scored card index to the top of the played stack
```

The comment needs to be updated to reflect the different variable...

source/game.c
```diff
@@ -2210,59 +2213,69 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
...
+                            previous_scored_card_index = scored_card_index;
+                            do
+                            {
+                                scored_card_index++;
+                                if (scored_card_index > played_top)
+                                {
+                                    return;
+                                }
                             }
+                            while (!card_object_is_selected(played[scored_card_index]));
```
MeirGavish, line 2337: It's weird that such a loop was not needed before, do you understand how it skipped non-selected played cards before?

source/game.c
```diff
@@ -2201,6 +2203,7 @@ static void played_cards_update_loop(bool* discarded_card, int* played_selection
 
                     if (i == 0 && (timer % FRAMES(30) == 0) && timer > FRAMES(40))
                     {
+                        // So pretend "played_selections" is now called "scored_card_index" and it counts the number of cards that have been scored
```
MeirGavish: I'm not sure this comment still means what it was supposed to mean originally...

source/game.c
MathisMartin31: Yes I think I know how it worked before.

There was a for loop that went from the current index to the hand size, so each time we ran the code we could pick up from where we left off. We incremented the index whether the card scores or not, but we returned if it did. This way, non scoring cards could be skipped in a single frame, and scoring a card interrupt the code flow to give time for the animation to play out before resuming scoring.

This is basically also how it's done with Jokers (iterate starting from index, increment index, return when scoring, let animation play, resume from index, repeat) but here it was problematic for retriggers because the previously triggered card, the one we need to go back to, could be several cards away from the current index.

The solution I came up with was to save the index of the card each time we score one, and iterate over cards in a while loop until we either encounter the end of the hand, or find a scoring card. This way, when we need to retrigger, we only need to recover the index of the previous scoring card we saved earlier.

MathisMartin31:
> This worries me immensely since this is a partial replace of `played_selections` but I can't say how I'd change it (make it a full replace?)

I'll see if `played_selection` needs replacement elsewhere in the function, but even if we don't use it when `scored_card_index` and vice versa, they are well named for their respective contexts, not the others'.

What I mean is that I replaced the former with the latter when scoring cards, but I don't know if we should do it when playing selected cards.

I also don't know if we should define `played_selections` properly somewhere (instead of declaring a static var within a function) and access it as a global variable, the same way as `scored_card_index` and other global vars.

I did both in my last commit, along with your other suggestions, let me know what you think

source/game.c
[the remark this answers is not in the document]
MathisMartin31: I should have already done this extra level of indentation in one of my first commits, and have moved the assignment of `scored_card_object` after the range check, as I should have done from the start

source/game.c
MathisMartin31: I replaced `played_selections` entirely so I remove the comment as well

MathisMartin31:
I think so yes, `played_selections` was only used in 7 places after the partial replace and was defined kind of weirdly. Having it being the same variable is a good trade off imo

MeirGavish:
> I think so yes, `played_selections` was only used in 7 places after the partial replace and was defined kind of weirdly. Having it being the same variable is a good trade off imo

Okay, great! It seems better to me too. So basically it's a rename and a use of the same variable. I merged #168, once this PR is updated I will give it one last check and merge.

source/game.c
MeirGavish: Cool, thanks!

MathisMartin31:
I used git rebase, but I still ended up with a lot of commits from the memory pool PR. Marginally better than a single merge commit but I hope it's not too bothersome ^^'

MeirGavish:
@MathisMartin31 It looks like you did both a rebase and a merge. I shouldn't be seeing all the changes and commits from the other PRs, it seems like you need to `git reset --hard` this branch to [commit a674d50](https://github.com/cellos51/balatro-gba/pull/218/commits/a674d50d622b784f7fdb32b3aab2817ad2496008) (you might need the full hash) before the merge and then force push it. Make a backup branch of this branch first just in case.

MathisMartin31:
> you need to git reset --hard this branch to [commit a674d50](https://github.com/cellos51/balatro-gba/pull/218/commits/a674d50d622b784f7fdb32b3aab2817ad2496008) (you might need the full hash) before the merge and then force push it

It was so long since the last time I had to rebase that I forgot to do that instead of normal pull/push, the branch should be fixed now ^^

MeirGavish:
I tested this now and it seems like Hanging Chad is broken...

I think this happens when the card activates a different joker.

https://github.com/user-attachments/assets/137fcf26-d1ba-433b-bacc-0eb6cf2655c4

MathisMartin31:
But hoooow I tested them before commiting T_T

I'll take a look, thanks for the video :)

MathisMartin31:
@MeirGavish I rebased my branch on main and tested it with Fibonacci like you did plus other Jokers and it worked fine

https://github.com/user-attachments/assets/b80fb27a-e5b3-4984-a1ef-2c6db1c00239

MeirGavish:
Okay, I do have a backup of this branch before the rebase and it does occur there, I think that this issue was fixed by #205.
I think what caused this bug is that Hanging Chad was activated on the last hand of the round and `JOKER_EVENT_ON_HAND_SCORED_END` was then never reached and the number of retriggers was never reset, this was then fixed by moving the reset to before the hand is scored. I didn't fully test to confirm this so this is more of an educated guess. 
If it's true it might mean Seltzer has the same problem since it uses `JOKER_EVENT_ON_HAND_SCORED_END` but it's not integrated so I will not test it.
It might also mean the issue occurred in main before #205 but I'm not sure.
Stay on the lookout for more bugs.

Issue #222, opened 2025-11-12
Integrate Photograph


MeirGavish:
Optimize palette for Photograph art by @MathisMartin31 and integrate it into the game

MeirGavish:
Already doing this but I thought I might open an issue in the meantime

PR #223, opened 2025-11-12
Integrated Photograph joker art by @MathisMartin31

MeirGavish:
closing #222 
I did one more color optimization for Photograph and managed to remove one more color (used color from the outline in the hand), so it now uses 15, making it possible to pair him with Stencil.
I had to do some joker swapping so the color pairings are still good, I put Fibonacci last since it doesn't use that many colors so it should be okay if we add another joker with it.


MathisMartin31:
Oh damn, I didn't see we could pair Photograph with Stencil good job ^^

MeirGavish:
> Oh damn, I didn't see we could pair Photograph with Stencil good job ^^

Yeah, Stencil and Business Card use very few colors, I still had to remove one color from Photograph for it to work.

MathisMartin31:
Red Card would have also been a very good candidate but it's completely out of scope since we don't have any card Packs

Issue #220, opened 2025-11-10
Clean Up `played_cards_update_loop()`


MeirGavish:
The function `played_cards_update_loop()` is very long and has many nested layers, it needs to be broken up into smaller functions and cleaned up. Try to also improve the logic flow and make it clearer, limit the scope of variables to where they are logically needed, and make their usage and meaning clearer by either consolidating duplicates, renaming, or adding documentation.

`cards_in_hand_update_loop()` has similar problems and needs a similar treatment but that doesn't have to be a part of this issue.

MathisMartin31:
Since I'm starting to know my way around this part of the code, maybe I can take care of this issue after PRs #205 #212 and #218 are merged?

MeirGavish:
@MathisMartin31 First break it up into smaller functions in an initial commit/PR and then continue to other refactors there if needed.

Issue #156, opened 2025-09-27
Encapsulate game variables and move game states into their own files.


ricfehr3:
🛑 This is blocked on: https://github.com/cellos51/balatro-gba/pull/155

We currently have game variables described in the scope of `game.c`. Encapsulating the game state variables into a struct would allow files outside of `game.c` to modify the game state variables:

```c
typedef struct
{
    int score;
    GameState current_state;
    ...
} GameVariables;
```

This would allow cleaning up the state machine functions by moving them to their own files. One set of `.c`/`.h` per game state. This new game state struct could be passed into the state machine callbacks or via a getter function: `get_game_variables()`. 

This will allow further cleaning `game.c` and will allow for an easier time moving the state machine logic into it's own file.

MeirGavish:
I think separating the game states to their own files would be great. I wonder if maybe it would be better to separate to two structs. 
One for the variables for the entire run - money, jokers, etc. 
And one for a single round - number of hands left, number of discards left etc.
Then the latter can maybe stay in a game round file.
These are all thoughts I'm having without looking at the code so I don't know if it might be too much.

ricfehr3:
I think you're right in separating the variables into separate structs. I don't think it would be too much. 

Accessing these structs should be through a function that returns a pointer to a static instance. Maybe we should separate a `GameVars` vs `RunVars` and either have them accessible on their own or take it one step further with `struct AllVars { GameVars game_vars; RunVars run_vars }`? That's a decision for who writes it too.

MathisMartin31:
@ricfehr3 I also think this would be a great idea! I quickly spoke about it in my latest message in issue #103 , adding a third struct being `HandVars` for retriggers to be easier to implement by keeping track of a few additional things, like current Card/Held Card/Joker being scored, outside of the scoring loop

ricfehr3:
> [@ricfehr3](https://github.com/ricfehr3) I also think this would be a great idea! I quickly spoke about it in my latest message in issue [#103](https://github.com/cellos51/balatro-gba/issues/103) , adding a third struct being `HandVars` for retriggers to be easier to implement by keeping track of a few additional things, like current Card/Held Card/Joker being scored, outside of the scoring loop

I can work on this again this weekend. Thanks for referencing this :D . I'll try to make a friendly but detailed description of the work involved with that too.

MathisMartin31:
@ricfehr3 Nice! That'd be awesome ^^

[...]

PR #224, opened 2025-11-12
Break up the played cards scoring loop into several functions

MathisMartin31:
From issue #220 

This PR aims at cleaning up the spaghetti code that is the `played_cards_update_loop` function.

The first step is to break up the different cases into separate functions, then move them into different files when the time is right and game state variables are better encapsulated.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -2493,7 +2493,7 @@ static void played_cards_update_loop()
             
             case PLAY_BEFORE_SCORING:
 
-                if (play_before_scoring_cards_update(played_idx))
+                if (play_before_scoring_cards_update())
                 {
                     return;
                 }


source/game.c
```diff
@@ -2203,10 +2203,265 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+                }
+
+                play_state = PLAY_PLAYING;
+                cards_drawn = 0;
+                hand_selections = 0;
+                played_top = -1; // Reset the played stack
+                scored_card_index = 0;
+                _joker_scored_itr = list_itr_create(&_owned_jokers_list);
+                timer = TM_ZERO;
+                return; // return early to avoid accessing an invalid index
```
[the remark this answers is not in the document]
MathisMartin31: I did this because with the rework, returning early in the sub function leads to a break statement. Plus, I can't break here because I'm technically not in a loop/switch.

Although, now I do question the usefulness of the prior early break, I'll have to look into it

source/game.c
```diff
@@ -2203,10 +2203,265 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+        }
+    }
+
+    if (card_object_is_selected(played[i]) && played_top - i <= scored_card_index - 1)
+    {
+        (*played_y) -= int2fx(10);
+    }
+}
+
+static void play_ended_update(bool* discarded_card, bool* sound_played, int i, FIXED* played_x, FIXED* played_y)
```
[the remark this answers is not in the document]
MathisMartin31: Maintaining "consistency" would mean passing 3 useless arguments to 6 other functions, so I don't know if we would be gaining much clarity doing that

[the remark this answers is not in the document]
ricfehr3: I wouldn't listen to this advice :-1:

ricfehr3:
One thing here that is getting to me is passing around a lot of pointers related to the state of the hand. I don't think we should be passing around multiple pointers to bools.

I think this PR would benefit from capturing all the variables that are passed as pointers and putting them into some sort of "hand state" struct or similar. Then, we can pass a pointer to that struct in each function and operate on that struct and choose variables from it to modify. 

My reasoning is:
1. Less pointers means less problems.
2. passing one struct allows for calls to be modified easier that may need more info from other parts of the struct in the future (not likely, but nice to have).
3. Consistency for all hand modifying functions.

source/game.c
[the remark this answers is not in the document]
MeirGavish: I actually think the break there was a mistake and this is meant to exit the for loop, not the switch case (probably with a return), since this block uses `card_object_destroy()` so when continuing after the switch case there **is** access to an invalid index.

source/game.c
```diff
@@ -2203,10 +2203,265 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+            _joker_scored_itr = list_itr_create(&_owned_jokers_list);
+            timer = TM_ZERO;
+            play_state = PLAY_BEFORE_SCORING;
+        }
+    }
+
+    if (card_object_is_selected(played[i]) && played_top - i >= scored_card_index)
+    {
+        (*played_y) -= int2fx(10);
+    }
```
MeirGavish: What bothers me is this block (also mentioned in [previous PR comment](https://github.com/cellos51/balatro-gba/pull/205#discussion_r2515058892)), that's copied for all cases with minor changes. It especially bothers me that it's copied in cases that have nothing to do with the played cards such as `PLAY_SCORING_JOKERS` and `PLAY_SCORING_HELD`. I think the correct thing to do is actually remove the `played_y` and `played_x` variables altogether and let the functions update `played[i]->sprite_object->tx` and `played[i]->sprite_object->ty` directly.
Then functions that don't need to update it just won't.

Define a constant for the initial magic numbers
```
BG_POINT CARDS_PLAYED_POS = { 120, 70 };
```

I've been trying to write a code suggestion for how this would be done but at this point I think it would be easier for me to just do it so you can also disregard this comment and I can do it, it might be a bit of a trickier refactor.

source/game.c
```diff
@@ -2203,10 +2203,265 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
     return false;
 }
 
-static void played_cards_update_loop(bool* discarded_card, bool* sound_played)
+static void play_playing_update(int i, FIXED* played_y)
```
```suggestion
static void play_playing_update(int played_idx, FIXED* played_y)
```

Now that it's a parameter to a function it needs a clearer name (same goes for the other functions)

source/game.c
```diff
@@ -2203,10 +2203,265 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+    }
+
+    if (card_object_is_selected(played[i]) && played_top - i >= scored_card_index)
+    {
+        (*played_y) -= int2fx(10);
+    }
+}
+
+// returns true if the scoring loop has returned early
+static bool play_before_scoring_update(int i, FIXED* played_y)
```
MeirGavish: This function can be further broken down

source/game.c
MeirGavish: Also this function name is too generic, I think the state name needs to be changed from PLAY_PLAYING to PLAY_STARTING and the function name should match it.

source/game.c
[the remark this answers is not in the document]
MeirGavish: So as mentioned in a previous comment, I think the correct thing to do here is to actually remove `played_y` and `played_x` altogether.

MathisMartin31:
> I think this PR would benefit from capturing all the variables that are passed as pointers and putting them into some sort of "hand state" struct or similar

I don't know if I should do it in this PR or in another smaller one, but maybe it's our cue to address issues #156 and #174?

source/game.c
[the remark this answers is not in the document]
MathisMartin31: Then there's no more need to return early now is there? This block leads directly to the end of the function and the state has been changed so we will not pass through the problematic index access.

Only thing that would worry me is that `*discarded_card = true`, I'll have to look into if it's problematic or not

source/game.c
```diff
@@ -2229,253 +2484,50 @@ static void played_cards_update_loop(bool* discarded_card, bool* sound_played)
...
-                
-                    if (card_object_is_selected(played[i]))
+                    if (play_scoring_jokers_update(i, &played_y))
                     {
-                        played_y -= int2fx(10);
+                        return;
                     }
                     break;
 
                 case PLAY_ENDING: // This is the reverse of PLAY_PLAYING. The cards get reset back to their neutral position sequentially
```
MeirGavish: I think the comments for the cases should also be moved into their respective functions. Also this specific comment needs to be changed when the name of the state is changed from `PLAY_PLAYING` to `PLAY_STARTING` - it makes it all make sense along with PLAY_ENDING.

source/game.c
```diff
@@ -2203,10 +2203,265 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+        *discarded_card = true;
+    }
+    return;
+}
+
+static void played_cards_update_loop(bool* discarded_card, bool* sound_played)
+{
     // So this one is a bit fucking weird because I have to work kinda backwards for everything because of the order of the pushed cards from the hand to the play stack
     // (also crazy that the company that published Balatro is called "Playstack" and this is a play stack, but I digress)
     for (int i = 0; i <= played_top; i++)
```
MeirGavish: It doesn't let me add a comment for the `if (played[i] != NULL)` block but we can further reduce nesting levels by inverting the condition

```
if (played[i] == NULL)
{
     continue;
}
```

source/game.c
```diff
@@ -2203,10 +2203,265 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+    }
+}
+
+static void play_ended_update(bool* discarded_card, bool* sound_played, int i, FIXED* played_x, FIXED* played_y)
 {
-    // TODO: Break this function up into smaller ones.
+    if (!*discarded_card && played[i] != NULL && timer > FRAMES(40))
+    {
+        (*played_x) = int2fx(240);
+        (*played_y) = int2fx(70);
```
```suggestion
        (*played_x) = int2fx(CARD_DISCARD_PNT.x);
        (*played_y) = int2fx(CARD_DISCARD_PNT.y);
```
Since we're at it, let's remove this magic number, there's already an existing constant for it.

MeirGavish:
> One thing here that is getting to me is passing around a lot of pointers related to the state of the hand. I don't think we should be passing around multiple pointers to bools.
> 
> I think this PR would benefit from capturing all the variables that are passed as pointers and putting them into some sort of "hand state" struct or similar. Then, we can pass a pointer to that struct in each function and operate on that struct and choose variables from it to modify.
> 
> My reasoning is:
> 
> 1. Less pointers means less problems.
> 2. passing one struct allows for calls to be modified easier that may need more info from other parts of the struct in the future (not likely, but nice to have).
> 3. Consistency for all hand modifying functions.

Yeah, the existing bools `sound_played` and `discarded_card` are being passed between the `cards_in_hand_update_loop()` and `played_cards_update_loop()`, maybe the easiest thing would be to make them global. Otherwise I guess it can be a struct but this wouldn't be a generic hand struct, these variables are very specific to the playing state so it could be a playing struct. Also I think maybe these variables can be more correctly scoped and not reused in several places where they are not necessarily related.
I think anything other than simply outright making them global should really go in another PR.

source/game.c
MathisMartin31: I'm working on this but some things have started to break so I'm busy fixing it but it's tedious. Some checks and conditions are really obscure and completely undocumented, I'll try to add some useful comments.

MeirGavish: Yeah, I said it is a bit tricky and can be postponed to another PR.

MathisMartin31: I already fixed #226 and I think it's related enough to this PR that it's worth the extra effort. Plus, the broken up functions will be way cleaner. It's just taking a bit of time to wrap my head around the code

MathisMartin31: Okay I had only one problem and I was just being a big dum dum, I didn't realize I put `return` instead of `continue` when implementing [this suggestions](https://github.com/cellos51/balatro-gba/pull/224#discussion_r2522603751) :facepalm: 

So when freeing the first discarded card, we would always check it, find it's NULL and return, rince and repeat and you get a deadlock.

I only realized it by displaying a lot of stuff with `tte_printf`

MeirGavish: Oh, you can [connect a debugger](https://github.com/cellos51/balatro-gba/discussions/80) to debug instead of using `tte_printf`

source/game.c
MathisMartin31: I don't really know what to call it except `play_starting_update` now :/

source/game.c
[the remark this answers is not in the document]
MathisMartin31: In the end I kept the early return (with proper return) because as you said, it does avoid accessing a card_object we just freed

source/game.c
MathisMartin31: I divided it between the scoring cards state, and a new scoring "on scored" jokers state. This way we also don't need to do the weird thing we were doing before. It was something along the lines of beginning by trying to score the Jokers from the previous card, first card excluded, then score the card and repeat. It made sense to do it this way to score both cards and their Jokers in the same state, but we don't need it with two separate states

MathisMartin31:
@MeirGavish @ricfehr3 

All requested changes should be addressed by now, the only thing preventing us from putting the new state update functions into a separate file should be to encapsulate global variables into structs

source/game.c
MeirGavish: So just to make sure I understand, you broke it out to another state that is called from `played_cards_update_loop()` rather than from here, right? Also, we're still scoring the joker for the previous card rather than the current one scored, right? And you mean that separating it made sense because the joker is not activated for the same card currently scored, right?

source/game.c
MeirGavish: How about `play_starting_played_card_update`? To indicate it updates the played card. Same goes for `play_ending`.

include/game.h
```diff
@@ -50,11 +50,12 @@ enum HandState
...
 enum PlayState
 {
-    PLAY_PLAYING,
+    PLAY_STARTING,
     PLAY_BEFORE_SCORING,
     PLAY_SCORING_CARDS,
-    PLAY_SCORING_HELD,
     PLAY_SCORING_JOKERS,
+    PLAY_SCORING_HELD,
+    PLAY_SCORING_INDEPENDENT,
```
MeirGavish:
```suggestion
    PLAY_SCORING_CARD_JOKERS,
    PLAY_SCORING_HELD,
    PLAY_SCORING_INDEPENDENT_JOKERS,
```

These need to be renamed, the scoring jokers state is now used only for the scored cards rather than all jokers and it needs to reflect that.
The independent state needs to reflect it's scoring jokers.
The corresponding functions need to be renamed too.

source/game.c
```diff
@@ -2203,287 +2208,351 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+    played[played_idx]->sprite_object->ty = int2fx(HAND_PLAY_POS.y);
+
+    if (card_object_is_selected(played[played_idx]) && played_top - played_idx >= scored_card_index)
+    {
+        played[played_idx]->sprite_object->ty -= int2fx(10);
+    }
+}
+
+// returns true if the scoring loop has returned early
+static bool play_before_scoring_update(int played_idx)
```
```suggestion
static bool play_before_scoring_update()
```

Now functions that don't use `played_idx` don't need to pass it.

source/game.c
```diff
@@ -2203,287 +2208,351 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
     return false;
 }
 
-static void played_cards_update_loop(bool* discarded_card, bool* sound_played)
+
+static void play_starting_update(int played_idx)
```
MeirGavish: Putting a comment to my rename suggestion in my [reply](https://github.com/cellos51/balatro-gba/pull/224#discussion_r2529897087).
The state should stay with the same name however.

source/game.c
```diff
@@ -2203,287 +2208,351 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
-                    {
-                        played_y -= int2fx(10);
-                    }
+    return false;
+}
 
-                    play_state = PLAY_SCORING_CARDS;
-                    break;
+// returns true if the scoring loop has returned early
+static bool play_scoring_held_update(int played_idx)
```
```suggestion
static bool play_scoring_held_card_update(int played_idx)
```

Maybe also rename this function and state

source/game.c
```diff
@@ -2203,287 +2208,351 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+
+    return false;
+}
 
-            switch (play_state)
+// Activate "on scored" Jokers for the previous scored card if any
+// returns true if the scoring loop has returned early
+static bool play_scoring_jokers_update(int played_idx)
+{
+    if (timer % FRAMES(30) == 0 && timer > FRAMES(40))
```
MeirGavish, line 2307: Just making sure - you are sure the `played_idx` check is not needed here, right? If so it can be removed as a parameter as this function scores the jokers for the previous scored card.

source/game.c
```diff
@@ -2203,287 +2208,351 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+    }
+
+    play_state = PLAY_SCORING_CARDS;
+    return false;
+}
+
+// returns true if the scoring loop has returned early
+static bool play_scoring_cards_update(int played_idx)
+{
+    if (timer % FRAMES(30) == 0 && timer > FRAMES(40))
```
MeirGavish, line 2250: Just making sure, you're sure you don't need the `played_idx` check here, right? If so, the `played_idx` parameter can be removed from this function as it doesn't use it.

source/game.c
```diff
@@ -2203,287 +2208,351 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+}
 
-                        if (scored_card_index == 0)
-                        {
-                            play_state = PLAY_ENDED;
-                            timer = TM_ZERO;
-                        }
-                    }
+// Basically a copy of HAND_DISCARD
+// returns true if played[played_idx] has been invalidated
```
```suggestion
// returns true if we reached the end of the play state
```
This is more correct, `played[played_idx]` being invalidated is just a side-effect of reaching the end of the state and resetting the variables.

source/game.c
```diff
@@ -2203,287 +2208,351 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+            case PLAY_ENDING:
+
+                play_ending_update(played_idx);
+                break;
+
+            case PLAY_ENDED:
+
+                if (play_ended_update(played_idx))
+                {
+                    continue;
```
MeirGavish, line 2547: Why is this a continue and not a return? Since `played_top` is reset the next iteration will exit anyway. If there's a reason, add a comment explaining it.

source/game.c
```diff
@@ -2203,287 +2208,351 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+            timer = TM_ZERO;
+        }
+    }
 
-                    if (i == 0 && (timer % FRAMES(10) == 0 || !card_object_is_selected(played[played_top - scored_card_index])) && timer > FRAMES(40))
-                    {
-                        scored_card_index--;
+    if (card_object_is_selected(played[played_idx]))
+    {
+        played[played_idx]->sprite_object->ty = int2fx(70);
```
```suggestion
        played[played_idx]->sprite_object->ty = int2fx(HAND_PLAY_POS.y);
```
Now that the constant is defined we can remove this magic number

MeirGavish (changes requested):
Looks great! Good job! Just a few comments.

MeirGavish:
I'm requesting another review from Copilot to make sure I didn't miss anything.

source/game.c
MathisMartin31: That's correct, but now that you say it like this I wonder why I didn't just put it into a function and called that from `play_scoring_cards_update`.

It's technically called the `previous_scored_card_index`, but really this is the card that we have just scored, it's just that seeking the next scoring index is done after scoring the first card.

I think I can rearrange thing a bit to make it a bit less convoluted

MeirGavish:
I see an issue here now, the cards slide back all together at once when the hand score is calculated, that's when the bounce of #226 happens, the cards should stay in their place and slide back one by one after a delay.

https://github.com/user-attachments/assets/7b407284-83e6-49cc-bd57-480048aa2f07

source/game.c
MathisMartin31: Okay I just did that. This is what happens now:

- Start by seeking the next scoring card. If it's the current one then we score that, so this works when starting with index 0

- Score the card, then go to the jokers scoring state

- Score all Jokers for the card we just scored

- If we need to retrigger, then return early without incrementing `scored_card_index`, else increment it and return normally

This way there is no need for a `previous_scored_card_index` and everything is laid out in the same order it's happening in game. I think it's less confusing this way

source/game.c
MathisMartin31: I renamed some functions following your suggestion, I think it does look more explicit

source/game.c
MathisMartin31: Yes I'm sure, leaving it would force the function to wait an extra tick (40 frames) every time we switch to this state

source/game.c
MathisMartin31: It's done to save some cycles. When we return true it means we discarded the card `played_idx`.

Returning would mean starting over from 0, then looping over `played_idx` until we find a non NULL card (see check at the beginning of the update loop), then start the process of discarding it.

Continuing allows to start the process of discarding the next card instantly by going to `played_idx+1`, instead of starting over from index 0.

It's a very small optimization but I think we should take everything we can, especially at the end of scoring when there's a lot going on visually

source/game.c
MathisMartin31: I defined it and completely forgot about it...

MathisMartin31:
> I see an issue here now, the cards slide back all together at once when the hand score is calculated, that's when the bounce of #226 happens, the cards should stay in their place and slide back one by one after a delay.

Oooooooooh I didn't even notice that damn. I'll take a look

Edit: should be fixed now

MeirGavish:
> > I see an issue here now, the cards slide back all together at once when the hand score is calculated, that's when the bounce of #226 happens, the cards should stay in their place and slide back one by one after a delay.
> 
> Oooooooooh I didn't even notice that damn. I'll take a look
> 
> Edit: should be fixed now

@MathisMartin31 Great! I'm not sure I'll have time to give it another pass today, but tomorrow if not.

source/game.c
```diff
@@ -2203,287 +2207,349 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+    played[played_idx]->sprite_object->ty = int2fx(HAND_PLAY_POS.y);
+
+    if (card_object_is_selected(played[played_idx]) && played_top - played_idx >= scored_card_index)
+    {
+        played[played_idx]->sprite_object->ty -= int2fx(10);
+    }
+}
+
+// returns true if the scoring loop has returned early
+static bool play_before_scoring_cards_update(int played_idx)
```
```suggestion
static bool play_before_scoring_cards_update()
```

The index is not needed here too.

source/game.c
```diff
@@ -2203,287 +2207,349 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
+                break;
+
+            case PLAY_ENDED:
+
+                if (play_ended_played_cards_update(played_idx))
+                {
+                    // we continue here instead of returning to instantly go to
+                    // the next card to discard, instead of starting over
+                    // from index 0 and going up to that card again
+                    continue;
```
```suggestion
                    // we continue here instead of returning for performance 
                    // to instantly go to the next card to discard at played_idx+1, 
                    // instead of  starting over from index 0 and going up
                    // to that card again 
                    continue;
```

Updated the comment a bit to clarify that it's for performance, and added explicit `played_idx+1` from your explanation just in case it makes it clearer. I don't fully understand it so if you think this is incorrect let me know

source/game.c
MeirGavish: Cool! Sounds good!

source/game.c
MathisMartin31: No no that's precisely what I meant to do there ^^

Issue #110, opened 2025-08-29
Music isn't finished


cellos51:
The current music that plays, ends abruptly a little less than half-way through the song. I don't really want to work on the music anymore, so if anyone who is comfortable working with trackers and would want to finish this, I would appreciate it. 

The MIDI I was basing the song on can be found here: https://musescore.com/user/8237321/scores/14590660

MeirGavish:
I was going to ask about this! I don't have a lot of experience with trackers or music/audio software but I have some musical background and it could be fun to try.
1. Which tracker did you use? 
2. Where did you get the samples? 
3. Also, how did you combine the samples for individual notes into single chord samples?

I've also been playing around to try improve audio quality and reduce noise but didn't manage to get good results.

cellos51:
I used milky tracker and the samples were sourced from demo tracks I found on the milky tracker website. I don't remember how I combined the chords but I'm pretty sure you can do it in milky tracker.

MeirGavish:
Cool, thanks! I'll try it out.

MeirGavish:
@cellos51 
By the way, I'm assuming that the reason you combined the chords into single samples is that separating them to multiple channels reduces the quality when they are mixed in by maxmod, right?

cellos51:
I merged them because they would use up more tracks and then the other sounds would cut out

MeirGavish:
I wonder if allocating more channels in mmInitDefault() would solve that problem... either way I think it's better to not have too many channels because it also increases the CPU usage. Thanks!

cellos51:
Yeah other mod songs I've seen from GBA homebrew have very few tracks. Most things are merged

MeirGavish:
I've removed my assignment for this task as I will probably not be able to work on it while on vacation, but I still want to work on it once I return if nobody else picks it up.
I will also document what I did so far, I mostly played around with the samples. I couldn't figure out how to mix chord samples in a tracker so I did it with Audacity and it worked well. Here are the stages for what I did,
1. Export the Rhodes sample from the tracker as a wav file
2. Layer it in different tracks in Audacity
3. Pitch shift each track to create the chord
4. Use the amplify effect to reduce the volume of all tracks so once they are mixed and added it's not too loud and clipping. I used -12db because there are 4 notes in the chord so I wanted 1/4  of the volume for each one.
5. Mix the tracks into a single track
[...]

PR #227, opened 2025-11-14
Music - Relayered chord samples

MeirGavish:
As part of issue #110 I relayered the chord samples as described in [this comment](https://github.com/cellos51/balatro-gba/issues/110#issuecomment-3270152683) except I skipped the -12dB volume decrease, the chords sounded too low in the mix if I did that and it doesn't seem like they were clipping after mixing without it.

I think these samples are better and it sounds better overall now, you can compile the rom, give it a listen, and tell me what you think. It's important to listen to the compiled version rather than the .xm file because maxmod changes it a bit when mixing and rendering.

A few more details:
What was bothering me was that 8-bit Music Theory's transcription notated the third and fourth chords differently - it notated an Am6 where the Musescore had B7sus4 - in the end I opted for the Musescore transcription because I also looked at [Luke Pickman's cover](https://youtu.be/Kx0jTGd7urs) and you can very clearly see he's playing B7sus4. I also played it along with the original and it sounds better and in general music theory-wise seems like a more natural functional chord progression.

I also changed the chord labeling a bit - I added the bass note to all the chords to clearly identify the inversion and also that same chord was labelled E9sus4 which seems incorrect so I changed it to B7sus4/F# (with the bass note).

Also the file had 8 channels but only used 6 so I deleted the extra unused channels, I don't know if it makes any actual improvement in quality or performance but I thought I might as well.

I solely used Milky Tracker to not introduce any file changes specific to other software but since OpenMPT feels a little more intuitive I'm not sure I'm gonna keep doing that.


ricfehr3:
Nice! I captured both versions of the tracks in audacity, lined them up, then played one out of the left channel and the other out of right channel. I swapped my headphones on the same ear and I _think_ I prefer the older version simply because the mixing lets the bass be a bit punchier. It sounds like there's a bit of a low gurgle introduced in the new mix that washes out the punch of the bass and it hits a bit less hard. 

I'm not really an expert, but I think any introduction of low frequencies probably will mess with the quality on real hardware too because those tiny speakers don't have a good bass response.

Let me know what you think, this is all subjective, of course.

MeirGavish:
Thanks for listening and pointing it out!
I'm not entirely sure how the original chord samples were made, they fade quicker and have less harmonical content, it seems the quantization introduces a low frequency band and it might be what you're hearing, I will check if I can improve it by reducing quantization noise and maybe filtering out the low frequencies so the chords don't compete with the bass at the lower frequencies. In the end I'm not sure I can overcome the quantization artifacts because any edits I make will be affected again by the export.

ricfehr3:
No worries at all, love to hear it (literally, haha). You seem way more musically inclined than I do, so I'll trust your process. I'll give it another listen if you update it.

MeirGavish:
As I thought denoising and exporting again in 8-bit just increases the noise so I guess any post-quantization processing done to combat it will be nullified when exported again.
I tried something else, instead of pitch shifting the sample itself to the target chord, I based all the chords on C like the base instrument sample and used the tracker notation to play the correct notes. I think it sounds much better. Not entirely sure why though. Maybe Maxmod or the tracker module does some sort of filtering to match the written note. The thing is, it requires more work because I in addition to changing the samples I need to go through the entire track and change the notation. 

I did it for the first few bars and I'm uploading a zip file with the compiled ROM and the xm file so you can give it a listen and tell me if you think it's better, I think I'll try doing that for the entire track.
[balatro-gba-music-relayered-cbased.zip](https://github.com/user-attachments/files/23675888/balatro-gba-music-relayered-cbased.zip)

GreenNinja2525:
@MeirGavish I also think it sounds better. Much more punch like the original. In my opinion, I think it is great. Great job! I'm not very good at music so I tried my best to provide some feedback. It sounds better than the default one we had before.

MeirGavish:
@GreenNinja2525 Thanks! Thank you for listening and giving feedback! Which one did you listen to? The recent one I uploaded in my comment or the one originally in this PR?

GreenNinja2525:
@MeirGavish the latest one, I listened to the other one and I honestly did not notice too much difference. This latest one definitely sounds different, much better in my opinion. I thought the chords before sounded less varied and like crunched together, the new one like sounds cleaner and less crunchy. I like it better. Awesome job 👍

MeirGavish:
Great! Thanks! I guess I will do the entire track like this.

ricfehr3:
I do agree, it sounds good. I think it sounds a bit better than the initial and first draft you had @MeirGavish . I'm going to give it a sound test on a couple of real GBAs and give it one more comparison if you can wait to merge this until then?

You can merge it too and we can revisit it.

MeirGavish:
@ricfehr3 Sure, it's still unfinished, I did it only for the first few bars and still need to update the rest of the piece so at some point it reverts to the original

MeirGavish:
Also, I got sidetracked and have been working on #200 (for which I hope to open a PR tomorrow)

MeirGavish:
> I do agree, it sounds good. I think it sounds a bit better than the initial and first draft you had @MeirGavish . I'm going to give it a sound test on a couple of real GBAs and give it one more comparison if you can wait to merge this until then?

@ricfehr3 Just make sure to listen to the zip I uploaded since the branch itself is still not up to date with these.

ricfehr3:
Just listened on my real gameboy and I think the new one sounds great. A bit brighter and punchier even on these little crappy speakers.

MeirGavish:
> Just listened on my real gameboy and I think the new one sounds great. A bit brighter and punchier even on these little crappy speakers.

Great! Thanks! I'll give this update a full go.

MeirGavish:
Okay, I did the entire piece with the new chords. Should be good now.

MeirGavish:
I see the first chord's sample rate is much higher than the rest, I want to check that before merging...

MeirGavish:
Something still sounds off in my latest version and I'm not entirely sure why... I'll check and try to fix it.

MeirGavish:
Okay, I'm starting to sort of understand things. Pitch-shifting the samples also shifts their harmonic content and thus their required sample rate. So it makes sense the C-based chords would sound better since they have a smaller pitch-shift. In my mistake I forgot to downsample the chord and it sounded better. The thing is, that even after the pitch shift it doesn't seem to have any harmonic content past 5.2kHz so theoretically this shouldn't happen. 
Anyway, I will redo the chords in a higher sample rate. I still think 44.1kHz is overkill so I'll play around with it a bit and see what rate makes sense.

MeirGavish:
Yeah, I'll just go with 44.1khz, it just sounds much better even when compared to 32khz, I'm fine with allocating more memory for music.

ricfehr3:
> Yeah, I'll just go with 44.1khz, it just sounds much better even when compared to 32khz, I'm fine with allocating more memory for music.

Yeah go for it! The ROM size is pretty small still.

MeirGavish:
There, switched all the chords to 44.1kHz, it sounds good now. The ROM grew to 4MB though. 😂😅 
The ROM is mostly music anyway, samples take a lot of memory.
Hopefully I won't have to add many more samples as I work on the music.
This time I also kept the Audacity files for all the chords so I can easily re-export them.

ricfehr3:
It sounds good! 4MB is still within reason. In my opinion, anything <16MB is acceptable. 32MB was the biggest cart, and 16MB was very common. Every emulator will handle that size, and if someone wanted to flash this onto their own cart, it would still be a reasonable size (especially considering that most bootleg carts are 16MB minimum for pokemon).

I'll let you decide when to merge. But it's sounding good.

ricfehr3 (approved):

MeirGavish:
Thanks! I rebased and will merge without squashing this time just so the samples stay in the git history.

PR #234, opened 2025-11-16
Changed "Hand" and "Discard" text in UI to "Hands" and "Discards"

ImNotGoose:


MeirGavish:
Wondering about this. You made the discards and hands boxes bigger to fit the text so they are not fully consistent and aligned with the Rounds and Ante boxes (or with each other) and the numerical text is now not fully centered within the box. But maybe it's better to have the full text?

MathisMartin31:
~Also, the backgrounds have changed but the code hasn't, have the hardcoded grit color indexes been verified? I had trouble with them when I added the score flames effect.~
Edit: I just checked the colors they're fine, but the other points still stand for me

That, Meir's remarks and no description for the PR are throwing me off a bit. I think abbreviating some words in favor of art and UI consistency is better imo

ImNotGoose:
Yeah I was thinking it looked a bit weird but I cant tell if its better to keep it the other way or this way

Issue #136, opened 2025-09-20
Describe and enforce a style guide


ricfehr3:
Since multiple people are working in this codebase, a style guide / some sort of enforcement in the CI would be helpful. There's inconsistencies in the style used throughout this project. I'm proposing a few outputs:

1. Simple style guide covering (but not limited too):

- [x] Indentation ('Allman' style is most used in this project)
- [x] **Struct** declaration 
  - [x] With/without typedef?
  - [x] Header/implementation file conventions (i.e. prefix with an underscore to illustrate the scope in implementation file)
- [x] **Enum** declaration 
  - [x] With/without typedef?
  - [x] Header/implementation file conventions
  - [x] Prefix the enum type?
- [x] **union** declaration 
  - [x] With/without typedef?
  - [x] Header/implementation file conventions
- [x] Order of declarations
  - [x] Macros
  - [x] enums
  - [x] unions
  - [x] structs
  - [x] static variable declarations (const first)
  - [x] static function declarations
  - [x] header function implementations
  - [x] static function implementations
- [x] Case styles for
  - [x] Structs
  - [x] Struct members
  - [x] Unions
  - [x] Union members
  - [x] Enum
  - [x] Enum members
  - [x] Const variables
  - [x] Static variables
  - [x] Global variables
  - [x] Static functions
  - [x] Global functions
  - [x] Arrays
  - [x] Variables

2. A clang-format file enforcing as many of these designs as possible: [Configurable example online ](https://zed0.co.uk/clang-format-configurator/)
- [ ] Implement clang-format check in CI

3. Refactored codebase to match the agreed format

I feel that introducing anything beyond this would be overkill. A static code analysis tool like `clang-tidy` would be cool, but would probably require a pretty significant refactor and tweaking it for the devkitpro environment sounds difficult already. 

As long as there's an agreed style on the above topics (we can talk if more or less is in order), common sense and touches of personal style are fine.

This is a nice to have for sure. I can volunteer to do it after my other PRs/issues.

ricfehr3:
Style guide I've been writing, I'll be adding to this comment while I add to it:

# Indentation
- Use "Allman" style
- Use 4-spaces, NO TABS

# Variables
- Local variables:
	- Use `snake_case`
	- Ex: `foo_bar_rulez`, `cool_local_var`
- Global variables:
	- Use `snake_case` 
	- Prefix with `g_` to avoid collisions
	- Ex: `g_game_state`, `g_lookup_table`
[...]

PR #240, opened 2025-11-16
Add clang format

ricfehr3:
Add a .clang-format file and apply it to the bitset files.

I plan on doing this piecewise, file by file. I will likely be adding docs for doxygen as well. Going to do this to complete #136 .

### DIFF
--- a/.github/workflows/run_tests_ci_workflow.yml
+++ b/.github/workflows/run_tests_ci_workflow.yml
@@ -17,8 +17,9 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Clang Format
-        run: clang-format --dry-run include/*.h source/*.c > /dev/null 2>&1
-        # When existing code conforms, re-add this:
+        run: clang-format --dry-run -Werror include/bitset.h source/bitset.c
+        # Currently contains only the code that has been formatted to conform
+        # When all existing code conforms, change to this:
         # run: clang-format --dry-run -Werror include/*.h source/*.c
 
       - name: Run Tests


source/bitset.c
```diff
@@ -121,8 +125,7 @@ int bitset_find_idx_of_nth_set(const Bitset *bitset, int n)
 
 BitsetItr bitset_itr_create(const Bitset* bitset)
 {
-    BitsetItr itr =
-    {
+    BitsetItr itr = {
```
ricfehr3, line 128: For some reason, only on designated initializers such as this, it refuses to apply allman. This is a known issue: https://github.com/llvm/llvm-project/issues/79054 . I think we should just deal with it for now. Everything else seems to be working fine, and there's nothing we can do 🤷

MeirGavish: Oh well, maybe some day they will resolve the issue...

source/bitset.c
```diff
@@ -10,9 +11,9 @@ void bitset_set_idx(Bitset *bitset, int idx, bool on)
     // These are more efficient, but removed for readability
     // See: https://github.com/cellos51/balatro-gba/pull/132#discussion_r2365966071
     // Divide by 32 to get the word index
-    //uint32_t i = idx >> 5;
+    // uint32_t i = idx >> 5;
```
MeirGavish, line 14: So clang also checks commented out code?

.clang-format
```diff
@@ -0,0 +1,42 @@
...
+
+# Include sorting
+SortIncludes: CaseSensitive
+IncludeBlocks: Regroup
+
+ContinuationIndentWidth: 4
+AlignArrayOfStructures: Left
+PackConstructorInitializers: Never
+
+SpaceBeforeParens: Never
```
MeirGavish: I actually like a space after keyword statements but not after functions...

```
for (int j = 0; j < bitset->nbits; j++)
{
     if (base == n)
     {
...
```

```
bool bitset_get_idx(Bitset* bitset, int idx);

...

    tracker += __builtin_popcount(bitset->w[i]);
```

.clang-format
```diff
@@ -0,0 +1,42 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
+UseTab: Never
+
+# Allman brace style
+BreakBeforeBraces: Allman
+AllowShortBlocksOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortLoopsOnASingleLine: true
```
MeirGavish: Why allow single-line loops but not ifs? Just wondering

.clang-format
```diff
@@ -0,0 +1,42 @@
...
+# Include sorting
+SortIncludes: CaseSensitive
+IncludeBlocks: Regroup
+
+ContinuationIndentWidth: 4
+AlignArrayOfStructures: Left
+PackConstructorInitializers: Never
+
+SpaceBeforeParens: Never
+AlignOperands: AlignAfterOperator
```
MeirGavish, line 34: So how would we break this?

```
int this_is_a_number_with_a_very_very_long_name = and_this_is_another_number_with_a_very_long_name + and_here_is_another_number_with_a_very_long_name
```
My intuition says something like this,
```
int this_is_a_number_with_a_very_very_long_name = 
    and_this_is_another_number_with_a_very_long_name + 
    and_here_is_another_number_with_a_very_long_name
```

Not sure if this is a very likely example but I'm using an extreme examples get the idea. It doesn't have to enforce this, just allow it, so I'm just making sure it's allowed.

.clang-format
```diff
@@ -0,0 +1,42 @@
...
+BreakBeforeBinaryOperators: None
+
+IndentPPDirectives: AfterHash
+AlignConsecutiveMacros: Consecutive
+
+Cpp11BracedListStyle: false
```
MeirGavish: Do we need this? If so why? It doesn't seem relevant since we have no C++

MeirGavish (changes requested):
Looks good! Thanks! Just a few questions and one request for SpaceBeforeParens

Also, how will this work? Will it automatically run as a test on PRs? Can it be run manually? Can we take a file and use this to re-format it automatically?

source/bitset.c
ricfehr3: It does, but it only is adding a space with `SpacesBeforeTrailingComments`. I like it though, what do you think?

.clang-format
ricfehr3: Got it!

.clang-format
ricfehr3: I felt that return statements are often in those ifs. Making it more obvious it's returning. I'm going to vote for newlines in everything though, just to be consistent. What do you think?

.clang-format
ricfehr3: In our case that particular one would look like:

```c


int this_is_a_number_with_a_very_very_long_name = 
    very_long_name + another_very_long_name
```

That is, on one line below. That's just due to the 120 character limit. If the function names were longer it would look similar to what you have:

```c
int this_is_a_number_with_a_very_very_long_name = 
    wow_what_a_long_maybe_even_longer_name + 
    very_long_name
```

But to answer your question, yes that is allowed.

.clang-format
ricfehr3: Nah, I'll remove it. It was a last ditch effort to fix that bug I mentioned with the github issue.

ricfehr3:
> Looks good! Thanks! Just a few questions and one request for SpaceBeforeParens
> 
> Also, how will this work? Will it automatically run as a test on PRs? Can it be run manually? Can we take a file and use this to re-format it automatically?

Thanks for reviewing this Meir. I left a couple questions:

- **Do we want to use `SpacesBeforeTrailingComments`**
- **Do we want a newline for all single line loops and ifs? Or same line (I vote newline)**

---

Answering your question: _Also, how will this work? Will it automatically run as a test on PRs? Can it be run manually? Can we take a file and use this to re-format it automatically_

Yes, there's a few things we can do with this, We have time to decide as I'll need to piecewise do all the other files before running any checks. But the workflow I would like to use is rejecting PRs that fail the checks and leaving the responsibility to the developer the fix their own code. The only reason I want that is I don't really trust any automatic occurring code change. This shouldn't be too hard because most every IDE picks up on .clang-format files and will automatically correct or warn the developer. 

To run this check for our ci to use and developers if they want, I'd like to add a new Makefile target that runs something like:

```sh
clang-format --dry-run -Werror include/bitset.h
```
Example of bad output:
```sh
include/bitset.h:171:20: error: code should be clang-formatted [-Wclang-format-violations]
int bitset_itr_next (BitsetItr* itr);
                   ^
```

.clang-format
MeirGavish: Yeah, I guess consistency is better, I usually prefer them in another line even if it means 4 whole lines for something like

```
for (int i = 0; i < len; i++)
{
     arr[i]++;
}
```

.clang-format
MeirGavish: Cool, thanks!

source/bitset.c
MeirGavish: Not quite sure I understand, from the documentation `SpacesBeforeTrailingComments` seems to control the minimum number of spaces before a comment on the same line as code. If you want to set it, I'm fine with it, maybe set it to 4 for a single tab.
How does clang affect this particular example where the code is commented out? 
Will it enforce other rules on commented out code? For example will it give an error for code like this for the pointer alignment?
```
// int *p_num = NULL;
```

ricfehr3: That was my bad, I misunderstood the docs. The correct option is:

```
ReflowComments: true
SpacesInLineCommentPrefix:
  Minimum: 1
  Maximum: 1
```

This makes all comments after `//` have 1 space exactly. I applied it here.

ricfehr3:
> Approving. The only thing is maybe you can already add the workflow to run the check just for the files you already formatted to make sure they stay that way? I like the idea of having a workflow that will fail for PRs that don't follow the format and I agree, it shouldn't automatically update code without the developer knowing. I just want to be able to manually say "take this file or chunk of code, and change the formatting to be as defined" so we don't have to manually add newlines and spaces etc. I bet this can be done with IDE integration, I can check how to configure it later.

Yeah, most IDE's should hook into clang-format. Mine has a command to run it on the current file I'm editing, I know some have autocorrection too.

As for the CI. I went ahead and added the check in our tests action, but I held off on it producing any error. I want to wait to turn it on once the rest of the codebase is up to date, otherwise it's a big ol' log of failures. 

I'm going to merge this one soon and get started on the next one. You can too if you get to it first @MeirGavish , don't want to modify the actions without your approval, but this one is pretty small.

.github/workflows/run_tests_ci_workflow.yml
```diff
@@ -9,9 +9,18 @@ jobs:
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang-format
       - name: Checkout repository
         uses: actions/checkout@v5
 
+      - name: Clang Format
+        run: clang-format --dry-run include/*.h source/*.c > /dev/null 2>&1
+        # When existing code conforms, re-add this:
+        # run: clang-format --dry-run -Werror include/*.h source/*.c
```
MeirGavish:
```suggestion
        run: clang-format --dry-run -Werror include/bitset.h source/bitset.c
        # Currently contains only the code that has been formatted to conform
        # When all existing code conforms, change to this:
        # run: clang-format --dry-run -Werror include/*.h source/*.c
```

Why not like this? That way we can add every file when we format it and make sure it stays that way, right?

ricfehr3: That's a good idea.

Issue #225, opened 2025-11-13
[BUG] Blueprint/Brainstorm and retrigger Jokers do not behave correctly


MathisMartin31:
**Describe the bug**
I tested it with Blueprint+Hanging Chad but I suspect this issue affects most of copying and retrigger Jokers.
When Blueprint copies Hanging Chad, it will make the card score again, but to do so it will call Hanging Chad's JokerEffect function, which means its internal counter will decrease and when it's its turn to be scored, it will have no retriggers left and will have no effect.

**To Reproduce**
Give yourself or buy Blueprint and Hanging Chad so that the former copies the latter.

**Expected behavior**
Blueprint should make the first scored card score again twice, then Hanging Chad should also make it score again twice for a total of 5 times.

**Screenshots**
Encountered on my 3DS during lunch break, will post a video after work.

**What version**
main branch

**What Emulator or Platform were you playing on**
3DS Virtual Console, injected with NSUI

MeirGavish:
Oh dear. That's a big problem I missed with adding Joker state - Blueprint and Brainstorm update the copied Joker's state.

MathisMartin31:
The fix would be to get the `joker_effect` of the copied Joker, and to apply it with the `joker_object` of the Blueprint/Brainstorm. This way the copying Joker would completely behave like the Joker it copies, but will update its own data

MeirGavish:
That would work for retriggers but not for jokers that have a longer lasting state because then they would completely copy the behavior including having their own state and not just the score. The only joker we currently have in the code that has this is Seltzer and it would mean Brainstorm and Blueprint could expire like Seltzer. 
It also means that if we were to implement jokers that have an accumulative state such as Ceremonial Dagger for example, the copying joker would not necessarily have the same score because it has a different state.

MeirGavish:
We can still do it like that, because there are no such jokers actually in the game now, we just need to be aware that if we ever do want to add them it would be a problem.

MathisMartin31:
Oh you're right, that's gonna be an issue. If we come up with an idea to address this we should still do the more complicated change to avoid doing the work twice, but if not we could do this.

~~What we could do is when scoring Blueprint/Brainstorm, copy the state of the copied Joker, apply the effect to its own data that is now identical to the copied Joker, and filter the resulting effect to take out any expiration effect.~~

[...]

PR #243, opened 2025-11-16
Implement the Spare Trousers Joker

MathisMartin31:
Implementing Spare Trousers, and adding the possibility for Jokers to display effect text when triggering the `JOKER_EVENT_ON_HAND_PLAYED` event. I tested the Joker and the scaling seems to be working. 

@NoWhammies10 I took the liberty of tweaking a little bit the art you made for the Joker, tell me if this version looks good to you or if you like the original better.

This PR depends on #213 to use the optimized checks for existing hands it implements, so it will stay as a draft until the former is merged and I will rebase on main when it is. The relevant commits to this PR are the last two. 

Also @MeirGavish, I'm now seeing that when we score Cards and Jokers, we go through each state and there is a delay between each one. This is normal, but these delays add up when several states in a row don't need to display anything, leading to long pauses that feel a bit weird. Probably outside the scope of this PR, but we should think about it


MeirGavish:
Wait, we reached 50 jokers, we said we won't go past that... 😅
It's not like I don't want it in but we need to put a stop somewhere to not implement a full competing game...
Also, adding more jokers that rely on state would make interactions with Blueprint and Brainstorm more complicated as pointed out in issue #225 and we may need to tackle that to not introduce more bugs.

MathisMartin31:
> Wait, we reached 50 jokers, we said we won't go past that... 😅

Ooooooh shit I forgot about that, sorry ><

> Also, adding more jokers that rely on state would make interactions with Blueprint and Brainstorm more complicated as pointed out in issue https://github.com/cellos51/balatro-gba/issues/225 and we may need to tackle that to not introduce more bugs.

Yeah maybe I should focus on this first

MeirGavish:
This makes me sad...

MathisMartin31:
Don't be, it's okay. I haven't deleted the branch on my end, so it's ready to go if we have another discussion regarding scope in the future :)

ricfehr3:
:cry: rough

PR #244, opened 2025-11-16
Modify blueprint and brainstorm implementation

MathisMartin31:
This PR aims to fix issue #225 

https://github.com/user-attachments/assets/ff401dda-3207-4b38-8d69-3e7d03822be9

I tested a bunch of other Jokers and they all worked so far so I'll make this a proper PR instead of a draft. Misprint scared me for a hot second where it didn't activate sometimes until I remembered it could do +0 Mult, thus technically having no effect xD

Only thing is that what we talked about with @MeirGavish about copying the target data for scaling Jokers. I didn't do it since we don't have these kind of Jokers yet. We will have Seltzer but it's not in the game yet.

### DIFF
--- a/include/joker.h
+++ b/include/joker.h
@@ -81,8 +81,8 @@ typedef struct
     u8 rarity;
 
     // General purpose values that are interpreted differently for each Joker (scaling, last retriggered card, etc...)
-    u32 scoring_state;
-    u32 persistent_state;
+    s32 scoring_state;
+    s32 persistent_state;
 } Joker;
 
 typedef struct JokerObject
--- a/source/joker_effects.c
+++ b/source/joker_effects.c
@@ -405,7 +405,7 @@ static JokerEffect raised_fist_joker_effect(Joker *joker, Card *scored_card, enu
 {
     JokerEffect effect = {0};
 
-    u32* p_lowest_value_index = &(joker->scoring_state);
+    s32* p_lowest_value_index = &(joker->scoring_state);
 
     switch (joker_event)
     {
@@ -596,7 +596,7 @@ static JokerEffect acrobat_joker_effect(Joker *joker, Card *scored_card, enum Jo
 static JokerEffect hanging_chad_joker_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
 {
     JokerEffect effect = {0};
-    u32* p_remaining_retriggers = &(joker->scoring_state);
+    s32* p_remaining_retriggers = &(joker->scoring_state);
 
     switch (joker_event)
     {
@@ -756,7 +756,7 @@ __attribute__((unused))
 static JokerEffect photograph_joker_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
 {
     JokerEffect effect = {0};
-    u32* p_first_face_index = &(joker->scoring_state);
+    s32* p_first_face_index = &(joker->scoring_state);
 
     switch (joker_event)
     {
@@ -808,11 +808,12 @@ static JokerEffect triboulet_joker_effect(Joker *joker, Card *scored_card, enum 
 static JokerEffect dusk_joker_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
 {
     JokerEffect effect = {0};
-    u32* p_last_retriggered_index = &(joker->scoring_state);
+    s32* p_last_retriggered_index = &(joker->scoring_state);
 
     switch (joker_event)
     {
         case JOKER_EVENT_ON_HAND_PLAYED:
+            // start at -1 so that a first index of 0 can satisfy the retrigger condition below
             *p_last_retriggered_index = UNDEFINED;
             break;
         
@@ -820,10 +821,7 @@ static JokerEffect dusk_joker_effect(Joker *joker, Card *scored_card, enum Joker
             // Only retrigger current card if it's strictly after the last one we retriggered
             if (get_num_hands_remaining() == 0)
             {
-                // need the second check because scoring_state is unsigned, so
-                // it starts as UINT32_MAX which is bigger than the scored index,
-                // but we do need to retrigger the card in that case
-                effect.retrigger = (*p_last_retriggered_index < get_scored_card_index() || *p_last_retriggered_index == UNDEFINED);
+                effect.retrigger = (*p_last_retriggered_index < get_scored_card_index());
                 if (effect.retrigger)
                 {
                     *p_last_retriggered_index = get_scored_card_index();
@@ -923,7 +921,7 @@ static JokerEffect blueprint_brainstorm_joker_effect(Joker *joker, Card *scored_
 static JokerEffect hack_joker_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
 {
     JokerEffect effect = {0};
-    u32* p_last_retriggered_index = &(joker->scoring_state);
+    s32* p_last_retriggered_index = &(joker->scoring_state);
 
     switch (joker_event)
     {
@@ -939,7 +937,7 @@ static JokerEffect hack_joker_effect(Joker *joker, Card *scored_card, enum Joker
                 case THREE:
                 case FOUR:
                 case FIVE:
-                    effect.retrigger = (*p_last_retriggered_index < get_scored_card_index() || *p_last_retriggered_index == UNDEFINED);
+                    effect.retrigger = (*p_last_retriggered_index < get_scored_card_index());
                     if (effect.retrigger)
                     {
                         *p_last_retriggered_index = get_scored_card_index();
@@ -962,8 +960,8 @@ __attribute__((unused))
 static JokerEffect seltzer_joker_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
 {
     JokerEffect effect = {0};
-    u32* p_last_retriggered_idx = &(joker->scoring_state);
-    u32* p_hands_left_until_exp = &(joker->persistent_state);
+    s32* p_last_retriggered_idx = &(joker->scoring_state);
+    s32* p_hands_left_until_exp = &(joker->persistent_state);
 
     switch (joker_event)
     {
@@ -976,10 +974,10 @@ static JokerEffect seltzer_joker_effect(Joker *joker, Card *scored_card, enum Jo
             break;
         
         case JOKER_EVENT_ON_CARD_SCORED_END:
-            // Works the same way as Dusk, but if it can still trigger
+            // Works the same way as Dusk, but checks if it can still trigger
             if (*p_hands_left_until_exp > 0)
             {
-                effect.retrigger = (*p_last_retriggered_idx < get_scored_card_index() || *p_last_retriggered_idx == UNDEFINED);
+                effect.retrigger = (*p_last_retriggered_idx < get_scored_card_index());
                 if (effect.retrigger)
                 {
                     *p_last_retriggered_idx = get_scored_card_index();
@@ -1012,7 +1010,7 @@ static JokerEffect seltzer_joker_effect(Joker *joker, Card *scored_card, enum Jo
 static JokerEffect sock_and_buskin_joker_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
 {
     JokerEffect effect = {0};
-    u32* p_last_retriggered_face_index = &(joker->scoring_state);
+    s32* p_last_retriggered_face_index = &(joker->scoring_state);
 
     switch (joker_event)
     {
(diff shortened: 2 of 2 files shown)


include/joker.h
```diff
@@ -69,7 +69,9 @@ enum JokerEvent
 #define JOKER_STENCIL_ID 16
 #define PAREIDOLIA_JOKER_ID 30
 #define JOKER_BRAINSTORM_ID 40
```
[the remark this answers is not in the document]
MeirGavish: I guess `JOKER_STENCIL_ID` should be also changed to `STENCIL_JOKER_ID` to be consistent with the other definitions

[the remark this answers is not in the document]
MathisMartin31: You're right I'll do a pass on the Joker IDs

source/joker_effects.c
```diff
@@ -838,47 +838,73 @@ static JokerEffect dusk_joker_effect(Joker *joker, Card *scored_card, enum Joker
...
-    }
-
-    List* jokers = get_jokers_list();
-    JokerObject* first_joker = list_get_at_idx(jokers, 0);
-
-    if (first_joker != NULL && first_joker->joker->id != JOKER_BRAINSTORM_ID)
+    // find the copied Joker, may need to bounce around Blueprints and a Brainstorm
+    // If we encounter NULL, we have a Blueprint at the end of the list that can't copy anything.
+    // If we go through the starting Joker again, we are in a loop and need to exit
+    JokerObject* copied_joker_object = starting_joker_object;
```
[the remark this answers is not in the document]
MathisMartin31: The Jokers list will never be modified during a Joker effect function but I guess it doesn't hurt to add a check for extra safety

MathisMartin31:
@MeirGavish Because except for how to jump to the copied Joker (one to the right or the leftmost one), all the logic will be the same. Splitting it into two functions would have duplicated something like 3/4ths of the code.

I haven't implemented the copy of scaling Jokers data since we don't have any to test it yet, but both Jokers are going to behave the same there too

source/joker_effects.c
```diff
@@ -838,47 +838,80 @@ static JokerEffect dusk_joker_effect(Joker *joker, Card *scored_card, enum Joker
...
+                // make also sure we don't expire
+                effect.expire = false;
+
+                // exit the loop
+                copied_joker_object = NULL;
+
+                break;
+        }
     }
+    while(copied_joker_object != NULL && copied_joker_object != starting_joker_object);
```
MeirGavish: I think this enters an infinite loop for
Blueprint, Brainstorm, Brainstorm
When activating the last Brainstorm, it loops between the left Blueprint and Brainstorm without ever returning to the original Brainstorm.

source/joker_effects.c
```diff
@@ -838,47 +838,80 @@ static JokerEffect dusk_joker_effect(Joker *joker, Card *scored_card, enum Joker
...
-    List* jokers = get_jokers_list();
 
-    ListItr itr = list_itr_create(jokers);
-    JokerObject* curr_joker_object;
+    // No need for this kind of init since these Jokers
+    // will have their data copied when needed
+    if (joker_event == JOKER_EVENT_ON_JOKER_CREATED)
+    {
+        return effect;
+    }
```
MeirGavish: There may be other events we want to avoid. In general we want to make sure it's a scoring event rather than a notifying event but I don't think we can currently differentiate them well.

source/joker_effects.c
```diff
@@ -838,47 +838,80 @@ static JokerEffect dusk_joker_effect(Joker *joker, Card *scored_card, enum Joker
...
+            // We encountered a Joker that isn't a Copying Joker and copy it now
+            // but how we copy it depends on this Joker's ID because they don't
+            // all handle data the same way.
+            default:
+                u8 copied_joker_id = copied_joker_object->joker->id;
+                const JokerInfo* copied_joker_info = get_joker_registry_entry(copied_joker_id);
+
+                // For Jokers that retain data across rounds,
+                // copy this data to the copying JokerObject
+                // if (copied_joker_id == SELTZER_JOKER_ID) {...}
```
MeirGavish: Not very fond of this, it means that every joker that's added may need to update Blueprint and Brainstorm. I think the correct thing to do is to have 2 different variables for the persistent data and the scoring data. The persistent data would always be copied and the scoring data will have its own instance for the copying joker. This might mean changing the `joker_effect` function which could be a big refactor. I did say it's okay to do it like this for now since there are no such jokers currently so I'm fine with just replacing this comment with a TODO with this suggestion. I guess it actually needs to be at the joker struct definition rather than here but there needs to be some note here too.

source/joker_effects.c
MeirGavish: There are other events we may want to avoid. I'm wondering if it might be better to define the events that copying jokers do activate on rather than the ones that don't. We want all the scoring events, but not ON_CREATED and ON_ROUND_END etc.

MeirGavish (changes requested):
Okay, I was also wondering why you made the other changes such as make the traversal iterative rather than recursive. I guess making it recursive would mess up the joker state by affecting all the traversed jokers?
There are a few design considerations here, I added my comments. Also now that Blueprint and Brainstorm are merged, there's a copying loop that's missed.

source/joker_effects.c
MathisMartin31: Ah you mean if we have two brainstorms? I didn't think of that, I'll take a look. I think the easiest fix is check if we go twice through a brainstorm, as we did before

source/joker_effects.c
MathisMartin31: I think we do want to copy JOKER_EVENT_ON_ROUND_END, there is at least Rocket that's a scaling Joker scoring on this event.

The other thing is if we take Ceremonial Dagger, it scores on Blind Selected by destroying a Joker. Blueprint and Brainstorm don't destroy Jokers there, but they copy other Jokers that score on Blind Selecte like Riff Raff or Marble Joker.

What I'm getting at is I think we do need to have per-Joker logic in Brainstorm and Blueprint if we want to avoid copying non compatible ones

source/joker_effects.c
MathisMartin31: If we go the two data route, I think keeping the data struct with two halves as the state and having only a u32 for the scaling would be enough. Quite a few non scaling Jokers need two values to keep track of, but I haven't seen any that needs more than one to keep track of the scaling

```
typedef struct 
{
    ...
    union
    {
        s32 state;
        struct {
            s16 state0;
            s16 state1;
        } halves;
    };
    u32 scaling;
} Joker;
```

source/joker_effects.c
MathisMartin31: Should be good now: here only one blueprint actually copies anything, since the other one, and the two brainstorms lead to an infinite loop that we catch and break out of ^^

https://github.com/user-attachments/assets/6e0dba90-0ee2-4444-bf45-e82748c47116

source/joker_effects.c
MeirGavish: Actually I think the right thing to do here is break up the union and have two fields - `s32 scoring_state` and maybe `s32 persistent_state`. Come to think of it, I think it's not as big of a refactor as I thought.

source/joker_effects.c
MeirGavish: From the Balatro wiki [Blueprint page](https://balatrogame.fandom.com/wiki/Blueprint) (same goes for Brainstorm):

_In general, Blueprint will not copy the following effects:_

_- Anything that triggers at the end of a round (ex.  [Golden Joker](https://balatrogame.fandom.com/wiki/Golden_Joker))_

And The Rocket is listed in the incompatible jokers. I think the only problematic joker is Ceremonial Dagger and I guess we would have to add specific logic for that but let's avoid as much per-joker logic as possible.

source/joker_effects.c
MathisMartin31: I did that in my latest commit. Everything seems to be working fine, although I still can't test the `persistent_state` since we have no real scaling Joker in the game. At least not copying the `scoring_data` works.

And yes the refactor wasn't that big of a deal ^^

source/joker_effects.c
MathisMartin31: I completely missed that, I just thought some Jokers were compatible and some weren't. I added the check.

There will be work to do to exclude incompatible Jokers but:
- as you said, the ones triggering at the end of the round are all excluded
- Jokers with passive effects can't be copied, but since their effect is passive their `joker_effect_func` is `joker_effect_noop`, which doesn't do anything
- There are only 5 problematic Jokers left:
    - Sixth Sense, Trading Card, Invisible Joker that are incompatible but don't fall into the above categories
    - Ceremonial Dagger and Madness that are both compatible, but we don't want the copying Joker to destroy a card by triggering before the copied Joker

Other than that Blueprint and Brainstorm should be pretty straightforward

include/joker.h
```diff
@@ -78,15 +79,10 @@ typedef struct
...
-    union
-    {
-        s32 data;
-        struct {
-            s16 data0;
-            s16 data1;
-        } halves;
-    };
+    u32 scoring_state;
+    u32 persistent_state;
```
MeirGavish: Why make it unsigned? It's much better to make it signed.

source/joker_effects.c
```diff
@@ -820,7 +820,10 @@ static JokerEffect dusk_joker_effect(Joker *joker, Card *scored_card, enum Joker
             // Only retrigger current card if it's strictly after the last one we retriggered
             if (get_num_hands_remaining() == 0)
             {
-                effect.retrigger = (*p_last_retriggered_index < get_scored_card_index());
+                // need the second check because scoring_state is unsigned, so
+                // it starts as UINT32_MAX which is bigger than the scored index,
+                // but we do need to retrigger the card in that case
+                effect.retrigger = (*p_last_retriggered_index < get_scored_card_index() || *p_last_retriggered_index == UNDEFINED);
```
MeirGavish: This isn't very good, it's a side-effect of making it unsigned - UINT32_MAX = UNDEFINED for u32. UNDEFINED shouldn't be used with an unsigned variable and is one of the reasons I mislike using unsigned variables.

MeirGavish: It might be good to refactor and not rely on UNDEFINED < 0 but it needs to be done differently and specifically check for UNDEFINED before updating *p_last_retriggered_index but that should probably be done in a different PR.

source/joker_effects.c
MeirGavish: Good job looking into it! ~~I guess DNA could be problematic too.~~ Actually DNA is compatible. Invisible Joker is an ON_JOKER_SOLD question which includes Luchador and Diet Cola.

include/joker.h
MathisMartin31: The idea was the we rarely need negative values, and never do when the data stored is a score. Using a u32 instead of a s32 gives us twice the value range at the cost a small check against UNDEFINED but I get the argument that it's a bit more messy

source/joker_effects.c
MathisMartin31: Oh yeah I completely skipped the "On Sold" event in my mind lol.

Luchador is compatible but copying it won't do much so we could handle it whichever way is easier for us, Diet Cola is compatible and Invisible isn't.

All in all, we could allow the copy except for Invisible, or prevent the copy except for Diet Cola (since triggering Luchador doesn't do anything)

source/joker_effects.c
MathisMartin31: I could change `UNDEFINED` for `UINT32_MAX`? This does the same thing except it's maybe more self evident that we are checking for `UINT32_MAX` because we are using unsigned.

Or I could just bite the bullet and use signed integers, because I'm making such a big deal of the extra scaling room but no Joker has ever scaled up to 2^32 - 1 outside of the Cryptid-tier bullshit ><. Hell, the largest scaling Jokers I've seen are Wee Jokers and I've never seen them go past 6k

source/joker_effects.c
MeirGavish: Yeah, `persistent_state` can be tested if/when new jokers that use it are added.

include/joker.h
MeirGavish: This is a generic field that can be used by many jokers for many different purposes. It is also subtracted for Seltzer, Ice Cream, etc which introduces the danger of underflow. It isn't like score that we know for certain shouldn't be negative, we know it could grow to very large numbers so there's a benefit to making it unsigned, and we know it isn't subtracted and don't expect more subtractions to be introduced. The use of `UNDEFINED` already makes it unfit to be unsigned.

source/joker_effects.c
MeirGavish: If anything, it's better to check for `UNDEFINED` which is the actual assigned value - `UINT32_MAX` is the side-effect value.
The numbers are not expected to grow so large and it's much better to use that extra bit to be able to detect negative values.

MathisMartin31: I made them signed again, and reverted the checks to how they were before the change. I tested a few and it looks like it's working

Issue #184, opened 2025-10-27
Fix usage of `memcmp`


ricfehr3:
This is a continuation of the conversation here: https://github.com/cellos51/balatro-gba/pull/160#discussion_r2464115667

@MeirGavish @MathisMartin31 fyi

(Moving to an issue because it shouldn't be in the PR)

*Context: chunk from PR*

```c
    JokerEffect joker_effect = joker_get_score_effect(joker_object->joker, scored_card);

    if (memcmp(&joker_effect, &(JokerEffect){0}, sizeof(JokerEffect)) != 0)
```

---

I'm looking at this again, and maybe I was just tired. But I didn't understand what was even going on here originally, haha. 

So, here's what I'm seeing now:
- `&(JokerEffect){0}` <- compound literal to give a temporary pointer of an all zero `JokerEffect`
- So.. the `memcmp` is looking to see if the joker_effect is all 0's (default)
- The only reason it would be all 0's is when the function `joker_get_score_effect` doesn't have a joker entry or the joker doesn't have an effect function.

So I say, change the `joker_get_score_effect` function to return if the operation completed or not and to take a pointer to a `JokerEffect` to fill:

```c
bool joker_get_score_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event, JokerEffect *ret_effect)
{
    const JokerInfo *jinfo = get_joker_registry_entry(joker->id);
    if (!jinfo || !jinfo->joker_effect || !ret_effect) return false;

    *ret_effect = jinfo->joker_effect(joker, scored_card, joker_event);
    return true;
}
```

then adjust the call to it accordingly:

```c
JokerEffect joker_effect;
if(!joker_get_score_effect(joker_object->joker, scored_card, joker_event, &joker_effect)) return false
```

MeirGavish:
Well, technically the joker effect can still be all 0's if it does have an effect but it just didn't trigger. For example one of the sinful jokers didn't trigger because the scored hand was not of the appropriate suit, so it will just not change the mult and return all 0s.
This means that the bool return value needs to propagate to the joker effect functions so they return false in that case too.
Otherwise `joker_get_score_effect()` would not return false and while no scoring will be displayed, `joker_object_shake()` would still be called.

That said I'm all for this, I'm also not too fond of the initialization in every joker effect function `JokerEffect effect = {0};`, it's an unnecessary performance overhead, I think we may be able to avoid it entirely with this return value.

MathisMartin31:
[...]

PR #245, opened 2025-11-19
Refactor JokerEffect scoring

MathisMartin31:
@MeirGavish @ricfehr3 

This is my proposed solution to issue #184

With this I do several things:

- The `joker_effect` struct is passed as argument as a NULL `JokerEffect**` to the Joker effect functions. This allows it to be allocated only if the Joker actually has an effect
- Allocation of this `JokerEffect` struct is now done by leveraging the memory pool system to avoid calling `malloc`
- We no longer return the `JokerEffect` by value. Instead, we now return an aggregation of flags that can easily be checked against by the `joker_object_score` function to only access fields that will have been initialized with useful values

This PR will stay as a draft for now to gather your opinions, and update as other important PRs get merged because this one has a lot of conflicts with #212, #213 and #244

I tested some Jokers, but not all since this PR will be subject to a lot of changes. The ones I tested (around 10 on them) did work though :)

### DIFF
--- a/source/joker.c
+++ b/source/joker.c
@@ -340,7 +340,7 @@ bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum
     {
         set_and_shift_text(joker_effect->message, &cursorPosX, &cursorPosY, TTE_WHITE_PB);
     }
-    if (effect_flags_ret & JOKER_EFFECT_FLAG_EXPIRE && joker_effect.expire)
+    if (effect_flags_ret & JOKER_EFFECT_FLAG_EXPIRE && joker_effect->expire)
     {
         // TODO make Jokers expire
     }


MathisMartin31:
Quick remark here: I allocated enough room for `MAX_ACTIVE_JOKERS` (8) different `JokerEffect` structs in the memory pool. I think only one should be all that's needed since only one Joker can be scoring at the same time, but I wasn't sure so I went ahead and did that. Let me know if we should try and allocate only one or if it doesn't matter that much

source/joker.c
```diff
@@ -159,12 +160,12 @@ void joker_destroy(Joker **joker)
...
 
-JokerEffect joker_get_score_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
+u8 joker_get_score_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event, JokerEffect **joker_effect)
 {
     const JokerInfo *jinfo = get_joker_registry_entry(joker->id);
-    if (!jinfo) return (JokerEffect){0};
+    if (!jinfo) return JOKER_EFFECT_NONE;
 
-    return jinfo->joker_effect(joker, scored_card, joker_event);
+    return jinfo->joker_effect(joker, scored_card, joker_event, joker_effect);
```
ricfehr3: I think you should maybe change the `joker_effect` function to `joker_effect_function` or `callback` or something. It's hard to understanding how a variable `joker_effect` is passed into a function `joker_effect`.

source/joker_effects.c
```diff
@@ -7,73 +7,76 @@
 #include <stdlib.h>
 
 
-#define SCORE_ON_EVENT_ONLY_WITH_CARD(scored_card, restricted_event, checked_event, effect) \
+#define MALLOC_JOKER_EFFECT(joker_effect) *joker_effect = POOL_GET(JokerEffect);
```
ricfehr3: I don't see any calls to `POOL_FREE` here, but I also was reading this and looking at your comment and I think this could just be one variable that is stored in this c file and reset at the beginning of scoring. I don't see why we would need more than one.

MathisMartin31: I changed it to a shared instance and it works fine ^^
I used malloc though, because we only need to do it once. I don't really know where to put the free, but since it's only free when quitting the game, which means powering off the console, I wasn't too worried

MathisMartin31:
@ricfehr3 I actually missed that I wasn't retriggering cards anymore, should be fixed now

source/joker_effects.c
```diff
@@ -775,39 +864,45 @@ static JokerEffect photograph_joker_effect(Joker *joker, Card *scored_card, enum
...
+    *joker_effect = shared_joker_effect;
 
     switch (scored_card->rank)
     {
-        case KING: case QUEEN:
-            effect.xmult = 2;
+        case QUEEN:
+        case KING:
+            shared_joker_effect->xmult = 2;
+            effect_ret = JOKER_EFFECT_XMULT;
```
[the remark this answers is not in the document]
MeirGavish: This was already like that in Triboulet, but honestly I'm thinking of removing Triboulet from the code since we have enough jokers as it is and I'd rather not keep maintaining code for an inactive joker. Also I'd rather not change how Balatro works just so we can have a Legendary joker in the game so I don't plan to activate it any time soon.
This should be done in a different designated PR so we can easily find this removal in the git history (opening an [issue](https://github.com/cellos51/balatro-gba/issues/263#issue-3667875255) for this).

source/joker_effects.c
```diff
@@ -1112,6 +1221,14 @@ static const size_t joker_registry_size = NUM_ELEM_IN_ARR(joker_registry);
 
 const JokerInfo* get_joker_registry_entry(int joker_id)
 {
+    // Need to init the shared JokerEffect struct only once
+    static bool effect_init = false;
+    if (!effect_init)
+    {
+        effect_init = true;
+        shared_joker_effect = (JokerEffect*)malloc(sizeof(JokerEffect));
```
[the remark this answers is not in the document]
MeirGavish: Use option 1, you can avoid the malloc altogether by defining
```
static JokerEffect shared_joker_effect = {0};
```
and then replace all the instances of 
```
*joker_effect = shared_joker_effect;
```
with
```
*joker_effect = &shared_joker_effect;
```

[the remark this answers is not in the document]
MeirGavish: This is okay, there's no need for allocations.

source/joker_effects.c
```diff
@@ -7,92 +7,98 @@
...
 }
-#define SCORE_ON_EVENT_ONLY(restricted_event, checked_event, effect) \
+#define SCORE_ON_EVENT_ONLY(restricted_event, checked_event) \
 if (checked_event != restricted_event) \
 { \
-    return effect; \
+    return JOKER_EFFECT_NONE; \
 }
 
+static JokerEffect *shared_joker_effect = NULL;
```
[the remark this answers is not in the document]
MeirGavish: I think since we use the flags, it's safe to not do this and save up on some performance. However, if ever this causes any bugs I think it would be better to do this.

include/joker.h
```diff
@@ -61,6 +61,20 @@ enum JokerEvent
...
+{
+    JOKER_EFFECT_NONE      = 0,
+    JOKER_EFFECT_CHIPS     = 1,  // 1 << 0
+    JOKER_EFFECT_MULT      = 2,  // 1 << 1
+    JOKER_EFFECT_XMULT     = 4,  // 1 << 2
+    JOKER_EFFECT_MONEY     = 8,  // 1 << 3
+    JOKER_EFFECT_RETRIGGER = 16, // 1 << 4
+    JOKER_EFFECT_EXPIRE    = 32, // 1 << 5
+    JOKER_EFFECT_MESSAGE   = 64  // 1 << 6
+};
```
MeirGavish: Why can't you use the explicit shifts? Because it's an enum? I guess these can be a defines anyway, the enum type is never used except for in this definition.

Also since we have a joker effect struct, a joker effect function, and now a joker effect flag, explicitly add `_FLAG_` to all to these names to make it clear what they define. e.g. `JOKER_EFFECT_FLAG_NONE`, `JOKER_EFFECT_FLAG_NONE` etc.

include/joker.h
```diff
@@ -102,11 +116,11 @@ typedef struct  // These jokers are triggered after the played hand has finished
     char* message; // Used to send custom messages e.g. "Extinct" or "-1" (Bananas and food Jokers)
 } JokerEffect;
 
-typedef JokerEffect (*JokerEffectFunc)(Joker *joker, Card *scored_card, enum JokerEvent joker_event);
+typedef u8 (*JokerEffectFunc)(Joker *joker, Card *scored_card, enum JokerEvent joker_event, JokerEffect **joker_effect);
```
MeirGavish: Change the return type to u32, you are already using 7 out of 8 bits for the flags, it leaves you room to add only one more flag before you need to change it again, it's better to just give it a good number of bits from the beginning.

source/joker.c
```diff
@@ -271,25 +272,18 @@ bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum
         return false;
     }
 
-    JokerEffect joker_effect = joker_get_score_effect(joker_object->joker, card_object->card, joker_event);
+    JokerEffect* joker_effect = NULL;
+    u8 joker_ret = joker_get_score_effect(joker_object->joker, card_object->card, joker_event, &joker_effect);
```
MeirGavish: Rename to `joker_flags_ret`

source/joker.c
```diff
@@ -307,41 +301,46 @@ bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum
...
     {
+        *mult  = u32_protected_add(*mult,  joker_effect->mult);
         char score_buffer[INT_MAX_DIGITS + 2];
-        snprintf(score_buffer, sizeof(score_buffer), "+%lu", joker_effect.mult);
+        snprintf(score_buffer, sizeof(score_buffer), "+%lu", joker_effect->mult);
         set_and_shift_text(score_buffer, &cursorPosX, &cursorPosY, TTE_RED_PB);
         sfx_id = SFX_MULT;
     }
-    if (joker_effect.xmult > 0)
+    // if xmult is zero, DO NOT multiply by it
```
MeirGavish: Now that we're using flags, I think there isn't really a need to check that the effect field is > 0 except for xmult so we can drop that check and I think we can finally display when misprint gives +0?

source/joker_effects.c
MeirGavish: See my [reply to Copilot's comment](https://github.com/cellos51/balatro-gba/pull/245#discussion_r2565417253) about changing this to a non-pointer.

source/joker_effects.c
```diff
@@ -7,92 +7,98 @@
...
 }
 
-static JokerEffect sinful_joker_effect(Card *scored_card, u8 sinful_suit, enum JokerEvent joker_event)
+static u8 sinful_joker_effect(Card *scored_card, u8 sinful_suit, enum JokerEvent joker_event, JokerEffect **joker_effect)
 {
-    JokerEffect effect = {0};
+    SCORE_ON_EVENT_ONLY_WITH_CARD(scored_card, JOKER_EVENT_ON_CARD_SCORED, joker_event)
 
-    SCORE_ON_EVENT_ONLY_WITH_CARD(scored_card, JOKER_EVENT_ON_CARD_SCORED, joker_event, effect)
+    u8 effect_ret = JOKER_EFFECT_NONE;
```
MeirGavish: Rename this in the functions to `effect_flags_ret` to differentiate from the effect struct.

source/joker_effects.c
```diff
@@ -7,92 +7,98 @@
...
 
     if (scored_card->suit == sinful_suit)
     {
-        effect.mult = 3;
+        shared_joker_effect->mult = 3;
+        effect_ret = JOKER_EFFECT_MULT;
     }
 
-    return effect;
+    *joker_effect = shared_joker_effect;
```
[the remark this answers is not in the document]
MeirGavish: Generally it should be okay with the flags, but I still think it's better to move this inside the if statement now.

source/joker_effects.c
```diff
@@ -101,228 +107,250 @@ static JokerEffect zany_joker_effect(Joker *joker, Card *scored_card, enum Joker
...
+    u8 effect_ret = JOKER_EFFECT_NONE;
+    *joker_effect = shared_joker_effect;
 
     if (scored_card->rank == TEN || scored_card->rank == FOUR)
     {
-        effect.chips = 10;
-        effect.mult = 4;
+        shared_joker_effect->chips = 10;
+        shared_joker_effect->mult = 4;
+        effect_ret = JOKER_EFFECT_CHIPS + JOKER_EFFECT_MULT;
```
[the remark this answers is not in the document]
MeirGavish: I agree with these

source/joker_effects.c
```diff
@@ -901,10 +998,10 @@ static JokerEffect blueprint_brainstorm_joker_effect(Joker *joker, Card *scored_
 
                 // Then regardless of if we copied the data above, apply the
                 // copied JokerEffect function to the local data
-                effect = copied_joker_info->joker_effect(joker, scored_card, joker_event);
+                effect_ret = copied_joker_info->joker_effect_func(joker, scored_card, joker_event, joker_effect);
 
                 // make also sure we don't expire
-                effect.expire = false;
+                shared_joker_effect->expire = false;
```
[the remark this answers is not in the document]
MeirGavish: I think it's better to check for `JOKER_EFFECT_NONE` just to make sure the effect is valid.

MeirGavish: Also, this should access `joker_effect`, not `shared_joker_effect` and not rely on that they were set to be the same.

source/joker_effects.c
```diff
@@ -7,92 +7,98 @@
...
     u8 suits[NUM_SUITS];
     u8 ranks[NUM_RANKS];
     get_played_distribution(ranks, suits);
 
     if (hand_contains_n_of_a_kind(ranks) >= 2)
     {
-        effect.mult = 8;
+        shared_joker_effect->mult = 8;
+        effect_ret = JOKER_EFFECT_MULT;
     }
```
MeirGavish: This might be a bit of a nitpick but I also think that in these functions the setting of the joker_effect should be inside the condition. Another nitpick - after the assignment access the assigned `joker_effect` and access `shared_joker_effect` only before the assignment, it's a little more logical that way.

source/joker_effects.c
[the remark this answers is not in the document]
MathisMartin31: Yeah zeroing out the struct every time would defeat the entire purpose of this optimization imo

source/joker.c
MathisMartin31: You're right, and yes we'll finally be able to be disappointed by Misprint xD

MathisMartin31:
I'm looking at the suggestions right now, but Blueprint/Brainstorm decided to not work anymore lol, I'll have to fix that first

Edit: Aaaaand it's fixed, I forgot to protect the bit shifts for the Flags enum I turned into defines

source/joker_effects.c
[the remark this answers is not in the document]
MathisMartin31: I did this instead, which looks both safe and self explanatory to me.

`effect_flags_ret &= ~JOKER_EFFECT_FLAG_EXPIRE;

source/joker.c
```diff
@@ -160,12 +161,12 @@ void joker_destroy(Joker **joker)
     *joker = NULL;
 }
 
-JokerEffect joker_get_score_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
+u8 joker_get_score_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event, JokerEffect **joker_effect)
```
MeirGavish: The return type should be `u32` here too

source/joker.c
```diff
@@ -336,11 +336,11 @@ bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum
     }
     // custom message for Jokers (including retriggers where Jokers will say "Again!")
     // joker_effect->message will have been set if the Joker had anything custom to say
-    if (joker_ret & JOKER_EFFECT_MESSAGE && joker_effect->message != NULL && joker_effect->message[0] != '\0') // Message is not empty
+    if (joker_ret & JOKER_EFFECT_FLAG_MESSAGE)
     {
         set_and_shift_text(joker_effect->message, &cursorPosX, &cursorPosY, TTE_WHITE_PB);
     }
-    if (joker_ret & JOKER_EFFECT_EXPIRE && joker_effect->expire)
```
MeirGavish: I think you should keep the check for the `joker_effect->expire` field, otherwise the field has no meaning and the actual effect bool is the flag. I think it's better to keep it consistent with the rest of the fields by making the flags just enablers to check the fields rather than actual effects.

include/joker.h
```diff
@@ -61,6 +61,18 @@ enum JokerEvent
     JOKER_EVENT_ON_BLIND_SELECTED,  // Triggers when selecting a blind (e.g. Dagger, Riff Raff, Madness..)
 };
 
+// These are flags that can be combined into a single u8 and returned by
```
MeirGavish:
```suggestion
// These are flags that can be combined into a single u32 and returned by
```

Update the comment to reflect the change

include/joker.h
```diff
@@ -61,6 +61,18 @@ enum JokerEvent
...
+// JokerEffect functions to indicate what effects they had
```
MeirGavish:
```suggestion
// JokerEffect functions to indicate which fields of the output JokerEffect are valid
```

Maybe make the documentation a little more specific to what the flags affect.

include/joker.h
```diff
@@ -98,15 +110,14 @@ typedef struct  // These jokers are triggered after the played hand has finished
     u32 xmult;
     int money;
     bool retrigger; // Retrigger played hand (e.g. "Dusk" joker, even though on the wiki it says "On Scored" it makes more sense to have it here)
-    bool expire; // Joker is destroyed (food jokers)
     char* message; // Used to send custom messages e.g. "Extinct" or "-1" (Bananas and food Jokers)
 } JokerEffect;
 
-typedef JokerEffect (*JokerEffectFunc)(Joker *joker, Card *scored_card, enum JokerEvent joker_event);
+typedef u32 (*JokerEffectFunc)(Joker *joker, Card *scored_card, enum JokerEvent joker_event, JokerEffect **joker_effect);
```
MeirGavish, line 119: I think we should also add documentation here too about how the joker effect functions behave - their return value, the fact that JokerEffect** is an output, etc.

source/joker_effects.c
```diff
@@ -7,92 +7,98 @@
...
 }
 
-static JokerEffect sinful_joker_effect(Card *scored_card, u8 sinful_suit, enum JokerEvent joker_event)
+static u32 sinful_joker_effect(Card *scored_card, u8 sinful_suit, enum JokerEvent joker_event, JokerEffect **joker_effect)
 {
-    JokerEffect effect = {0};
+    SCORE_ON_EVENT_ONLY_WITH_CARD(scored_card, JOKER_EVENT_ON_CARD_SCORED, joker_event)
 
-    SCORE_ON_EVENT_ONLY_WITH_CARD(scored_card, JOKER_EVENT_ON_CARD_SCORED, joker_event, effect)
+    u8 effect_flags_ret = JOKER_EFFECT_FLAG_NONE;
```
```suggestion
    u32 effect_flags_ret = JOKER_EFFECT_FLAG_NONE;
```

Update the return value type

source/joker.c
MathisMartin31: I was actually tempted to do the same with the `retrigger` field but I'll do as you said it's more consistent

source/joker_effects.c
MathisMartin31: Oops, I forgot that one ^^'

source/joker.c
```diff
@@ -307,41 +301,46 @@ bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum
...
     // custom message for Jokers (including retriggers where Jokers will say "Again!")
-    if (joker_effect.message != NULL && joker_effect.message[0] != '\0') // Message is not empty
+    // joker_effect->message will have been set if the Joker had anything custom to say
+    if (effect_flags_ret & JOKER_EFFECT_FLAG_MESSAGE)
     {
-        set_and_shift_text(joker_effect.message, &cursorPosX, &cursorPosY, TTE_WHITE_PB);
+        set_and_shift_text(joker_effect->message, &cursorPosX, &cursorPosY, TTE_WHITE_PB);
     }
-    if (joker_effect.expire)
+    if (effect_flags_ret & JOKER_EFFECT_FLAG_EXPIRE)
```
```suggestion
    if (effect_flags_ret & JOKER_EFFECT_FLAG_EXPIRE && joker_effect.expire)
```
[This comment](https://github.com/cellos51/balatro-gba/pull/245#discussion_r2567911852)

MathisMartin31:
@MeirGavish I pushed your suggestion but there was a small typo and the ROM didn't compile. Should be fixed now and ready for merging ^^

PR #246, opened 2025-11-19
Format and document utility files

ricfehr3:
@MeirGavish There's 0 code change here, it's all formatting and documentation. I added the Doxyfile as a separate commit too. If you checkout this PR locally, you can build the docs locally. For me on linux it's just

```sh
cd balatro-gba
doxygen
# doxygen outputs to the doc directory
firefox doc/html/index.html
```

It should make reviewing the documentation portions of this PR a bit more enjoyable.

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -119,7 +119,7 @@ void truncate_uint_to_suffixed_str(
  *
  * @return the number of digits in a number
  */
-static inline int u32_get_digits(uint32_t n) 
+static inline int u32_get_digits(uint32_t n)
 {
     if (n < 10)
         return 1;


Doxyfile
```diff
@@ -0,0 +1,3018 @@
+# Doxyfile 1.15.0
```
ricfehr3, line 1: This is pretty close to the default Doxyfile. I made some tweaks with DoxyWizard and excluded directories. 

The tweaks were minimal, I remember just setting the output directory, project name, and an option to cross reference code line numbers to their descriptions.

.clang-format
```diff
@@ -48,3 +48,5 @@ ReflowComments: true
 SpacesInLineCommentPrefix:
   Minimum: 1
   Maximum: 1
+
+PenaltyReturnTypeOnItsOwnLine: 99999999
```
ricfehr3, line 53: This is a not so obvious options. It basically makes sure we get this:

```c
void long_function_name(bool test,
                        int param,
                        char star);
```

And note this:

```c
void
long_function_name(bool test, int param, char star);
```

MeirGavish: That's good

include/audio_utils.h
```diff
@@ -1,16 +1,57 @@
...
+#define SFX_DEFAULT_PAN MM_PAN_CENTER
+
+/**
+ * @brief Play a sound effect
+ *
+ * @param id the sound id to play, name an all caps derivation of the
+ *             associated in the `audio/` directory. E.g. `audio/card_draw.wav`
+ *             becomes SFX_CARD_DRAW
+ * @param rate the pitch rate, the default value is @ref MM_BASE_PTCH_RATE
+ */
```
MeirGavish:
```suggestion
 * @brief Play a sound effect, wrapper for mmEffectEx()
 *
 * @param id the sound id to play, from maxmod compiled soundbank.h header
 * @param rate the pitch rate, the default value is @ref MM_BASE_PTCH_RATE
 */
```

Since it's basically a maxmod wrapper, it's better to refer it maxmod

include/audio_utils.h
```diff
@@ -1,16 +1,57 @@
...
+#ifndef AUDIO_UTILS_H
+#define AUDIO_UTILS_H
+
+#include <mm_types.h>
+
+/**
+ * @def MM_FULL_VOLUME
+ * @brief The maxium volume for maxmod
+ *
+ * Represented as the max value of a unsigned char.
```
MeirGavish:
```suggestion
```

I don't think the actual value should be in the documentation.

include/audio_utils.h
```diff
@@ -1,16 +1,57 @@
...
+ * @brief Utilities for using maxmod to play sound effects
+ */
+#ifndef AUDIO_UTILS_H
+#define AUDIO_UTILS_H
+
+#include <mm_types.h>
+
+/**
+ * @def MM_FULL_VOLUME
+ * @brief The maxium volume for maxmod
```
MeirGavish:
```suggestion
 * @brief The maximum volume for maxmod mm_sound_effect.volume
```
typo + I guess this better clarifies why it's an unsigned char...

include/audio_utils.h
```diff
@@ -1,16 +1,57 @@
...
+ * @def MM_BASE_PITCH_RATE
+ * @brief The default pitch rate for the sound effect played.
+ *
+ * Increasing the rate increases the pitch, while decreasing lowers the pitch.
+ */
+#define MM_BASE_PITCH_RATE 1024
+
+/**
+ * @def SFX_DEFAULT_VOLUME
+ * @brief Default volume for sound effects (loudest by default)
```
MeirGavish:
```suggestion
 * @brief Default volume for sound effects
```

In here this value can theoretically change so there's a stronger reason to not have it.

include/audio_utils.h
```diff
@@ -1,16 +1,57 @@
...
+ *
+ * Represented as the max value of a unsigned char.
+ */
+#define MM_FULL_VOLUME 255
+
+/**
+ * @def MM_PAN_CENTER
+ * @brief The center pan of stereo audio for maxmode
+ *
+ * Represented as half the max value of an unsigned char.
```
MeirGavish:
```suggestion
 * @brief The center pan of stereo audio for maxmod
```

Removed the value + typo

include/graphic_utils.h
```diff
@@ -1,183 +1,368 @@
...
+ */
+#define TILE_SIZE 8
+
+/**
+ * @def EFFECT_TEXT_SEPARATION_AMOUNT
+ *
+ * @brief If we need to show multiple effects at once
+ */
+#define EFFECT_TEXT_SEPARATION_AMOUNT 32
+
```
MeirGavish:
```suggestion
```

Actually this doesn't seem to be used, I think it's a leftover and another variable was used in its stead.

include/graphic_utils.h
```diff
@@ -1,183 +1,368 @@
...
 // When making this, missed that it already exists in tonc_math.h
 typedef RECT Rect;
 
-/* Gets the screenblock entry for the given coordinates (x, y).
- * x and y are in number of tiles.
- * Returns the screenblock entry.
+/**
+ * @brief Gets the screenblock entry for the given coordinates (x, y).
+ *
+ * @param pos a @ref BG_POINT with x and y are in number of tiles.
```
MeirGavish:
```suggestion
 * @param pos a @ref BG_POINT with x and y in number of tiles.
```
Typo

include/graphic_utils.h
```diff
@@ -1,183 +1,368 @@
...
+ * @brief Get the height of a rectangle
+ *
+ * @param rect a @ref Rect to measure
+ *
+ * @return The height of the rectangle.
+ */
 INLINE int rect_height(const Rect* rect)
 {
     return (((rect)->bottom) - ((rect)->top) + 1);
 }
```
MeirGavish, line 183: When merging with #269 (whichever is merged first) the documentation needs to be updated that width returns 0 if `rect->right < rect->left` and likewise height for top and bottom

include/graphic_utils.h
```diff
@@ -1,183 +1,368 @@
...
- * bg_sbb is the SBB of the background in which to move the rect
- * direction must be either SE_UP or SE_DOWN.
- * se_rect dimensions are in number of tiles.
- * 
+/**
+ * @brief Copies an SE rect vertically in direction by a single tile.
+ *
  * NOTE: This does not work with TTE_SBB, probably because it's 4BPP...
+ *
+ * @param bg_sbb the SBB of the background in which to move the rect: the
```
MeirGavish:
```suggestion
/**
 * @brief Copies an SE rect vertically in direction by a single tile.
 *
 * NOTE: This does not work with TTE_SBB, probably because it's 4BPP...
 * 
 * If you are doing this operation you are probably doing this in the main
 * background and you should use main_bg_se_copy_rect_1_tile_vert() instead.
 *
 * @param bg_sbb the SBB of the background in which to move the rect: the
```

Looking at it I think this documentation needs a note referring `main_bg_se_copy_rect_1_tile_vert` and thinking about it I'm wondering if it makes sense to expose this function at all.

include/graphic_utils.h
```diff
@@ -1,183 +1,368 @@
...
+ * @param dst destination charblock
+ *
+ * @param src destination charblock
+ *
+ * @param wcount Number of words to copy
+ *
+ * @param palette_offset palette offset to shift to
  */
-void memcpy16_tile8_with_palette_offset(u16* dst, const u16* src, uint hwcount, u8 palette_offset);
+void memcpy16_tile8_with_palette_offset(u16* dst, const u16* src, uint wcount, u8 palette_offset);
```
MeirGavish:
```suggestion
void memcpy16_tile8_with_palette_offset(u16* dst, const u16* src, uint hwcount, u8 palette_offset);
```

I think this needs to stay `hwcount` because it's the number of u16s which are half-words, correct me if I'm wrong. Likewise in the documentation

include/util.h
```diff
@@ -1,55 +1,170 @@
...
+ */
+uint32_t u32_protected_add(uint32_t a, uint32_t b);
+
+/**
+ * @brief Avoid overflow when adding two u16 integers
+ *
+ * @param a left operator **a + b**
+ * @param b left operator **a + b**
+ *
+ * @return the result of **a + b** or **UINT16_MAX**
```
MeirGavish:
```suggestion
 * @return the result of **a + b** or **UINT16_MAX** in case of overflow
```
Just a small clarification

include/util.h
```diff
@@ -1,55 +1,170 @@
...
+ */
+uint16_t u16_protected_add(uint16_t a, uint16_t b);
+
+/**
+ * @brief Avoid overflow when multiplying two u32 integers
+ *
+ * @param a left operator **a * b**
+ * @param b left operator **a * b**
+ *
+ * @return the result of **a * b** or **UINT32_MAX**
```
MeirGavish:
```suggestion
 * @return the result of **a * b** or **UINT32_MAX** in case of overflow
```

Here too

include/util.h
```diff
@@ -1,55 +1,170 @@
...
+ */
+#define UINT8_MAX_DIGITS 3
+
+/**
+ * @brief Avoid overflow when adding two u32 integers
+ *
+ * @param a left operator **a + b**
+ * @param b left operator **a + b**
+ *
+ * @return the result of **a + b** or **UINT32_MAX**
```
MeirGavish:
```suggestion
 * @return the result of **a + b** or **UINT32_MAX** in case of overflow
```

include/util.h
```diff
@@ -1,55 +1,170 @@
...
+ */
+uint32_t u32_protected_mult(uint32_t a, uint32_t b);
+
+/**
+ * @brief Avoid overflow when multiplying two u16 integers
+ *
+ * @param a left operator **a * b**
+ * @param b left operator **a * b**
+ *
+ * @return the result of **a * b** or **UINT16_MAX**
```
MeirGavish:
```suggestion
 * @return the result of **a * b** or **UINT16_MAX** in case of overflow
```

include/util.h
```diff
@@ -1,55 +1,170 @@
...
+
+/**
+ * @brief Get the number of digits in a number
+ * https://stackoverflow.com/questions/1068849/how-do-i-determine-the-number-of-digits-of-an-integer-in-c
+ *
+ * @param n value to find the number of decimal digits of
+ *
+ * @return the number of digits in a number
+ */
+static inline int get_digits(int n)
```
MeirGavish: Again with #269 this will need to be updated that it is meant for u32.

include/util.h
```diff
@@ -1,55 +1,170 @@
...
+/**
+ * @brief Get the number of digits closest to it's "even" value
+ *
+ * Useful for centering text on tiles
+ *
+ * @param n value to find the number of even decimal digits of
+ *
+ * @return The closest number of digits that is an even value.
+ */
 static inline int get_digits_even(int n)
```
MeirGavish, line 154:
```suggestion
/**
 * @brief Get the number of digits closest to it's "even" value, i.e. ceil(digits/2)
 *
 * Useful for centering text on tiles
 *
 * @param n value to find the number of even decimal digits of
 *
 * @return The value of ceil(get_digits(n)/2)
 */
static inline int get_digits_even(int n)
```

A more accurate documentation. Not bothering for `get_digits_odd()` because it's removed in #269.

ricfehr3:
@MeirGavish This one should be good to go. Huge thank you on having the changes ready to batch and commit. I'm keeping note of the dependency on #269 . If it's merged first I'll change it, other wise I guess we could change it in #269.

ricfehr3:
@MeirGavish Just rebased this one. I'll focus on this one and do the joker one last at #257

include/graphic_utils.h
```diff
@@ -68,104 +153,147 @@ enum OverflowDir
...
  * se_rect dimensions are in number of tiles.
- * 
+ *
  * NOTE: This does not work with TTE_SBB, probably because it's 4BPP...
+ *
+ * If you are doing this operation you are probably doing this in the main
+ * background and you should use main_bg_se_copy_rect_1_tile_vert() instead.
+ *
+ * @param bg_sbb the SBB of the background in which to move the rect: the
+ * direction must be either SE_UP or SE_DOWN.
```
MeirGavish:
```suggestion
 * @param bg_sbb the SBB of the background in which to move the rect.
```

SE_UP and SE_DOWN were removed. The possible values are now implied from the enum type so it's better to just remove it from the documentation.

include/util.h
```diff
@@ -49,36 +41,128 @@ static inline int get_digits_even(int n)
...
 
-uint32_t u32_protected_add (uint32_t a, uint32_t b);
-uint16_t u16_protected_add (uint16_t a, uint16_t b);
-uint32_t u32_protected_mult(uint32_t a, uint32_t b);
-uint16_t u16_protected_mult(uint16_t a, uint16_t b);
+/**
+ * @brief Get the number of digits in a number
+ * https://stackoverflow.com/questions/1068849/how-do-i-determine-the-number-of-digits-of-an-integer-in-c
+ *
+ * @param n value to find the number of decimal digits of
```
MeirGavish:
```suggestion
/**
 * @brief Get the number of digits in a 32-bit unsigned number
 * https://stackoverflow.com/questions/1068849/how-do-i-determine-the-number-of-digits-of-an-integer-in-c
 *
 * @param n 32-bit unsigned value to find the number of decimal digits of
```

Just to clarify it's meant specifically for an unsigned 32-bit number

include/util.h
```diff
@@ -49,36 +41,128 @@ static inline int get_digits_even(int n)
...
+ *
+ * @return the number of digits in a number
+ */
+static inline int u32_get_digits(uint32_t n
+) // https://stackoverflow.com/questions/1068849/how-do-i-determine-the-number-of-digits-of-an-integer-in-c
```
MeirGavish:
```suggestion
static inline int u32_get_digits(uint32_t n) 
```

Since the link now appears in the documentation, it can be removed from here and make this cleaner

ricfehr3:
I applied the changes and I'm merging now, despite github saying I need to re-request.

PR #252, opened 2025-11-22
Implement Seltzer and Joker Expiration effect

MathisMartin31:
Hi all, back again with a new PR ^^

As was discussed with @MeirGavish in PR #244, Seltzer was needed to fully test the new Blueprint/Brainstorm implementation. The Joker already had its code, and I did the art the other day, so I thought it was time to give it a try.

This PR introduces the Expiration Joker effect, where Jokers with a limited lifetime like food Jokers can expire. I tested it and it seems to work, especially with the new List implementation that makes it easy to modify a list that is being iterated on:

https://github.com/user-attachments/assets/876cc58b-580e-4803-8a91-a66092553f8f

I will leave this PR as a draft for now because I need to test the new Joker with the copying Jokers, but I don't expect too many issues there since most of the work has already been done in the mentioned PR.

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -120,10 +120,13 @@ int get_num_hands_remaining(void);
 
 u32     get_chips(void);
 void    set_chips(u32 new_chips);
+void    display_chips();
 u32     get_mult(void);
 void    set_mult(u32 new_mult);
+void    display_mult();
 int     get_money(void);
 void    set_money(int new_money);
+void    display_money();
 void    set_retrigger(bool new_retrigger);
 
 
--- a/source/game.c
+++ b/source/game.c
@@ -1163,11 +1163,11 @@ void display_score(u32 value)
     tte_printf("#{P:%d,48; cx:0x%X000}%lu%c", x_offset, TTE_WHITE_PB, display_value, score_suffix);
 }
 
-void display_money(int value)
+void display_money()
 {
-    int x_offset = 32 - get_digits_odd(value) * TILE_SIZE;
+    int x_offset = 32 - get_digits_odd(money) * TILE_SIZE;
     tte_erase_rect_wrapper(MONEY_TEXT_RECT);
-    tte_printf("#{P:%d,%d; cx:0x%X000}$%d", x_offset, MONEY_TEXT_RECT.top, TTE_YELLOW_PB, value);
+    tte_printf("#{P:%d,%d; cx:0x%X000}$%d", x_offset, MONEY_TEXT_RECT.top, TTE_YELLOW_PB, money);
 }
 
 // Show/Hide flaming score effect if we will score
@@ -1575,7 +1575,7 @@ void game_start()
     display_hands(hands); // Hand
     display_discards(discards); // Discard
 
-    display_money(money); // Set the money display
+    display_money(); // Set the money display
 
     tte_printf("#{P:%d,%d; cx:0x%X000}%d#{cx:0x%X000}/%d", ANTE_TEXT_RECT.left, ANTE_TEXT_RECT.top, TTE_YELLOW_PB, ante, TTE_WHITE_PB, MAX_ANTE); // Ante
 
@@ -2230,10 +2230,6 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
     {
         if (joker_object_score(joker, card_object, joker_event))
         {
-            display_chips();
-            display_mult();
-            display_money(money);
-
             return true;
         }
     }
@@ -2700,7 +2696,7 @@ static int calculate_interest_reward()
 static void game_round_end_cashout()
 {
     money += hands + blind_get_reward(current_blind) + calculate_interest_reward(); // Reward the player
-    display_money(money);
+    display_money();
 
     hands = max_hands; // Reset the hands to the maximum
     discards = max_discards; // Reset the discards to the maximum
@@ -3153,7 +3149,7 @@ static void game_shop_intro()
 static void game_shop_reroll(int *reroll_cost)
 {
     money -= *reroll_cost;
-    display_money(money); // Update the money display
+    display_money(); // Update the money display
 
     ListItr itr = list_itr_create(&_shop_jokers_list);
     JokerObject* joker_object;
@@ -3232,7 +3228,7 @@ void game_sell_joker(int joker_idx)
     
     JokerObject* joker_object = (JokerObject*)list_get_at_idx(&_owned_jokers_list, joker_idx);
     money += joker_get_sell_value(joker_object->joker);
-    display_money(money);
+    display_money();
     erase_price_under_sprite_object(joker_object->sprite_object);
 
     remove_owned_joker(joker_idx);
@@ -3267,7 +3263,7 @@ static void game_shop_buy_joker(int shop_joker_idx)
     JokerObject *joker_object = (JokerObject*)list_get_at_idx(&_shop_jokers_list, shop_joker_idx);
 
     money -= joker_object->joker->value; // Deduct the money spent on the joker
-    display_money(money);                // Update the money display
+    display_money();                     // Update the money display
     erase_price_under_sprite_object(joker_object->sprite_object);
     sprite_object_set_focus(joker_object->sprite_object, false);
     add_to_held_jokers(joker_object);
@@ -3875,7 +3871,7 @@ static void game_over_on_exit()
     display_mult();
     display_hands(hands);
     display_discards(discards);
-    display_money(money);
+    display_money();
     tte_printf(
         "#{P:%d,%d; cx:0x%X000}%d#{cx:0x%X000}/%d",
         ANTE_TEXT_RECT.left,
--- a/source/joker.c
+++ b/source/joker.c
@@ -353,9 +353,15 @@ bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum
         list_push_back(get_expired_jokers_list(), joker_object);
     }
 
+    // Update values
     set_chips(chips);
     set_mult(mult);
     set_money(money);
+
+    // Update displays
+    display_chips();
+    display_mult();
+    display_money();
 
     joker_object_shake(joker_object, sfx_id);
 


source/game.c
```diff
@@ -2191,12 +2192,21 @@ static bool check_and_score_joker_for_event(ListItr* starting_joker_itr, CardObj
...
             display_mult();
             display_money(money);
 
+            if (expire)
+            {
+                // remove the joker directly, no need for the protections of
+                // the remove_owned_joker function because they don't concern Jokers
+                // that can expire
+                list_itr_remove_current_node(starting_joker_itr);
+                joker_object_destroy(&joker);
```
MeirGavish: It's not good to rely on this, all joker removals should go through a single function so any other changes done in the future for removed jokers are also performed. Currently this is missing the call to `_set_shop_joker_avail()` so an expired Seltzer will not become available to the shop again and [it should](https://www.reddit.com/r/balatro/comments/1i6d27a/can_food_jokers_reappear_in_the_shop_after_being/) (the wiki doesn't state it shouldn't either).
I guess the correct thing to do is to add several functions to remove owned jokers - `remove_held_jokers_by_idx`, `remove_held_jokers_by_itr` which both call a function like`handle_owned_joker_removal(JokerObject* joker_object)` that all removals should go through.

MathisMartin31:
I'm testing it right now, and I'm thinking that Blueprint and Brainstorm should not copy events that are used solely to update the persistent state of scaling/food Jokers. Looking at the wiki, "On Hand Scored End" seems to be a strong contender for such an exception, as I have yet to find a Joker that actually scores something there.

MeirGavish:
> I'm testing it right now, and I'm thinking that Blueprint and Brainstorm should not copy events that are used solely to update the persistent state of scaling/food Jokers. Looking at the wiki, "On Hand Scored End" seems to be a strong contender for such an exception, as I have yet to find a Joker that actually scores something there.

What about Space Joker? It has a chance to upgrade the hand's level when you play it but before it is scored.

MathisMartin31:
> What about Space Joker? It has a chance to upgrade the hand's level when you play it but before it is scored.

Looking at the wiki, Space Joker is one of 4 Jokers that scores "On (Hand) Played", 3 of which are compatible with Blueprint/Brainstorm. The fourth one is Midas Mask but I don't think it would be a huge issue to copy it because turning a Gold card into a Gold card doesn't do anything besides being inefficient.

Looking at our code, the "On Hand Played" event is used to update the `scoring_state` of some Jokers, and it is being copied right now with no apparent issues.

All in all, the "On Hand Played" event and Space Joker are fine to copy ^^

Edit: Thinking about it, maybe you mixed up "On Hand **Played**" and "On Hand **Scored End**" ?

MeirGavish:
> Edit: Thinking about it, maybe you mixed up "On Hand **Played**" and "On Hand **Scored End**" ?

Yes, I misread it 😅
It seems "On Hand **Scored End**" should be good to ignore.

source/joker_effects.c
```diff
@@ -966,31 +966,35 @@ static JokerEffect seltzer_joker_effect(Joker *joker, Card *scored_card, enum Jo
...
 
         case JOKER_EVENT_ON_HAND_SCORED_END:
+            (*p_hands_left_until_exp)--;
             if (*p_hands_left_until_exp > 0)
             {
-                (*p_hands_left_until_exp)--;
-                effect.message = "-1";
+                // Need to do this for now because the message's memory can't really be allocated
+                // So we can't use snprintf to craft a message depending on the number of hands left
+                static char* seltzer_messages[9] = {
```
ricfehr3: Slightly more correct to use `static const`

source/joker.c
```diff
@@ -264,7 +264,7 @@ void set_and_shift_text(char* str, int* cursor_pos_x, int* cursor_pos_y, int col
     *cursor_pos_x += joker_score_display_offset_px;
 }
 
-bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum JokerEvent joker_event, u32 *chips, u32 *mult, int *money, bool *retrigger)
+bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum JokerEvent joker_event, u32 *chips, u32 *mult, int *money, bool *retrigger, bool *expire)
```
ricfehr3: I think it may be easier to pass a pointer to a `JokerEffect` struct here rather than `int* money, bool* retrigger, bool* expire`, and bubbled it up to the static vars in `game.c` for `retrigger` and `money`.

MeirGavish: Yeah, I've never been fond of passing all these pointers in this function... Since chips, mult, and money are all global variables maybe it's just better to provide get and set functions for all of them in game.c/.h and let this function just call them.

source/game.c
MathisMartin31: Oh yeah you're right.

One caveat I see though, is that : let's say I have Showman and get a Joker twice, then I sell Showman and then one of my duplicates either is sold or expires. Is it available right then, or do we check if we have any of these Jokers left?

source/joker_effects.c
```diff
@@ -966,31 +966,35 @@ static JokerEffect seltzer_joker_effect(Joker *joker, Card *scored_card, enum Jo
...
+            if (effect.retrigger)
             {
-                effect.retrigger = (*p_last_retriggered_idx < get_scored_card_index());
-                if (effect.retrigger)
-                {
-                    *p_last_retriggered_idx = get_scored_card_index();
-                    effect.message = "Again!";
-                }
-            } 
+                (*p_last_retriggered_idx) = get_scored_card_index();
```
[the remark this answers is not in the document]
MathisMartin31: Yeah I had an error once with something like `*some_pointer++` and went paranoid over pointers ever since lol

source/joker_effects.c
```diff
@@ -966,31 +966,35 @@ static JokerEffect seltzer_joker_effect(Joker *joker, Card *scored_card, enum Jo
...
         case JOKER_EVENT_ON_HAND_SCORED_END:
+            (*p_hands_left_until_exp)--;
             if (*p_hands_left_until_exp > 0)
             {
-                (*p_hands_left_until_exp)--;
-                effect.message = "-1";
+                // Need to do this for now because the message's memory can't really be allocated
+                // So we can't use snprintf to craft a message depending on the number of hands left
+                static char* seltzer_messages[9] = {
+                    "1", "2", "3", "4", "5", "6", "7", "8", "9"
```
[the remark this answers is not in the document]
MathisMartin31: We cannot reach this part of the code without decrementing `*p_hands_left_until_exp` first, which makes 9 its highest value, thus 8 the highest index we will seek in the array

source/game.c
MeirGavish: Yeah, the shop doesn't currently have proper full support for Showman and multiple jokers, the system currently assumes you can only have one of each joker. This is true not only for expiration, but also for selling jokers, if you have 2 of the same joker and sell one of them, it will become available to shop even if you don't have Showman.

source/joker_effects.c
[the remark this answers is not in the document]
MeirGavish: They are not needed in this case but it doesn't matter so much

source/joker_effects.c
```diff
@@ -966,31 +966,35 @@ static JokerEffect seltzer_joker_effect(Joker *joker, Card *scored_card, enum Jo
...
             break;
 
         case JOKER_EVENT_ON_HAND_SCORED_END:
+            (*p_hands_left_until_exp)--;
             if (*p_hands_left_until_exp > 0)
             {
-                (*p_hands_left_until_exp)--;
-                effect.message = "-1";
+                // Need to do this for now because the message's memory can't really be allocated
+                // So we can't use snprintf to craft a message depending on the number of hands left
```
MeirGavish, line 1135: Oh, I guess this was a miss on my part for the message, I didn't realize Seltzer displays a dynamic message (maybe there are others I missed?)... it can be changed back to a buffer... or I guess you can also allocate a static buffer and write to it... or you can leave it like this I guess...

source/joker_effects.c
```suggestion
                static char* seltzer_messages[] = {
```

If you leave it like this you can also delete the magic number size and rely on the assigned size

source/game.c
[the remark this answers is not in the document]
MeirGavish: Why is `expire` a global variable anyway? It's not used anywhere but this function, it should be a local variable, it would be much cleaner.

MathisMartin31: Yeah I think we should worry about this later (aka *if* we add Showman)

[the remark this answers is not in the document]
MathisMartin31: You're right I did that to match the retriggers and stuff but that's kind of dumb ^^'

source/joker_effects.c
MathisMartin31: Hmm, maybe we should switch back to the buffer, but in PR #245 once this one is merged or vice versa. Having a pre-allocated buffer should be easier for us to deal with with a single shared JokerEffect instance

source/joker_effects.c
MathisMartin31: Oh yeah you're right I missed that ^^

source/joker.c
MathisMartin31: I added getters for all of them but `expire`, since it is now a local variable

source/game.c
MathisMartin31: So I changed how it works, and while the Joker removal isn't completely centralized in one function, all can be easily tracked from `jokers_update_loop`.

For the small animation I wanted to work, it had to be done slightly differently than selling Jokers but the logic is very similar

MathisMartin31:
I know this adds a bit of complexity to the PR but I thought the issue was directly linked to what I was doing.

I made a remark on some PR some time ago that there were pauses between the different states of scoring. This is normal behaviour, since scoring everything without these pauses would render the process more difficult to follow.

However, even when we had nothing to score at a given step, these pauses were still made. Meaning that if, for example, we had a Sin Joker that triggers "On Card Scored", no "On Card Held" Jokers, and then an Independent Joker like the Default one, then we would have waited twice between the "On Card Scored" and "Independent" steps.

What I did is add a global bool storing if whether or not **anything** scored at a given step, and if not, directly call the next step update function instead of going through the main loop in `played_cards_update_loop`.

I thought this belonged to this PR because I separated the "On Hand Scored End" into another separate state to be able to have that pause between them and the Independent Jokers, but this added yet another jarring delay when we had nothing that scored during that step.

include/game.h
```diff
@@ -113,7 +116,12 @@ List*           get_jokers_list(void);
 int get_deck_top(void);
 int get_num_discards_remaining(void);
 int get_num_hands_remaining(void);
-int get_money(void);
+
+uint32_t*   get_chips(void);
+uint32_t*   get_mult(void);
+int*        get_money(void);
+bool*       get_retrigger(void);
```
MeirGavish: Don't return pointers now, return the value in the `get` functions, and provide `set` functions for modification.
Also include tonc.h here and use `u32`.

include/joker.h
```diff
@@ -125,7 +125,7 @@ JokerObject *joker_object_new(Joker *joker);
 void joker_object_destroy(JokerObject **joker_object);
 void joker_object_update(JokerObject *joker_object);
 void joker_object_shake(JokerObject *joker_object, mm_word sound_id); // This doesn't actually score anything, it just performs an animation and plays a sound effect
-bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum JokerEvent joker_event, u32 *chips, u32 *mult, int *money, bool *retrigger, bool *expire); // This scores the joker and returns true if it was scored successfully (Card = NULL means the joker is independent and not scored by a card)
+bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum JokerEvent joker_event, bool *expire); // This scores the joker and returns true if it was scored successfully (Card = NULL means the joker is independent and not scored by a card)
```
MeirGavish: This is related to this and previous PRs, this documentation isn't very correct anymore. The case of `card_object = NULL` needs to be changed to say that NULL can be passed there if the event does not concern a card, maybe also give examples for events that do concern a card.

source/game.c
```diff
@@ -231,6 +230,8 @@ static bool discarded_card = false;
 static ListItr _joker_scored_itr;
 static ListItr _joker_card_scored_end_itr;
 static ListItr _joker_round_end_itr;
+// if no jokers have had any effect at a certain step, do not reset timer to TM_ZERO
+static bool has_any_joker_scored = false;
```
MeirGavish: This comment explains an action that doesn't happen here rather than the variable, explain the meaning of the variable instead.

source/game.c
```diff
@@ -3669,6 +3757,37 @@ static void game_main_menu_on_update()
...
+        {
+            // get joker idx
+            int expired_joker_idx = 0;
+            ListItr joker_itr = list_itr_create(&_owned_jokers_list);
+            JokerObject* expired_joker;
+            while ((expired_joker = list_itr_next(&joker_itr)) && expired_joker != joker_object)
+            {
+                expired_joker_idx++;
+            }
+            remove_owned_joker(expired_joker_idx);
```
MeirGavish, line 3757: Is this what you meant when you said you had to do it slightly differently for the animation to work? That is, putting the call to `remove_owned_joker()` here and not at the time of expiration like when the joker is sold. Is it so the re-spacing of the joker not happen before the animation ends? If so, add a comment explaining that.

source/game.c
```diff
@@ -3669,6 +3757,37 @@ static void game_main_menu_on_update()
...
+
+    ListItr itr = list_itr_create(&_expired_jokers_list);
+    JokerObject* joker_object;
+
+    while((joker_object = list_itr_next(&itr)))
+    {
+        joker_object_update(joker_object);
+
+        // let just enough frames pass that we see it rotating and shrinking
+        if (timer % FRAMES(3) == 0)
```
MeirGavish: Make a define for this magic number... Though honestly I don't think I would have re-used the shake animation for the expiration but we can leave it like this for now and make a dedicated expiration animation later.

source/game.c
```diff
@@ -2377,15 +2437,35 @@ static bool play_scoring_independent_jokers_update(int played_idx)
...
+
+        tte_erase_rect_wrapper(PLAYED_CARDS_SCORES_RECT);
+
         if (check_and_score_joker_for_event(&_joker_round_end_itr, NULL, JOKER_EVENT_ON_HAND_SCORED_END))
         {
             return true;
         }
 
         play_state = PLAY_ENDING;
-        timer = TM_ZERO;
```
MeirGavish, line 2387: Shouldn't you have moved this timer reset somewhere? I actually think it should be done outside this function in `played_cards_update_loop()`

source/game.c
MathisMartin31: Yes this is exactly what I had in mind, adjusting the timing for it to look good was kind of tricky.
I'll put a comment there to explain

source/game.c
MathisMartin31: I'll be removing this variable anyway, since its only purpose is for the step skipping when nothing scores and this will be part of another PR

source/game.c
MathisMartin31: I'll put it back how it was and we'll see what to do with the timers and the steps in the next PR dedicated to step pacing

source/game.c
```diff
@@ -230,7 +229,7 @@ static bool discarded_card = false;
 static ListItr _joker_scored_itr;
 static ListItr _joker_card_scored_end_itr;
 static ListItr _joker_round_end_itr;
-// if no jokers have had any effect at a certain step, do not reset timer to TM_ZERO
+// stores whether any joker has had any effect during a certain scoring step
 static bool has_any_joker_scored = false;
```
MeirGavish: Remove this variable for now since it's unused

source/game.c
```diff
@@ -2184,19 +2218,32 @@ static void cards_in_hand_update_loop()
...
             display_chips();
             display_mult();
             display_money(money);
 
+            if (expire)
+            {
+                joker_start_expire_animation(joker);
+            }
+
+            has_any_joker_scored = true;
```
```suggestion
```

Remove the unused variable for now.

source/joker.c
```diff
@@ -341,13 +351,14 @@ bool joker_object_score(JokerObject *joker_object, CardObject* card_object, enum
...
         set_and_shift_text(joker_effect.message, &cursorPosX, &cursorPosY, TTE_WHITE_PB);
     }
+
+    joker_object_shake(joker_object, sfx_id);
+
     if (joker_effect.expire)
     {
-        // TODO make Jokers expire
+        // TODO: Expire sound effect
     }
```
MeirGavish: Looking at this I'm suddenly wondering, why is the expiration code in `check_and_score_joker_for_event` and not here? It feels like the division between the functions is a little arbitrary and the following should be here too instead of there

```
            display_chips();
            display_mult();
            display_money(money);
```

and `check_and_score_joker_for_event()` should just be a for-loop.

It makes me think the responsibilities are not properly dividied and maybe a bigger refactor might be needed here but that's for another time.

MathisMartin31: I put some things inside `joker_object_score`, I think it makes sense how it is now but tell me if you disagree

MeirGavish (approved):
Great!

PR #253, opened 2025-11-23
Clang Format Blinds

ricfehr3:
Quick cleanup of bitset docs and rename of of the `bitset_allocate_idx()` function. 

**Edit**: Ignore the branch name. I messed up checking out another branch, but it's not worth it to get another pr with that small bitset change. 

btw @MeirGavish I'm going to defer adding doxygen docs to another ticket. Partly out of laziness, but more importantly we are going to modify this code soon.

Also labeling this one and others because I'm about to flood PRs of clang-format additions

Related to #136

### DIFF
--- a/source/blind.c
+++ b/source/blind.c
@@ -4,6 +4,7 @@
 #include "boss_blind_gfx.h"
 #include "graphic_utils.h"
 #include "small_blind_gfx.h"
+#include "util.h"
 
 #include <tonc.h>
 
@@ -43,7 +44,7 @@ static Blind _blind_type_map[BLIND_TYPE_MAX] = {
 
 static void s_blind_gfx_init(enum BlindType type);
 
-__attribute__((unused)) void blind_set_boss_graphics(const unsigned int* tiles, const u16* palette)
+GBLA_UNUSED void blind_set_boss_graphics(const unsigned int* tiles, const u16* palette)
 {
     // TODO: This function is unused and not fully fleshed out.
     // We need to support more boss blind graphics in the future.
--- a/include/util.h
+++ b/include/util.h
@@ -3,6 +3,8 @@
 
 #include <stdint.h>
 
+#define GBLA_UNUSED __attribute__((unused))
+
 static inline int get_digits(int n) // https://stackoverflow.com/questions/1068849/how-do-i-determine-the-number-of-digits-of-an-integer-in-c
 {
     if (n < 10) return 1;
--- a/source/game.c
+++ b/source/game.c
@@ -277,7 +277,7 @@ static inline void _set_shop_joker_avail(int joker_id, bool avail)
     bitset_set_idx(&_avail_jokers_bitset, joker_id, avail);
 }
 
-__attribute__((unused))
+GBLA_UNUSED
 static inline bool _get_shop_joker_avail(int joker_id)
 {
     return bitset_get_idx(&_avail_jokers_bitset, joker_id);
--- a/source/joker_effects.c
+++ b/source/joker_effects.c
@@ -733,10 +733,10 @@ static JokerEffect bootstraps_joker_effect(Joker *joker, Card *scored_card, enum
 }
 
 
-// Using __attribute__((unused)) for jokers with no sprites yet to avoid warning
+// Using GBLA_UNUSED, aka __attribute__((unused)), for jokers with no sprites yet to avoid warning
 // Remove the attribute once they have sprites
 // no graphics available but ready to be used if wanted when graphics available
-__attribute__((unused))
+GBLA_UNUSED
 static JokerEffect shoot_the_moon_joker_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
 {
     JokerEffect effect = {0};
@@ -752,7 +752,7 @@ static JokerEffect shoot_the_moon_joker_effect(Joker *joker, Card *scored_card, 
 }
 
 
-__attribute__((unused))
+GBLA_UNUSED
 static JokerEffect photograph_joker_effect(Joker *joker, Card *scored_card, enum JokerEvent joker_event)
 {
     JokerEffect effect = {0};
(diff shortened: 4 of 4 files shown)


include/blind.h
```diff
@@ -26,17 +26,15 @@ enum BlindColorIndex
     BLIND_BACKGROUND_SHADOW_COLOR_INDEX = 7,
 };
 
-#define BLIND_TYPE_INFO_TABLE                   \
-    BLIND_INFO(SMALL, small, FIX_ONE, 3)        \
-    BLIND_INFO(BIG, big, (FIX_ONE * 3) / 2, 4)  \
+#define BLIND_TYPE_INFO_TABLE                                                                                          \
+    BLIND_INFO(SMALL, small, FIX_ONE, 3)                                                                               \
+    BLIND_INFO(BIG, big, (FIX_ONE * 3) / 2, 4)                                                                         \
```
MeirGavish: Does clang enforce the macro newline character to be at the 120th column? It's looks a little far and jumps a line in the differ for me. I think 120 is a good absolute upper limit  though we still need to strive not to reach it and usually try to stop at 80-100 columns.

include/blind.h
```diff
@@ -26,17 +26,15 @@ enum BlindColorIndex
...
     BLIND_INFO(BOSS, boss, FIX_ONE * 2, 5)
 
 enum BlindType
 {
-#define BLIND_INFO(NAME, name, multi, reward)   \
-    BLIND_TYPE_##NAME ,
-    BLIND_TYPE_INFO_TABLE
-    BLIND_TYPE_MAX,
+#define BLIND_INFO(NAME, name, multi, reward) BLIND_TYPE_##NAME,
+    BLIND_TYPE_INFO_TABLE BLIND_TYPE_MAX,
```
MeirGavish: Can't we have those in different lines?

source/blind.c
```diff
@@ -1,52 +1,49 @@
...
+        .gfx_info =                                                                                                    \
+        {                                                                                                              \
+                .tiles = name##_blind_gfxTiles,                                                                        \
+                .palette = name##_blind_token_palette,                                                                 \
+                .tid = NAME##_BLIND_TID,                                                                               \
+                .pb = NAME##_BLIND_PB,                                                                                 \
+        },                                                                                                             \
+        .score_req_multipler = multi,                                                                                  \
+        .reward = _reward,                                                                                             \
+},
```
MeirGavish: Same here for the newline marker

source/blind.c
```diff
@@ -1,52 +1,49 @@
...
-    BlindGfxInfo* p_gfx = &_blind_type_map[type].gfx_info;
-    memcpy32(&tile_mem[4][p_gfx->tid], p_gfx->tiles, BLIND_SPRITE_COPY_SIZE);
-    memcpy16(&pal_obj_bank[p_gfx->pb], p_gfx->palette, PAL_ROW_LEN);
-}
-
+static void s_blind_gfx_init(enum BlindType type);
 
-__attribute__((unused))
-void blind_set_boss_graphics(const unsigned int* tiles, const u16* palette)
+__attribute__((unused)) void blind_set_boss_graphics(const unsigned int* tiles, const u16* palette)
```
MeirGavish: We can't have the unused attribute in another line?

source/blind.c
```diff
@@ -1,52 +1,49 @@
...
-{
-    // TODO: Re-add grit copy. You need to decouple the blind graphics first.
-    // This will allow this function to change the boss graphics info
-    //GRIT_CPY(&tile_mem[4][_blind_type_map[type].pal_info.tid], tiles);
-    BlindGfxInfo* p_gfx = &_blind_type_map[type].gfx_info;
-    memcpy32(&tile_mem[4][p_gfx->tid], p_gfx->tiles, BLIND_SPRITE_COPY_SIZE);
-    memcpy16(&pal_obj_bank[p_gfx->pb], p_gfx->palette, PAL_ROW_LEN);
-}
-
+static void s_blind_gfx_init(enum BlindType type);
```
MeirGavish, line 45: Does our style guide demand forward declarations for static functions? I don't mind them but I usually avoid them if possible so there is no separate definition and declaration.

include/blind.h
ricfehr3: I agree, it's a little jarring. 

I found this [option](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignescapednewlines) and applied `Left`. If you like it here. Let's merge this one next and I can redo the other PRs with that addition.

include/blind.h
ricfehr3: I just looked into it and it looks like x-macro formatting is not done well with clang format. So I think I'll just turn off formatting in our xmacros.

source/blind.c
ricfehr3: I'm looking into this one. I would like it as well.

source/blind.c
ricfehr3: This is definitely up for argument. I personally like to do this mostly because it makes it so we don't have to worry about functions being declared out of order when moving them around. I also think it looks neat and well organized to have a "table of contents" of functions. But whatever decision we make, I'll make it work.

source/blind.c
ricfehr3: I think... it's not going to work. What I think we could do is take anything that is `__attribute__` and put them behind macros like `GBLA_UNUSED`. It would help with the visual vomit. Otherwise we can disable formatting for these declarations.

source/blind.c
MeirGavish: Yeah, I'm fine with that, there are benefits to both approaches.

source/blind.c
MeirGavish: Yeah, we should define the macro

include/blind.h
MeirGavish: Cool! `Left` is fine, I think `LeftWithLastLine` looks better in the example but it doesn't matter much.

include/blind.h
MeirGavish: Yeah, probably the best thing to do

MeirGavish (approved):

ricfehr3:
@MeirGavish Again as my last merge, since you approved the changes I'm going to merge this one in and rebase the other branches so you can merge the next on easily.

PR #257, opened 2025-11-23
Clang format joker files and def files

ricfehr3:
Clang format files relating to the joker and def files related to #136

### DIFF
--- a/source/joker_effects.c
+++ b/source/joker_effects.c
@@ -330,6 +330,7 @@ static u32 sock_and_buskin_joker_effect(
     JokerEffect** joker_effect
 );
 
+// clang-format off
 /* The index of a joker in the registry matches its ID.
  * The joker sprites are matched by ID so the position in the registry
  * determines the joker's sprite.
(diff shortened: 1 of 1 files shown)


source/joker_effects.c
```diff
@@ -1,31 +1,170 @@
...
+        return effect;                                                                                                 \
+    }
+#define SCORE_ON_EVENT_ONLY(restricted_event, checked_event, effect)                                                   \
+    if (checked_event != restricted_event)                                                                             \
+    {                                                                                                                  \
+        return effect;                                                                                                 \
+    }
 
 // Joker Effect functions
+static JokerEffect joker_effect_noop(Joker* joker, Card* scored_card, enum JokerEvent joker_event);
```
ricfehr3: Not going to prefix the `s_` here. I don't see it adding anything to make this file more readable.

MeirGavish: Yeah, makes sense

source/joker.c
```diff
@@ -1,38 +1,35 @@
...
 #define DEF_JOKER_GFX(idx) joker_gfx##idx##Tiles,
 #include "../include/def_joker_gfx_table.h"
 #undef DEF_JOKER_GFX
 };
-static const unsigned short *joker_gfxPal[] = 
-{
+
+static const unsigned short* joker_gfxPal[] = {
 #define DEF_JOKER_GFX(idx) joker_gfx##idx##Pal,
 #include "../include/def_joker_gfx_table.h"
```
```suggestion
#include "def_joker_gfx_table.h"
```

Since I noticed this now and we're already changing things in this file

source/joker.c
```diff
@@ -395,3 +335,72 @@ int joker_get_random_rarity()
...
+{
+    return (get_joker_registry_size() + NUM_JOKERS_PER_SPRITESHEET - 1) / NUM_JOKERS_PER_SPRITESHEET;
+}
+
+static int s_joker_get_spritesheet_idx(u8 joker_id)
+{
+    return joker_id / NUM_JOKERS_PER_SPRITESHEET;
+}
+
+// TODO: This should be generalized so any sprite can have dynamic swapping
```
```suggestion
```

This TODO can be removed, all the other sprites share a palette per category so only the jokers need palette swapping. It's a bit crude to have a non-generic sprite management system but there's no reason to do it differently now.

source/joker_effects.c
```diff
@@ -1017,10 +1145,11 @@ static JokerEffect sock_and_buskin_joker_effect(Joker *joker, Card *scored_card,
         case JOKER_EVENT_ON_HAND_PLAYED:
             *p_last_retriggered_face_index = UNDEFINED;
             break;
-        
+
         case JOKER_EVENT_ON_CARD_SCORED_END:
             // Works the same way as Dusk, but for face cards
-            effect.retrigger = ((*p_last_retriggered_face_index < get_scored_card_index()) && card_is_face(scored_card));
+            effect.retrigger =
+                ((*p_last_retriggered_face_index < get_scored_card_index()) && card_is_face(scored_card));
```
```suggestion
                ((*p_last_retriggered_face_index < get_scored_card_index()) && 
                 card_is_face(scored_card));
```
Maybe break it further

source/joker_effects.c
```diff
@@ -277,22 +416,23 @@ static JokerEffect joker_stencil_effect(Joker *joker, Card *scored_card, enum Jo
     // +1 xmult per empty joker slot...
     int num_jokers = list_get_len(jokers);
 
-    effect.xmult = (MAX_JOKERS_HELD_SIZE) - num_jokers;
+    effect.xmult = (MAX_JOKERS_HELD_SIZE)-num_jokers;
```
MeirGavish: We can't have spaces between operators?

source/joker.c
```diff
@@ -198,57 +137,52 @@ JokerObject *joker_object_new(Joker *joker)
...
+    memcpy32(&tile_mem[4][tile_index],
+             &joker_gfxTiles[joker_spritesheet_idx][joker_idx * TILE_SIZE * JOKER_SPRITE_OFFSET],
+             TILE_SIZE * JOKER_SPRITE_OFFSET);
+
+    sprite_object_set_sprite(joker_object->sprite_object,
+                             sprite_new(ATTR0_SQUARE | ATTR0_4BPP | ATTR0_AFF,
+                                        ATTR1_SIZE_32,
+                                        tile_index,
+                                        joker_pb,
+                                        JOKER_STARTING_LAYER + layer));
```
MeirGavish: I think I have another comment about this in some other pending PR - in general I'm fine with both styles but sometimes I prefer the original one (maybe with  function parenthesis in the same line as the call to clarify it's a function call) here since it avoids pushing all the arguments all the way to the write in case of long function names.

ricfehr3: Gonna resolve this, it should be good now with the #259 pr.

ricfehr3:
@MeirGavish Okay! So that took a while, but this should be the last PR on the clang formatting. Documentation is another story but I'd like to code again for a bit so that will not be my immediate focus. 

I went ahead and threw on an extra couple of commits to get the last files (main.c & selection_grid.h). We can totally make a new PR for them, but it seemed easy to add to this to complete it. 

I appreciate the extra set of eyes and reviews through this journey, haha.

source/main.c
```diff
@@ -94,14 +103,14 @@ int main()
...
     {
         VBlankIntrWait();
         mmFrame();
-		key_poll();
+        key_poll();
         update();
         draw();
     }
 
-	return 0;
```
MeirGavish, line 106: Nice to weed out those sneaky tabs

include/joker_gfx.h
```diff
@@ -1,3 +1,4 @@
+// clang-format off
```
MeirGavish, line 1: Why did you need to turn it off for this file? I don't mind, just wondering.

include/selection_grid.h
```diff
@@ -10,18 +10,25 @@ typedef struct SelectionGridRow SelectionGridRow;
...
+    SelectionGrid* selection_grid,
+    int row_idx,
+    const Selection* prev_selection,
+    const Selection* new_selection
+);
 typedef int (*RowGetSizeFunc)();
 // Called for any non-directional key hit
-// The key will not be passed, the function will have to check key_hit() etc. for the key it wants to check
+// The key will not be passed, the function will have to check key_hit() etc. for the key it wants
+// to check
```
MeirGavish, line 24: Github doesn't let me make this a suggestion you can commit but we missed this comment on #258 so if you could just change it to the following I will appreciate it

```
// Called for any non-directional key transit (either hit down or release up)
// The key itself and hit/release event will not be passed,
// the function will have to check key_hit(), key_released() etc. for the key it wants to check
```

source/joker_effects.c
```diff
@@ -1,33 +1,438 @@
...
+    {UNCOMMON_JOKER, 8, fibonnaci_joker_effect}, // 50
+    {
+     UNCOMMON_JOKER, 6,
+     seltzer_joker_effect, }, // 51
+// The following jokers don't have sprites yet,
+// uncomment them when their sprites are added.
+#if 0
+    { COMMON_JOKER,   5, shoot_the_moon_joker_effect,   },
+#endif
+};
```
MeirGavish: I prefer to keep this aligned too... not that happy about turning off formatting for so many areas but I guess it's better, we'll have to just manually make sure the lines here don't become too long. Also I don't understand why the last line for Seltzer is broken, it shouldn't be.

include/selection_grid.h
ricfehr3: The order of the headers gets funny, it goes "0->1->10->11..", alphabetizing them in a non-intuitive way.

source/joker_effects.c
ricfehr3: I agree, I turned clang format off. The original was far more aesthetic.

MeirGavish: @ricfehr3 I still don't see the change here, did you push your changes?

ricfehr3: @MeirGavish Whoops, I pushed it here instead of my fork. I fixed it and deleted the branch in this repo.

source/joker_effects.c
MeirGavish: @ricfehr3 Just wondering here, why does clang-format not accept the spaces here while it does in the other cases?

PR #258, opened 2025-11-24
Add swapping order of Owned Jokers

MathisMartin31:
@MeirGavish This PR is the first step towards solving issue #158.

How this works:

- Card Selection key has been moved from pressing A to releasing A
- While holding A, hit Left/Right on the D-pad to move the held card around
- If we hit Left/Right too fast after pressing A, we will select the card before changing focus to the next one.
- If we end up in the above situation, everything will work as if we had released A before hitting Left/Right. Meaning, pressing Left/Right again will not move the card we landed on, and releasing A will not select it
- Card swapping doesn't wrap for now, will figure it out later if it's useful

Here is an example of the swapping working:

https://github.com/user-attachments/assets/a916f866-c356-43b6-8f03-09dfe8de4113

A few caveats:

- ~~Due to the card selection key now being releasing A, pressing A and Left/Right in quick succession to quickly select cards can be hit or miss and it doesn't feel really good (I failed to select the 7 in the above video, you can see a brief pause when I tried to press A). Maybe swapping cards shouldn't rely on the A button but I don't know what other button to use~~ I have added some checks and safeties to try to mitigate this issue but it's not perfect imo.
- the `hand` array has **not** been modified to use the `List` structure. It was a bit complex and there were a few limitations here and there that made it finicky. I couldn't get it to work properly

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -387,14 +387,13 @@ static StateInfo state_info[] = {
 #include "../include/def_state_info_table.h"
 #undef DEF_STATE_INFO
 };
-// TODO: Properly format this and make clang-format happy with it.
-// clang-format off
+
 SelectionGridRow shop_selection_rows[] = {
     {0, jokers_sel_row_get_size,  jokers_sel_row_on_selection_changed,  jokers_sel_row_on_key_transit },
     {1, shop_top_row_get_size,    shop_top_row_on_selection_changed,    shop_top_row_on_key_transit   },
     {2, shop_reroll_row_get_size, shop_reroll_row_on_selection_changed, shop_reroll_row_on_key_transit}
 };
-// clang-format on
+
 static const Selection SHOP_INIT_SEL = {-1, 1};
 
 SelectionGrid shop_selection_grid = {
--- a/.clang-format
+++ b/.clang-format
@@ -52,3 +52,5 @@ SpacesInLineCommentPrefix:
 
 PenaltyReturnTypeOnItsOwnLine: 99999999
 AlignEscapedNewlines: Left
+
+BracedInitializerIndentWidth: 0


ricfehr3:
This is awesome. Another great feature to add in for sure. I think some of those input headaches could be solved by adding a state for the hand that the inputs set rather than relying on the `KEY_IS_DOWN` to figure that state out. 

My guess for these bugs is since the input is polled once per frame, and you're waiting between frames to figure out if the key is release. So one frame for key_down, then everything else done in a frame, then maybe the logic will properly get to the key_release code (if selection_y == 0 still). Having the key combos adjust a state and instead that state's update function is ran per frame. That way, you have clean logic for the `if(KEY_RELEASED()) -> play_card()` process. 

Don't stress about not using the `List` struct. It doesn't **need** to be used. From what I see, it makes sense to do what you have here right now. That MAX_HAND_SIZE can go as high as we need. The only issue is that an insertion into the hand is more difficult this way, but until we need that, we are good.

MathisMartin31:
Update: Jokers can now be swapped! ^^

In the video below I'm testing if Jokers are swapped correctly, and what happens if I try to hold A and swap to another row (nothing, as intended)

https://github.com/user-attachments/assets/73b5bf0b-bee3-4787-be49-cc757aba658d

MeirGavish:
@MathisMartin31 
Awesome! It seems like there's an issue with the displayed sell price when swapping, when you swap one that sells for $2 with the basic joker, it shows $1.
Edit: Semi-related to this bug, I have another idea that might make it look a little better and give the user some feedback - maybe we can hide the price while the "A" button is held and show it again once it's released?

source/game.c
```diff
@@ -705,34 +710,47 @@ void sort_hand_by_rank()
...
+            {
+                break;
+            }
+
+            // if there is one, shift it and all the cards that follow
+            // and decrement i to check that card again
+            for (int j = 0; j <= hand_top - non_null_card_idx + 1; j++)
+            {
+                hand[i + j] = hand[non_null_card_idx + j];
+            }
```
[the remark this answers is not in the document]
MathisMartin31: I left that comment by mistake, there was no need to decrement i.

source/game.c
```diff
@@ -1564,6 +1613,20 @@ static void game_playing_process_hand_select_input()
...
+                    moving_card = true;
+                    rearrange_card_sprites(); 
+                }
+                else
+                {
+                    hand_set_focus(selection_x + 1);
+                }
+            }
+
             hand_set_focus(selection_x - 1);
```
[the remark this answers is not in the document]
MathisMartin31: Yeah so the check of whether we should wrap around with/without swapping cards was a bit tricky and the easiest way of doing it I came up with was to compensate for the focus change beforehand. I'll take another look at it because I didn't think about it once I wrote that and in hindsight it looks dumb as hell

source/game.c
```diff
@@ -705,34 +710,47 @@ void sort_hand_by_rank()
...
+
+            // exit loop if there are not non NULL cards left/there are no more sprites to destroy
+            if (non_null_card_idx > hand_top)
+            {
+                break;
+            }
+
+            // if there is one, shift it and all the cards that follow
+            // and decrement i to check that card again
+            for (int j = 0; j <= hand_top - non_null_card_idx + 1; j++)
```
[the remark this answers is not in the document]
MathisMartin31: About that +1... honestly I don't know.

It doesn't crash with it there, and the animation looks janky and all broken without it so I stopped thinking about it and moved on :thinking:

source/game.c
```diff
@@ -1548,11 +1580,28 @@ void game_start()
...
+                    rearrange_card_sprites();
+                }
+                else
+                {
+                    hand_set_focus(selection_x - 1);
+                }
+                
+            }
+
             hand_set_focus(selection_x + 1); // The reason why this adds 1 is because the hand is drawn from right to left. There is no particular reason for this, it's just how I did it.
```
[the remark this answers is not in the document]
MathisMartin31: This should now be fixed, and hand_set_focus is now only called when we actually need to either change or swap cards, which means we don't hear the "card selected" SFX when trying to swap a card when at the end or beginning of the hand anymore.

Same remark for all other comments from Copilot regarding the same matter

MathisMartin31:
I updated how card moving works when pressing the D-pad Left/Right and A in quick succession. A bit better now but not quite as good as before.

I added a maxmum number of frames where if you press Left/Right while holding A before the timer is over, you select/deselect the card first, then change focus without moving the card. This is how Balatro seems to be doing it where you can press A first, then quickly change card and it would act as if you pressed the A button and *then* Left/Right.

I'm still not satisfied yet, will revisit later but if anyone has any suggestions I'm all ears ^^'

MathisMartin31:
Converted this to draft for now. I thought I could quickly fix the issues that ocurred when quickly selecting and changing cards but the more I try to fix them, the worse the game feels so I'll take some time to start the fix from scratch

MathisMartin31:
> @MathisMartin31 Awesome! It seems like there's an issue with the displayed sell price when swapping, when you swap one that sells for $2 with the basic joker, it shows $1. Edit: Semi-related to this bug, I have another idea that might make it look a little better and give the user some feedback - maybe we can hide the price while the "A" button is held and show it again once it's released?

@MeirGavish It is now working like this, but I had to change ever so slightly how the `selection_grid` handles input by having "on key hit" methods also be called when a button is released. This doesn't appear to have broken anything but this was my only option to erase the price when hitting A and printing it when releasing.

MathisMartin31:
I did some more tweaking. Will update the spec in the PR description to clarify what I've done. Let me know how the game feels now :)

MeirGavish:
I'm going to review this next, it will take me a bit.

source/game.c
```diff
@@ -681,6 +681,13 @@ void set_seed(int seed)
     srand(rng_seed);
 }
 
+void swap_cards_in_hand(int a, int b)
+{
+    CardObject* temp = hand[a];
+    hand[a] = hand[b];
+    hand[b] = temp;
+}
```
```suggestion
static void swap_cards_in_hand(int idx_a, int idx_b)
{
    CardObject* temp = hand[idx_a];
    hand[idx_a] = hand[idx_b];
    hand[idx_b] = temp;
}
```

Rename the parameters to signify those are hand indices and add static to the function since it's not used outside this file.

Also making it static should add the prefix `s_` to the name but since this is game.c this is not a must as this function could be extracted to a non-static in a hand file in the future.

MeirGavish:
@MathisMartin31 
So, before you merge this with main... I think it would be better if this PR was split into two separate PRs for the hand and the jokers, can you do that? It would make this much easier to review and merge. It should have been separate issues to begin with.
Edit: Either way, I think I will look at joker swapping first.

source/game.c
```diff
@@ -3138,13 +3250,20 @@ static void jokers_sel_row_on_selection_changed(SelectionGrid *selection_grid,
...
+    bool swapping = key_is_down(SELECT_CARD) && new_selection->y == row_idx && prev_selection->y == row_idx;
+
     if (prev_selection->y == row_idx)
     {
         JokerObject* joker_object = (JokerObject*)list_get_at_idx(&_owned_jokers_list, prev_selection->x);
         if(joker_object != NULL)
         {
             erase_price_under_sprite_object(joker_object->sprite_object);
-            sprite_object_set_focus(joker_object->sprite_object, false);
+            // keep focus on current Joker if swapping
```
```suggestion
            // Don't change focus from current Joker if swapping
```

I think the comment is a little clearer this way

source/selection_grid.c
```diff
@@ -75,7 +75,7 @@ void selection_grid_process_input(SelectionGrid *selection_grid)
     selection_grid_process_directional_input(selection_grid);
 
     u32 non_directional_key = KEY_ANY & ~KEY_DIR;
-    if (key_hit(non_directional_key))
+    if (key_hit(non_directional_key) || key_released(non_directional_key))
```
MeirGavish: So what this essentially does is turn this callback from `on_key_hit` to `on_key_transit` because `key_hit || key_released` is equivalent to `key_transit`.
There are 2 things you can do.

1. Add an `on_key_released` callback to the selection grid and add the logic there. I'm not sure this is the best course of action since it means updating the definitions of all rows and most will not have a key released function but a no-op instead. I don't think it's the one I prefer though it might be the more "logically correct one"?
2. Rename the callback function to `on_key_transit` and use that condition to make this behavior explicit and grounded in the function name. This will require every function to check it was called for the correct event for the key it wants. Most of them do that already to check which key was pressed... **except** for the reroll... That's a bug actually because it means that it activates for a press of any key, not just A. This bug is made worse by with this because now a reroll will be triggered both on key up and key down so twice for every press. I will open an issue for this and will probably make a quick fix for it.

source/game.c
```diff
@@ -3183,12 +3313,25 @@ void game_sell_joker(int joker_idx)
...
+    if (joker_object != NULL)
+    {
+        if (key_hit(SELECT_CARD))
+        {
+            erase_price_under_sprite_object(joker_object->sprite_object);
+        }
+        else
+        {
+            print_price_under_sprite_object(joker_object->sprite_object, joker_get_sell_value(joker_object->joker));
+        }
```
MeirGavish: This else is a problem because it assumes that the only other event is a release of the selection key while this is not necessarily true. Try holding A while pressing B (or any other) and it will restore the price. Add a check for a key release of the same button.

MathisMartin31:
> @MathisMartin31 So, before you merge this with main... I think it would be better if this PR was split into two separate PRs for the hand and the jokers, can you do that? It would make this much easier to review and merge. It should have been separate issues to begin with. Edit: Either way, I think I will look at joker swapping first.

Yeah I can totally do that. Since the Hand uses neither the `List` implementation nor the `selection_grid` stuff it's gonna be easy to separate the two

source/game.c
```diff
@@ -3153,10 +3272,21 @@ static void jokers_sel_row_on_selection_changed(SelectionGrid *selection_grid,
...
-            sprite_object_set_focus(joker_object->sprite_object, true);
-            print_price_under_sprite_object(joker_object->sprite_object, joker_get_sell_value(joker_object->joker));
+            if (!swapping)
+            {
+                sprite_object_set_focus(joker_object->sprite_object, true);
+            }
+            if (!key_is_down(SELECT_CARD))
+            {
+                print_price_under_sprite_object(joker_object->sprite_object, joker_get_sell_value(joker_object->joker));
+            }
```
MeirGavish, line 3708:
```suggestion
            if (!key_is_down(SELECT_CARD))
            {
                print_price_under_sprite_object(joker_object->sprite_object, joker_get_sell_value(joker_object->joker));
            }
```

Why not put it together with the above swapping condition? It will mean the price will be displayed for the case of entering the joker row while the key is held will display the price despite being in swapping mode but I guess it's not so bad since you technically can sell the joker while swapping and the call to `erase_price_under_sprite_object();` executed even when swapping in the `prev_selection` condition means the price will be erased when swapping. Maybe it's good to add a comment there explaining it's needed for this case... I'm starting to think my request to erase the price while swapping made the code a little messy...

MeirGavish (changes requested):
I looked at the joker swapping code and made a few comments. Good job, it's good to finally have this feature! There is also a reroll bug that I will probably make a quick fix for but it needs to be fixed either with this PR or before it because it makes the bug worse.

source/game.c
MeirGavish: I did look at this one before focusing on the jokers.

MeirGavish:
@MathisMartin31 You don't have to close this PR, you can keep it focused on jokers and take the card swapping to a new PR.

MathisMartin31:
Yeah it closed automatically again... I pushed commits so I'll reopen.

This PR will be dedicated to the Jokers, and I'll open a new one for the Cards in Hand

source/selection_grid.c
MathisMartin31: Oooooh I see what you mean. I see you opened a PR for this, maybe I can integrate this and the rename to on_key_transit into this PR? I think this may be easier to merge this way

source/game.c
MathisMartin31: I addressed this in its own branch, I'll put up a PR when I'm done addressing issues here

source/game.c
MathisMartin31: I added your change to this PR

source/selection_grid.c
MathisMartin31: I changed the names and added the check for the shop reroll, I played with it a bit and everything should be fine

source/game.c
MathisMartin31: I think adding a comment should be fine, it's a bit convoluted but I don't think it's too messy yet

MathisMartin31:
Aaaaargh I'm doing the clang-format thing but the check keeps failing I don't know whyyyyy T_T

MathisMartin31:
I don't know how to format that array, it went fine for the other PR whyyyyy T_T

MeirGavish:
> I don't know how to format that array, it went fine for the other PR whyyyyy T_T

@MathisMartin31 I'll take a look.

@ricfehr3 I wonder if there's a way to make clang-format give a more informative error message? Right now it doesn't give you any details to why a certain block fails. Also, I want to open an issue to separate the clang-format to its own workflow and not skip the build if other workflows fail. If one workflow fails you should be able to see if others succeed or fail.

MathisMartin31:
> I'll take a look

Thaaaanks ^^

source/game.c
```diff
@@ -389,9 +389,9 @@ static StateInfo state_info[] = {
 };
 
 SelectionGridRow shop_selection_rows[] = {
-    {0, jokers_sel_row_get_size,  jokers_sel_row_on_selection_changed,  jokers_sel_row_on_key_hit },
-    {1, shop_top_row_get_size,    shop_top_row_on_selection_changed,    shop_top_row_on_key_hit   },
-    {2, shop_reroll_row_get_size, shop_reroll_row_on_selection_changed, shop_reroll_row_on_key_hit}
+    {0, jokers_sel_row_get_size,  jokers_sel_row_on_selection_changed,  jokers_sel_row_on_key_transit },
+    {1, shop_top_row_get_size,    shop_top_row_on_selection_changed,    shop_top_row_on_key_transit   },
+    {2, shop_reroll_row_get_size, shop_reroll_row_on_selection_changed, shop_reroll_row_on_key_transit}
 };
```
MeirGavish, line 395: Relating to [this comment](https://github.com/cellos51/balatro-gba/pull/258#issuecomment-3614393416)
> I don't know how to format that array, it went fine for the other PR whyyyyy T_T

@MathisMartin31 So the reason is that now the line length exceeds 100 columns (if only by a little bit). 

I tried to clang auto-format it and the output was
```
SelectionGridRow shop_selection_rows[] = {
    {0,
     jokers_sel_row_get_size,  jokers_sel_row_on_selection_changed,
     jokers_sel_row_on_key_transit                                                              },
    {1, shop_top_row_get_size, shop_top_row_on_selection_changed,    shop_top_row_on_key_transit},
    {2,
     shop_reroll_row_get_size, shop_reroll_row_on_selection_changed,
     shop_reroll_row_on_key_transit                                                             }
};
```
which is horrible, it's very hard to tell which callback is for which row.
I'd want to be able to break it like this
```
SelectionGridRow shop_selection_rows[] = {
    {0, jokers_sel_row_get_size,
        jokers_sel_row_on_selection_changed,
        jokers_sel_row_on_key_transit},
    {1, shop_top_row_get_size, 
        shop_top_row_on_selection_changed,
        shop_top_row_on_key_transit},
    {2, shop_reroll_row_get_size, 
        shop_reroll_row_on_selection_changed,
        shop_reroll_row_on_key_transit}
};
```
which is compact,
or like this 
```
SelectionGridRow shop_selection_rows[] = {
    {
        0, 
        jokers_sel_row_get_size,
        jokers_sel_row_on_selection_changed,
        jokers_sel_row_on_key_transit
    },
    {
        1, 
        shop_top_row_get_size, 
        shop_top_row_on_selection_changed,
        shop_top_row_on_key_transit
    },
    {
        2, 
        shop_reroll_row_get_size, 
        shop_reroll_row_on_selection_changed,
        shop_reroll_row_on_key_transit
    }
};
```
which I would expect clang-format to more easily accept.

But the format doesn't accept any of them. What I will do is quickly commit a change to turn clang-format off here so the other workflows can run and @ricfehr3 if you can provide input and help with that it will be greatly appreciated. I'm also okay with leaving the clang format off for this section and doing its formatting manually.

source/game.c
```diff
@@ -389,9 +389,9 @@ static StateInfo state_info[] = {
...
```
```suggestion
// TODO: Properly format this and make clang-format happy with it.
// clang-format off
```
Related to [this comment](https://github.com/cellos51/balatro-gba/pull/258/files#r2591982742)

source/game.c
```diff
@@ -389,9 +389,9 @@ static StateInfo state_info[] = {
...
 
 SelectionGridRow shop_selection_rows[] = {
-    {0, jokers_sel_row_get_size,  jokers_sel_row_on_selection_changed,  jokers_sel_row_on_key_hit },
-    {1, shop_top_row_get_size,    shop_top_row_on_selection_changed,    shop_top_row_on_key_hit   },
-    {2, shop_reroll_row_get_size, shop_reroll_row_on_selection_changed, shop_reroll_row_on_key_hit}
+    {0, jokers_sel_row_get_size,  jokers_sel_row_on_selection_changed,  jokers_sel_row_on_key_transit },
+    {1, shop_top_row_get_size,    shop_top_row_on_selection_changed,    shop_top_row_on_key_transit   },
+    {2, shop_reroll_row_get_size, shop_reroll_row_on_selection_changed, shop_reroll_row_on_key_transit}
 };
```
```suggestion
// clang-format on
```

Related to [this comment](url)

source/selection_grid.c
MeirGavish: @MathisMartin31 Oh sorry, I missed this comment and merged it...

source/selection_grid.c
```diff
@@ -60,26 +59,27 @@ void selection_grid_move_selection_vert(SelectionGrid *selection_grid, int direc
...
     selection_grid_process_directional_input(selection_grid);
 
     u32 non_directional_key = KEY_ANY & ~KEY_DIR;
-    if (key_hit(non_directional_key))
+    if (key_transit(non_directional_key))
     {
-        Selection* selection = &selection_grid->selection; // To make the next line shorter and more readable
-        selection_grid->rows[selection->y].on_key_hit(selection_grid, selection);
+        Selection* selection =
+            &selection_grid->selection; // To make the next line shorter and more readable
```
MeirGavish:
```suggestion
        // To make the next line shorter and more readable
        Selection* selection = &selection_grid->selection; 
```

Break like this with comment and code in separate lines

source/game.c
```diff
@@ -3670,14 +3671,22 @@ static void jokers_sel_row_on_selection_changed(
...
         if (joker_object != NULL)
         {
             erase_price_under_sprite_object(joker_object->sprite_object);
-            sprite_object_set_focus(joker_object->sprite_object, false);
+            // Don't change focus from current Joker if swapping
+            if (!swapping)
+            {
+                sprite_object_set_focus(joker_object->sprite_object, false);
+            }
         }
```
MeirGavish, line 3686: Since you added a comment in the entering case maybe it's better to move the swapping condition to the top if statement `if (prev_selection->y == row_idx && !swapping)` since in the swapping case there's no price to erase too. Then the swapping comment here will move to the entering case too.

source/selection_grid.c
MeirGavish: If you want I can revert that PR.

MathisMartin31: Don't worry I'll revert it, it shouldn't create conflicts with mine

source/game.c
ricfehr3: I looked into it and there was a miracle configuration:

If you add `BracedInitializerIndentWidth: 0` in `.clang-format` and remove the `// clang-format on/off` comments, it fixes that one portion and only that one portion. So you can safely add it to this PR and it won't effect other files either. 

That configuration keeps it looking like the original. The allman style braces are not possible, but it looks far nicer the old way than the nastiness it was trying to push. 

@MathisMartin31 Thanks for bearing with the `clang-format` stuff. I know it's a bit annoying. Would you mind sharing what editor and os you use? I'm just curious to see what level of integration `clang-format` has. I personally do it manually, and I don't want to tell people to just do that, lol.

source/game.c
```diff
@@ -3687,13 +3696,31 @@ static void jokers_sel_row_on_selection_changed(
...
-            print_price_under_sprite_object(
-                joker_object->sprite_object,
-                joker_get_sell_value(joker_object->joker)
-            );
+            if (!swapping)
+            {
+                sprite_object_set_focus(joker_object->sprite_object, true);
+            }
+            // If we land on this row while the A button is being held, we are in swapping mode
+            // This means that we need to hide the prise, whether we were already
```
ricfehr3: nitpik: typo "prise" -> "price"

ricfehr3 (changes requested):
Thank you @MathisMartin31 . Great work. I personally appreciate you adding features that add to the style and gameplay, and this is another good one. I left a nitpik and a more important comment about a `.clang-format` change. I am also looking into making a PR to fix #281 , to hopefully make the clang-format stuff easier.

source/game.c
MathisMartin31: No problem ^^

I'm using VSCodium on Manjaro right now, so I have the Format Document command that should use the `.clang-format` file in the repo.

That's what I've been doing, but for some reason it was complaining even though it *should* be using the file so idk what went wrong

MathisMartin31: And I think your suggestion fixed the issue. I added it to the clang format, removed the on/off exception comment and ram the Format Document command again, and the array was left untouched ^^

Thanks :)

MeirGavish: @ricfehr3 Interesting, thanks a bunch! Why is it suddenly okay with more than 100 columns then?

MathisMartin31: I have no idea ><

MeirGavish: @ricfehr3 Also, I'm using VS Code on Windows 11 and I use clang-format --dry-run manually to see the errors because I didn't yet manage to make the IDE display them but I did manage to make alt+shift+f auto-format the file.

ricfehr3: I really don't know why. In full honesty I think it's undefined behavior. Nothing seems to connect changing this into some sort of override of the column limit. It was just one of the few parameters that had anything to do with initializer list (which clang-format seems to not work well with). I say we keep it until it fails us? 

And it's interesting to see that clang-format isn't auto running for y'all. I would have thought it would just run with VsCode. Maybe I should document this in the wiki, explaining how to do it manually too.

MeirGavish:
@ricfehr3 Can I merge this?

ricfehr3:
> @ricfehr3 Can I merge this?

Sorry, my mistake I meant to approve earlier.

MeirGavish:
> Sorry, my mistake I meant to approve earlier.

No problem! I hope I didn't wake you up for this. 😅

ricfehr3:
> > Sorry, my mistake I meant to approve earlier.
> 
> No problem! I hope I didn't wake you up for this. 😅

Haha, don't worry. I was up already. Now, should I be merging at 3am is another story, lol.

PR #259, opened 2025-11-24
Clang format  and refactor Game.h/c

ricfehr3:
Refactored `game.c` quite a bit. Mostly extracting all the file scoped functions into `static` functions. No real effect on the size of the ROM either, the only real effect is aesthetic.

There is no code change here. Just formatting and moving stuff around.

### DIFF
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -215,8 +215,8 @@ void sprite_object_update(SpriteObject* sprite_object)
     if (sprite_object->vrotation < epsilon && sprite_object->vrotation > -epsilon)
     {
         sprite_object->vrotation = 0;
-        sprite_object->rotation =
-            sprite_object->trotation; // Set the rotation to the target rotation
+        // Set the rotation to the target rotation
+        sprite_object->rotation = sprite_object->trotation;
     }
     else
     {
@@ -224,12 +224,13 @@ void sprite_object_update(SpriteObject* sprite_object)
         sprite_object->rotation += sprite_object->vrotation;
     }
 
+    // Apply rotation and scale to the sprite
     obj_aff_rotscale(
         sprite_object->sprite->aff,
         sprite_object->scale,
         sprite_object->scale,
         -sprite_object->vx + sprite_object->rotation
-    ); // Apply rotation and scale to the sprite
+    );
     sprite_position(sprite_object->sprite, fx2int(sprite_object->x), fx2int(sprite_object->y));
 }
 


MeirGavish:
@ricfehr3 Okay, I'm really sorry about this, I should have said this sooner, I thought I was generally okay with it but the more I'm seeing it, the more it bothers me and I just have to say it now - I don't really like beginning names with an underscore. 
I was going to say this and thought it will be a little longer before this PR comes, sorry, I should have already said this on the format issue.
This bothers me more for functions than variables, it can stay for variables.
I do think it's good to differentiate static functions and variables from others but I think the static functions should begin with `s_`, 
I also had another thought that probably should have shared - I had the idea of prefixing all our functions and constants with `gbal_`/`GBAL_` to differentiate our defined ones from library ones.
The thing about game.c though is that `static` is a bit problematic for that file - most functions there are static and many of them shouldn't be, the file needs to be separated to a files for each state and then many of these functions will need to be unstaticed so they are available to all states. I'm also pretty sure there are also several functions there that are not static and should be.

ricfehr3:
> @ricfehr3 Okay, I'm really sorry about this, I should have said this sooner, I thought I was generally okay with it but the more I'm seeing it, the more it bothers me and I just have to say it now - I don't really like beginning names with an underscore. I was going to say this and thought it will be a little longer before this PR comes, sorry, I should have already said this on the format issue. This bothers me more for functions than variables, it can stay for variables. I do think it's good to differentiate static functions and variables from others but I think the static functions should begin with `s_`, I also had another thought that probably should have shared - I had the idea of prefixing all our functions and constants with `gbal_`/`GBAL_` to differentiate our defined ones from library ones. The thing about game.c though is that `static` is a bit problematic for that file - most functions there are static and many of them shouldn't be, the file needs to be separated to a files for each state and then many of these functions will need to be unstaticed so they are available to all states. I'm also pretty sure there are also several functions there that are not static and should be.

No need to apologize, it's all chill. No stress to go through and rename stuff. I can remove all the underscores from the functions. 

I can put some thought into where to put those prefixes, especially for the pre-processor definitions. If it makes sense to put them on functions I'll also use my best judgement.

I modified `game.c` with the idea of organizing it for the eventual separation of its many parts into different files. They don't need to be `static`, really. I mostly wanted to organize the functions in a more digestible way. I can keep them as either `static` or not, as they were, if you'd like. Not sure it makes a big difference either way.

MeirGavish:
> I can put some thought into where to put those prefixes, especially for the pre-processor definitions. If it makes sense to put them on functions I'll also use my best judgement.

Do you mean my `gbal_`/`GBAL_` suggestion? I don't know if I've given it enough thought but the original thought was to put it on everything, it doesn't have to be done here, it's a lot to change. Also, I'm not sure how will it go with other prefixes - `gbal_s_` or  `s_gbal_`?
I just felt like it's relevant so I put it out there.

ricfehr3:
@MeirGavish That's definitely relevant, haha. I'm not going to hunt places down to put that prefix right now, but you're right, there is room for more discussion. As for this PR, I'm thinking we leave everything static as long as it's within the scope of the file. Once it leaves the file, that PR will properly name the functions and set scope.

I do think that the `s_` could be used for static functions instead of the `_`, but because we are planning to gut this file, I won't do that here.

source/game.c
```diff
@@ -49,190 +155,425 @@ typedef enum
...
+
+static const BG_POINT HELD_JOKERS_POS = {108, 10};
+static const BG_POINT JOKER_DISCARD_TARGET = {240, 30};
+static const BG_POINT CARD_DRAW_POS = {208, 110};
+static const BG_POINT CUR_BLIND_TOKEN_POS = {8, 18};
+static const BG_POINT CARD_DISCARD_PNT = {240, 70};
+static const BG_POINT HAND_START_POS = {120, 90};
+static const BG_POINT HAND_PLAY_POS = {120, 70};
+static const BG_POINT MAIN_MENU_ACE_T = {88, 26};
+// Naming the stage where cards return from the discard pile to the deck "undiscard"
```
MeirGavish: This comment belongs to PITCH_STEP_UNDISCARD_SFX (it's the only place I could find where "undiscard" is used), it seems to have lost its way through the refactors, since you're already moving things around, can you put it back there?

source/game.c
```diff
@@ -2844,142 +3651,59 @@ static void game_round_end_print_interest_reward(int interest_y_offset)
...
-        timer = TM_ZERO;
-    
-        obj_hide(round_end_blind_token->obj); // Hide the blind token object
-        tte_erase_rect_wrapper(BLIND_TOKEN_TEXT_RECT); // Erase the blind token text
+        tte_printf("#{P:%d,%d; cx:0x%X000}%d #{cx:0x%X000}Interest",
+                   ROUND_END_REWARD_TEXT_X,
+                   interest_y * TILE_SIZE,
+                   TTE_YELLOW_PB,
+                   _interest_reward,
+                   TTE_WHITE_PB);
```
MeirGavish: Was the old format not allowed by clang? Just wondering.

include/game.h
```diff
@@ -42,7 +42,8 @@ enum HandState
 {
     HAND_DRAW,
     HAND_SELECT,
-    HAND_SHUFFLING, // This is actually a misnomer because it's used for the deck, but it mechanically makes sense to be a state of the hand
+    // This is a misnomer because it's used for the deck, but it mechanically makes sense to be a state of the hand
+    HAND_SHUFFLING,
```
MeirGavish, line 60: Maybe add a newline here after `HAND_SELECT,` to indicate the  the comment refers to `HAND_SHUFFLING,`

source/game.c
```diff
@@ -2844,142 +3651,59 @@ static void game_round_end_print_interest_reward(int interest_y_offset)
...
 }
 
-// Shop
-#define REROLL_BASE_COST 5 // Base cost for rerolling the shop items
-static int reroll_cost = REROLL_BASE_COST;
-
-#define NEXT_ROUND_BTN_SEL_X 0
-
-#define REROLL_BTN_FRAME_PAL_IDX        7
-#define REROLL_BTN_PAL_IDX              3
```
MeirGavish, line 3008: Back in the day I put those here since they are specific to the shop... but I guess it's better to be consistent and keep this for when we split game.c into state files...

source/game.c
```diff
@@ -49,190 +155,425 @@ typedef enum
...
+
+static const SubStateActionFn _round_end_state_actions[] = {game_round_end_start,
+                                                            game_round_end_start_expand_popup,
+                                                            game_round_end_display_finished_blind,
+                                                            game_round_end_display_score_min,
+                                                            game_round_end_update_blind_reward,
+                                                            game_round_end_panel_exit,
+                                                            game_round_end_display_rewards,
+                                                            game_round_end_display_cashout,
+                                                            game_round_end_dismiss_round_end_panel};
```
```suggestion
static const SubStateActionFn _round_end_state_actions[] = {
    game_round_end_start,
    game_round_end_start_expand_popup,
    game_round_end_display_finished_blind,
    game_round_end_display_score_min,
    game_round_end_update_blind_reward,
    game_round_end_panel_exit,
    game_round_end_display_rewards,
    game_round_end_display_cashout,
    game_round_end_dismiss_round_end_panel
};
```

I'd say break this down.
Edit: I checked and clang doesn't accept it, but didn't we have other similar breaks?

source/game.c
```diff
@@ -49,190 +155,425 @@ typedef enum
...
+static void game_playing_process_hand_select_input(void);
+static void game_playing_process_input_and_state(void);
+static void game_playing_process_card_draw(void);
+static bool game_round_is_over(void);
+static void game_playing_handle_round_over(void);
+static void game_playing_discarded_cards_loop(void);
+static void card_in_hand_loop_handle_discard_and_shuffling(int card_idx,
+                                                           FIXED* hand_x,
+                                                           FIXED* hand_y,
+                                                           bool* break_loop);
```
```suggestion
static void card_in_hand_loop_handle_discard_and_shuffling(
    int card_idx,
    FIXED* hand_x,
    FIXED* hand_y,
    bool* break_loop
 );
```
This makes me think maybe this style is better since then it takes the parameters to only a single indentation level and then they are not forced all the way to the right by the length of the function name. Checked and clang doesn't accept it, can we make it?

source/game.c
```diff
@@ -49,190 +155,425 @@ typedef enum
...
+static const Rect GAME_WIN_MSG_TEXT_RECT = {112, 72, UNDEFINED, UNDEFINED};
+
+static const BG_POINT HELD_JOKERS_POS = {108, 10};
+static const BG_POINT JOKER_DISCARD_TARGET = {240, 30};
+static const BG_POINT CARD_DRAW_POS = {208, 110};
+static const BG_POINT CUR_BLIND_TOKEN_POS = {8, 18};
+static const BG_POINT CARD_DISCARD_PNT = {240, 70};
+static const BG_POINT HAND_START_POS = {120, 90};
+static const BG_POINT HAND_PLAY_POS = {120, 70};
+static const BG_POINT MAIN_MENU_ACE_T = {88, 26};
```
MeirGavish: Oh, but the rects were so nicely field-aligned... 🙁 
Can't we make clang allow it? Or at least disable clang  for this section?

MeirGavish:
Working on a review for this, the file is big so it's taking me a bit, I'll try to finish it soon.

ricfehr3:
> Working on a review for this, the file is big so it's taking me a bit, I'll try to finish it soon.

No worries at all. I'm having a chill holiday, so no rush :)

MeirGavish:
> No worries at all. I'm having a chill holiday, so no rush :) 

Happy Thanksgiving!

source/game.c
MeirGavish: Checked, nope....

MeirGavish (changes requested):
Submitting the review so you're not waiting a long time for it, I still need to give the file another pass though...

source/game.c
ricfehr3: Yeah, clang-format seems to be the most awesome, but limiting thing, haha. Still, I think it's a good addition but these tiny things are bugging me too.

source/game.c
ricfehr3: Going to disable clang-format here.

source/game.c
ricfehr3: Put it in again, I merged main back in an it looks like I started on a previous commit without that #define? Not sure, but it's here now.

source/game.c
ricfehr3: I like this. I got it working and I'll push it soon and you can decide if that works for you. It does look nicer to me.

source/game.c
ricfehr3: Looks like this was actually fixed [here](https://github.com/cellos51/balatro-gba/pull/259#discussion_r2559875012)

source/game.c
ricfehr3: Yeah, I'm thinking we keep it consistent for now. It shouldn't be too difficult to separate when we get to it. However nothing is set in stone, just let me know if you want to switch it back.

ricfehr3:
@MeirGavish This one should be good. I addressed all the comments. But there are now more changes here so understandable if you need more time, and sorry for the massive amount of changes

ricfehr3:
@MeirGavish Okay, after looking at the other clang format PRs, I'd like to merge this one first because the addition of that [request you had](https://github.com/cellos51/balatro-gba/pull/259#discussion_r2559875012) really cleaned everything up. I'd like to apply it to the other PRs, but this one contains a lot of the changes required already.

MeirGavish:
> @MeirGavish Okay, after looking at the other clang format PRs, I'd like to merge this one first because the addition of that [request you had](https://github.com/cellos51/balatro-gba/pull/259#discussion_r2559875012) really cleaned everything up. I'd like to apply it to the other PRs, but this one contains a lot of the changes required already.

@ricfehr3 But not before the truncate scores PR #269, right?
I'll try to give this another pass today.

ricfehr3:
> > @MeirGavish Okay, after looking at the other clang format PRs, I'd like to merge this one first because the addition of that [request you had](https://github.com/cellos51/balatro-gba/pull/259#discussion_r2559875012) really cleaned everything up. I'd like to apply it to the other PRs, but this one contains a lot of the changes required already.
> 
> @ricfehr3 But not before the truncate scores PR #269, right? I'll try to give this another pass today.

Yeah, that's correct. We should merge #269 before this for sure. This one should be the first of the remaining clang-formats though. And Thanks!

MeirGavish:
> > > @MeirGavish Okay, after looking at the other clang format PRs, I'd like to merge this one first because the addition of that [request you had](https://github.com/cellos51/balatro-gba/pull/259#discussion_r2559875012) really cleaned everything up. I'd like to apply it to the other PRs, but this one contains a lot of the changes required already.
> > 
> > 
> > @ricfehr3 But not before the truncate scores PR #269, right? I'll try to give this another pass today.
> 
> Yeah, that's correct. We should merge #269 before this for sure. This one should be the first of the remaining clang-formats though. And Thanks!

Thanks! Sure! I meant to do it yesterday but didn't get to it, I was pretty sleep deprived.

source/game.c
```diff
@@ -946,9 +995,10 @@ static void sort_hand_by_suit(void)
     {
         for (int b = a + 1; b <= hand_top; b++)
         {
-            if (hand[a] == NULL || (hand[b] != NULL && (hand[a]->card->suit > hand[b]->card->suit ||
-                                                        (hand[a]->card->suit == hand[b]->card->suit &&
-                                                         hand[a]->card->rank > hand[b]->card->rank))))
+            if (hand[a] == NULL ||
+                (hand[b] != NULL && (hand[a]->card->suit > hand[b]->card->suit ||
+                                     (hand[a]->card->suit == hand[b]->card->suit &&
+                                      hand[a]->card->rank > hand[b]->card->rank))))
```
MeirGavish, line 1042: Yeah, I like this better

include/game.h
```diff
@@ -47,7 +55,8 @@ enum HandState
 {
     HAND_DRAW,
     HAND_SELECT,
-    HAND_SHUFFLING, // This is actually a misnomer because it's used for the deck, but it mechanically makes sense to be a state of the hand
+    // This is actually a misnomer because it's used for the deck
+    HAND_SHUFFLING,
```
MeirGavish, line 60:
```suggestion
    // This is actually a misnomer because it's used for the deck
    // but it mechanically makes sense to be a state of the hand
    HAND_SHUFFLING,
```

Keep the full comment, I think it's important to understand this.

source/game.c
```diff
@@ -48,145 +150,392 @@ typedef enum
...
+static int jokers_sel_row_get_size(void);
+static void game_shop_reroll(int* reroll_cost);
+static void game_shop_create_items(void);
+static int game_shop_get_rand_available_joker_id(void);
+static void erase_price_under_sprite_object(SpriteObject* sprite_object);
+static void print_price_under_sprite_object(SpriteObject* sprite_object, int price);
+static void game_round_end_print_interest_reward(int interest_y_offset);
+static void game_round_end_print_hand_reward(int hand_y_offset);
+static void game_round_end_print_separator_ellipsis(void);
+static void game_round_end_extend_black_panel_down(int black_panel_bottom);
```
MeirGavish, line 271: Those are a lot of forward declarations here. Maybe improve the grouping a bit and add spaces between sections. I'm also okay with removing some of them like the ones that are specifically subdivisions of bigger functions and won't be called elsewhere.

source/game.c
```diff
@@ -48,145 +150,392 @@ typedef enum
...
-typedef void (*SubStateActionFn)();
-
 // Used as a No Operation for game states that have no init and/or exit function.
-// ricfehr3 did the work of determining whether a noop or a NULL check was more 
+// ricfehr3 did the work of determining whether a noop or a NULL check was more
 // efficient. Well, this is the answer.
 // Thanks!
 // https://github.com/cellos51/balatro-gba/issues/137#issuecomment-3322485129
-static void _noop() {}
+static void noop()
```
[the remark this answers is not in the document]
MeirGavish: I actually didn't know about this difference. Passing arguments to parameters not defined in the signature seems like such a bad practice.

MeirGavish:
I requested a review from Copilot even though this is just a refactor just to make sure there isn't something I am missing.

MeirGavish (changes requested):
Looks good, just restore a deleted comment, and maybe add a bit more organization to the forward declarations if possible (I guess that's something I missed in the previous review)

source/game.c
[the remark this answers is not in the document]
ricfehr3: Yeah, this is a niche one. I did add it to the code style guide, but I haven't been enforcing it 100%. I'm considering removing it. It really isn't applicable. While it is correctly stated that the standard says that, we are supporting one compiler, which does catch this error. Buuuuuut, after some testing I think we should stick to it, it's a c99 standard. I'm not sure what we auto default too, but supporting c99 is probably good. Maybe we should add a standard in the makefile to ensure it's right too? 

That all being said, adding this suggestion.

source/sprite.c
```diff
@@ -215,18 +215,21 @@ void sprite_object_update(SpriteObject* sprite_object)
...
-    obj_aff_rotscale(sprite_object->sprite->aff,
-                     sprite_object->scale,
-                     sprite_object->scale,
-                     -sprite_object->vx + sprite_object->rotation); // Apply rotation and scale to the sprite
+    obj_aff_rotscale(
+        sprite_object->sprite->aff,
+        sprite_object->scale,
+        sprite_object->scale,
+        -sprite_object->vx + sprite_object->rotation
+    ); // Apply rotation and scale to the sprite
```
ricfehr3: fix comment

source/sprite.c
```diff
@@ -215,18 +215,21 @@ void sprite_object_update(SpriteObject* sprite_object)
     if (sprite_object->vrotation < epsilon && sprite_object->vrotation > -epsilon)
     {
         sprite_object->vrotation = 0;
-        sprite_object->rotation = sprite_object->trotation; // Set the rotation to the target rotation
+        sprite_object->rotation =
+            sprite_object->trotation; // Set the rotation to the target rotation
```
ricfehr3: fix comment

source/game.c
MeirGavish: Yeah, it's okay

source/game.c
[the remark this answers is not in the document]
MeirGavish: @ricfehr3 So it seems like calling a function that doesn't have void but just empty parentheses with arguments does generate an error for C23 but not for other standards including C99.
Having learned about this, I now think it's general good practice to have void and make sure this is protected no matter the standard.
From local testing it seems like gcc defaults to C23 but I'm not 100% sure. 
I never noticed we don't have a standard in the makefile, I suppose it makes sense to add one so the project is explicit about which one it conforms to in case a new standard comes out that might break compatibility. If we add a standard I'm not sure which one, but my gut feeling tells me to have the latest so we support all the latest features and conform to the latest practices.

Edit: I opened issue #275 for this.

Issue #263, opened 2025-11-26
Remove Triboulet from the Code


MeirGavish:
Remove Triboulet from the code since we have enough jokers as it is and it's better to not keep maintaining code for an inactive joker. Also I'd rather not change how Balatro works just so we can have a Legendary joker in the game so I don't plan to activate it any time soon.
PR that added Triboulet - #146
Also Triboulet doesn't have art - Closed issue for the art - #190

PR #266, opened 2025-11-29
Remove triboulet from code

MathisMartin31:
Related to issue #263, remove unused code for the Triboulet legendary Joker that is unlikely to be in the game for the foreseeable future


MeirGavish (approved):
I actually thought to do this after #257 is merged but might as well.

PR #268, opened 2025-11-30
Replace bubble sort with insertion sort in sort_cards()

tygyh:
A less diabolical hand sorting.

### DIFF
--- a/source/card.c
+++ b/source/card.c
@@ -134,6 +134,8 @@ void card_object_set_sprite(CardObject* card_object, int layer)
         &card_tiles[_card_sprite_lut[card_object->card->suit][card_object->card->rank] * TILE_SIZE],
         TILE_SIZE * CARD_SPRITE_OFFSET
     );
+    // Create a new sprite with the specified layer. Since sprite layers are tied to
+    // OAM indices which can't be swapped, sprites must be recreated when z-order changes.
     Sprite* sprite = sprite_new(
         ATTR0_SQUARE | ATTR0_4BPP | ATTR0_AFF,
         ATTR1_SIZE_32,
--- a/source/hand.c
+++ b/source/hand.c
@@ -167,38 +167,70 @@ void swap_cards_in_hand(int idx_a, int idx_b)
     hand.cards[idx_b] = temp;
 }
 
-static inline void sort_hand_by_suit(void)
+// Compare two cards for sorting by suit (primary) and rank (secondary)
+// Returns true if card_a should come before card_b
+static bool card_compare_by_suit(void* a, void* b)
 {
-    for (int idx_a = 0; idx_a < hand.hand_top; idx_a++)
+    CardObject* card_a = (CardObject*)a;
+    CardObject* card_b = (CardObject*)b;
+
+    if (card_a == NULL)
     {
-        for (int idx_b = idx_a + 1; idx_b <= hand.hand_top; idx_b++)
+        return false;
+    }
+    if (card_b == NULL)
+    {
+        return true;
+    }
+    if (card_a->card->suit != card_b->card->suit)
+        return card_a->card->suit < card_b->card->suit;
+    return card_a->card->rank < card_b->card->rank;
+}
+
+// Compare two cards for sorting by rank only
+// Returns true if card_a should come before card_b
+static bool card_compare_by_rank(void* a, void* b)
+{
+    CardObject* card_a = (CardObject*)a;
+    CardObject* card_b = (CardObject*)b;
+
+    if (card_a == NULL)
+    {
+        return false;
+    }
+    if (card_b == NULL)
+    {
+        return true;
+    }
+    return card_a->card->rank < card_b->card->rank;
+}
+
+static void insertion_sort(SortArgs args)
+{
+    for (int i = 1; i < args.size; i++)
+    {
+        void* key = args.array[i];
+        int j;
+
+        // Shift elements that don't satisfy the comparison
+        for (j = i - 1; j >= 0 && !args.compare(args.array[j], key); j--)
         {
-            if (hand.cards[idx_a] == NULL ||
-                (hand.cards[idx_b] != NULL &&
-                 (hand.cards[idx_a]->card->suit > hand.cards[idx_b]->card->suit ||
-                  (hand.cards[idx_a]->card->suit == hand.cards[idx_b]->card->suit &&
-                   hand.cards[idx_a]->card->rank > hand.cards[idx_b]->card->rank))))
-            {
-                swap_cards_in_hand(idx_a, idx_b);
-            }
+            args.array[j + 1] = args.array[j];
         }
+        args.array[j + 1] = key;
     }
 }
 
-static inline void sort_hand_by_rank(void)
+static void sort_hand_by_suit(void)
 {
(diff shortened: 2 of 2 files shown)


source/game.c
```diff
@@ -717,38 +717,57 @@ void set_seed(int seed)
...
+{
+    if (card_a == NULL) return false;
+    if (card_b == NULL) return true;
+    return card_a->card->rank < card_b->card->rank;
 }
 
-void sort_hand_by_rank()
+// Insertion sort implementation for the hand array
+// Using insertion sort because:
+// 1. Hand size is small (max 16 cards)
```
```suggestion
// 1. Hand size is small
```

This is not correct, the number of cards in hand is currently maxed at 8 but we plan to remove or increase that bound, the number should not appear here.

source/game.c
```diff
@@ -760,8 +779,10 @@ void sort_cards()
         sort_hand_by_rank();
     }
 
-    // Update the sprites in the hand by destroying them and creating new ones in the correct order
-    // (This is feels like a diabolical solution but like literally how else would you do this)
+    // Recreate sprites to update z-order based on new hand positions.
+    // Each card's sprite layer must match its position in the hand array
+    // for proper overlapping visuals. Since sprite layers are tied to
+    // OAM indices which can't be swapped, we destroy and recreate them.
```
MeirGavish: Is this really the reason this is done? If so that's  great to document but just make sure. Also since the recreation was moved to `card_object_set_sprite()` the comment here should just say "reset" instead of recreate and add a comment about recreation in `card_object_set_sprite()`.

MeirGavish (changes requested):
Cool, this seems like a nice little refactor that removes some code duplication and switches to a bit of a better sorting algorithm. 
I only gave this a superficial review and addressed the documentation.
@ricfehr3 If you can give this a more thorough review and see that the sorting still behaves correctly I'll appreciate it. If you want me to give this a more thorough review before merging let me know.

source/game.c
MathisMartin31: I played with things a bit with my card swapping PR and I believe this is the reason. I was tempted to do something similar with the Jokers since reordering them messes with the z-index, and there is no associated property in the Sprite object.

ricfehr3:
I will be taking a deeper look at this in the evening here. However from a quick overview I'd like to request some changes. Since this is a copilot assisted PR, I'm not comfortable with merging this without some tests. I do like passing in a function pointer to the sort function, however I think the sorting algorithm itself should be moved to it's own file and the function pointer should pass `void*` pointers that the implementations have to cast. That way, we can add tests for it in our testing suite.

@tygyh thank you for submitting this change, I agree it's a good addition. I'm not trying to be facetious or anything, but we've had quality issues in the past with these copilot PRs. So please take extra care to try and cover as many testing cases as possible.

include/sort.h
```diff
@@ -0,0 +1,15 @@
...
+#include <stdbool.h>
+
+/**
+ * Generic insertion sort implementation
+ * 
+ * @param array Array of pointers to sort
+ * @param size Number of elements in the array
+ * @param compare Comparison function that returns true if first arg should come before second arg
+ */
+void insertion_sort(void** array, int size, bool (*compare)(void*, void*));
```
ricfehr3: I think it would be helpful to add a struct that holds the arguments in this function. My reasoning is it would make it scalable in the future by passing the same struct to other sort algos and if an addition is needed, changing the struct type no function declarations would need to change. 

We may never add another sorting algorithm, but I think it's good design, regardless.

include/sort.h
```diff
@@ -0,0 +1,15 @@
+#ifndef SORT_H
+#define SORT_H
+
+#include <stdbool.h>
+
+/**
+ * Generic insertion sort implementation
```
ricfehr3: Thank you for documenting this function. Two things here.

1. Prefix the first line with `@brief` for consistency with our other documented functions: `* @brief Generic insertion sort implementation`.
2. Please add the file header so the function can be picked up by doxygen. See this for an example: https://github.com/cellos51/balatro-gba/blob/main/include/bitset.h

source/game.c
```diff
@@ -717,36 +718,40 @@ void set_seed(int seed)
     srand(rng_seed);
 }
 
-void sort_hand_by_suit()
+// Compare two cards for sorting by suit (primary) and rank (secondary)
+// Returns true if card_a should come before card_b
+static inline bool card_compare_by_suit(void* a, void* b)
```
ricfehr3: I don't think it makes sense to use `inline` here since it is passed as a function parameter.

source/game.c
```diff
@@ -717,36 +718,40 @@ void set_seed(int seed)
...
+    if (card_b == NULL) return true;
+    if (card_a->card->suit != card_b->card->suit)
+        return card_a->card->suit < card_b->card->suit;
+    return card_a->card->rank < card_b->card->rank;
 }
 
-void sort_hand_by_rank()
+// Compare two cards for sorting by rank only
+// Returns true if card_a should come before card_b
+static inline bool card_compare_by_rank(void* a, void* b)
```
ricfehr3: [Same as above](https://github.com/cellos51/balatro-gba/pull/268/files#r2579453553)

ricfehr3 (changes requested):
I left some comments here for some changes, nothing crazy there. I do have some additional notes:

There are two things that I would like to see, since a new file is being added and the sort functionality is important.

1. Tests for our test suite. 
> I should probably document this process better, but you need to create the test Makefile, modify the `run_tests.sh` file, and write the test. I think the [bitset test](https://github.com/cellos51/balatro-gba/tree/main/tests/bitset) is a good template.
2. We are trying to format our code to a style guide. To make sure this PR conforms to it, can you add `sort.h` and `sort.c` to the file: [`./github/workflows/run_tests_ci_workflow.yml` ](https://github.com/cellos51/balatro-gba/blob/07cab81ba10a6acc384995a298d40393c33fe0f4/.github/workflows/run_tests_ci_workflow.yml#L21). You can run `clang-format` locally (your IDE may already do this) to auto-edit the file to the `.clang-format` file.

Thanks!

---

**Edit**: Please also add the new test's build directory to the .gitignore . I just made this mistake myself, haha.

ricfehr3:
@tygyh Hey, no rush at all on this. Is this still active? I’d like to help move it forward.
If there’s no update in a few days, I can take over the branch and finish it while preserving your authorship.

tygyh:
@ricfehr3 I've been busy and don't remember what was left to do. I've rebased the branch and you are welcome to take over.

MeirGavish:
Maybe this should be merged after making the release mentioned in #307 unless you are sure this doesn't break sorting.

tygyh:
> Maybe this should be merged after making the release mentioned in #307 unless you are sure this doesn't break sorting.

In that case I think this can wait. It's a change which can introduce a bunch of bugs and the performance of the game isn't tied to the sorting afaik.

ricfehr3:
> > Maybe this should be merged after making the release mentioned in #307 unless you are sure this doesn't break sorting.
> 
> In that case I think this can wait. It's a change which can introduce a bunch of bugs and the performance of the game isn't tied to the sorting afaik.

Sweet, I can assign this to you @tygyh and I'll let you do your thing while we work on #307. Sorry if I rushed you.

MeirGavish:
I tested to see if this might solve #427 and unfortunately it seems not...
image

PR #269, opened 2025-11-30
Truncate Scores

MeirGavish:
In order to fix issue #200 this PR adds functions to properly truncate generic suffixed numbers into strings e.g. 54000 -> "54K" and properly center them within a rect.
This is applied to all scores but money is not truncated because the function needs to accept u32 to be compatible with the score and it would not work correctly for a signed value (which money has to be). I suppose it's safe to assume that if someone actually reached $1 million then he surpassed UINT32_MAX score long ago. I did however use my function to center it.

I also did a few more things that were required.

- I added tests for my new function since it was independent enough to put in util.c
@ricfehr3 Please take a look to see I added the tests correctly.
- I added a few more fixes changing score numbers to u32, one of them caused a bug in the score flames.
- Now that there's a proper function to center text in a rect, I removed `get_digits_odd()`, I also wanted to remove `get_digits_even()` but there's still one use left for it, I think we should change and add a functionality to properly center text under cards.
- I added one more tile to the score text rect so the score box can be fully used and properly centered. This caused an awkward issue where the number reached the "SCORE" text and the white of the text merged there in an unsightly way.
image

So I updated the backgrounds to give one more pixel to the score box on the left to avoid it.
image

This broke `reset_top_left_panel_bottom_row()` because of the hardcoded tiles so I took the opportunity to refactor it and use `main_bg_se_copy_rect()`.

Here are a few more screenshots of it working for high scores.

image

It will use up as much characters as possible in order to maintain as much accuracy as possible.
image

image

The score and mult boxes only have space for 3 characters, they will inevitably overflow. However, the text is correctly erased afterwards.
image

However overflow is avoided when possible for 1000s
image

B Suffix. I thought "Kilo", "Million", "Billion" would be more readable than "Kilo", "Mega", "Giga".
image

UINT32_MAX Properly clips. (Also see previous screenshot for total round score box)
image

### DIFF
--- a/source/blind.c
+++ b/source/blind.c
@@ -8,8 +8,9 @@
 
 #include <tonc.h>
 
-// +1 is added because we'll actually be indexing at 1, but if something causes you to go to ante 0, there will still be
-// a value there.
+// Maps the ante number to the base blind requirement for that ante.
+// The game starts at ante 1 which is at index 1 for base requirement 300.
+// Ante 0 is also there in case it is ever reached.
 static const u32 ante_lut[] = {100, 300, 800, 2000, 5000, 11000, 20000, 35000, 50000};
 
 // Palettes for the blinds (Transparency, Text Color, Shadow, Highlight, Main Color) Use this:


MeirGavish:
@ricfehr3 Looking at the logs I see my test failed (due to sprintf format issues that I guess I need to look into) but the workflow is marked as passed, do you know why is that? I would expect it to fail.
I want to look into it before fixing the test itself.
Edit: Copilot found the issue, it was because of the missing -Werror, thanks!
Edit Edit: I reverted it because it fails for the `snprintf()` in my added function, we need to figure out why there is this mismatch between compilers.

tests/util/Makefile
```diff
@@ -0,0 +1,16 @@
+
+CC := gcc
+CFLAGS := -I../../include -I. \
+          -g -O3 -Wall
```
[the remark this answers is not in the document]
MeirGavish: @ricfehr3 Okay, I see why it didn't fail now, it acts as a warning not an error.

source/game.c
```diff
@@ -1129,52 +1123,48 @@ void change_background(enum BackgroundId id)
...
+    tte_printf("#{P:%d,%d; cx:0x%X000}%s", money_text_rect.left, money_text_rect.top, TTE_YELLOW_PB, money_str_buff);
 }
 
 // Show/Hide flaming score effect if we will score
 // more than the required amount or not
 void check_flaming_score()
 {
-    int curr_score = chips * mult;
+    u32 curr_score = u32_protected_mult(chips, mult);
     int required_score = blind_get_requirement(current_blind, ante);
```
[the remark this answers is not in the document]
MeirGavish: Right, missed it.

MathisMartin31:
Oooooh that's awesome work right there, love the look of the new scores!

source/blind.c
```diff
@@ -10,7 +10,7 @@
 
 // +1 is added because we'll actually be indexing at 1, but if something causes you to go to ante 0, there will still be
 // a value there.
```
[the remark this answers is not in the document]
MeirGavish: I think it needs to be updated to reflect why the ante starts at 100 and not 300.

include/util.h
```diff
@@ -45,8 +37,45 @@ static inline int get_digits_even(int n)
...
 
+#define ONE_K 1000
+#define ONE_M 1000000
+#define ONE_B 1000000000
+
+#define ONE_K_ZEROS 3
+#define ONE_M_ZEROS 6
+#define ONE_B_ZEROS 9
+
+// The suffix replaces everything past the third digit, e.g. "999K" -> "1M"
```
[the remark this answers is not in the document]
MeirGavish: Not sure I like this suggestion but maybe I need to reword things there to make it a little clearer...

tests/util/Makefile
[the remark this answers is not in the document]
MeirGavish: This makes the test fail due to `snprintf` compiler mismatch so I reverted it.

ricfehr3: I tested this, and it looks like the best option is to suppress this error in the test Makefile by adding:

- `-Wno-format`

I see what your saying about the compiler differences. That int length issue always sneaks up in the worst ways, haha.

tests/util/util_test.c
```diff
@@ -0,0 +1,133 @@
...
+
+    char max_uint_str_buff[UINT_MAX_DIGITS + 1] = {'\0'};
+    snprintf(max_uint_str_buff, sizeof(max_uint_str_buff), "%lu", UINT32_MAX);
+
+    suffixed_str_buff[0] = '\0';
+    truncate_uint_to_suffixed_str(UINT32_MAX, UINT_MAX_DIGITS, suffixed_str_buff);
+    assert(strcmp(suffixed_str_buff, max_uint_str_buff) == 0);
+}
+
+int main()
```
ricfehr3, line 129: These tests look great! Very well done. Seems to be very straightforward.

include/graphic_utils.h
```diff
@@ -142,19 +143,40 @@ void main_bg_se_move_rect_1_tile_vert(Rect se_rect, int direction);
...
 void tte_erase_rect_wrapper(Rect rect);
 
-/* Changes rect->left so it fits the digits of num exactly when right aligned to rect->right.
- * Assumes num is not negative.
+/* Changes rect->left so it fits a string exactly when right aligned to rect->right.
  * 
- * overflow_direction determines the direction the number will overflow
+ * overflow_direction determines the direction the string will overflow
  * if it's too large to fit inside the rect. 
  * Should be either OVERFLOW_LEFT or OVERFLOW_RIGHT.
```
ricfehr3: I think we could probably change `SCREEN_RIGHT`, `SE_UP`, `OVERFLOW_LEFT`, etc. to an enum. I think it would be cleaner.

include/graphic_utils.h
```diff
@@ -142,19 +143,40 @@ void main_bg_se_move_rect_1_tile_vert(Rect se_rect, int direction);
...
  * so it's a whole number of tiles to fit TTE characters
  * 
  * Note that both rect->left and rect-right need to be defined, top and bottom don't matter
  */
-void update_text_rect_to_right_align_num(Rect* rect, int num, int overflow_direction);
+void update_text_rect_to_right_align_str(Rect* rect, char* str, int overflow_direction);
+
+
+/** 
+ * @brief Updates a rect so a string is centered within it.
```
ricfehr3, line 160: Thank you for adding the documentation here.

source/graphic_utils.c
```diff
@@ -235,22 +236,41 @@ void tte_erase_rect_wrapper(Rect rect)
     tte_erase_rect(rect.left, rect.top, rect.right, rect.bottom);
 }
 
-void update_text_rect_to_right_align_num(Rect* rect, int num, int overflow_direction)
+void update_text_rect_to_right_align_str(Rect* rect, char* str, int overflow_direction)
```
ricfehr3: We can change this to a `const char* str`

ricfehr3 (commented):
Awesome work, as per usual @MeirGavish . Thank you for adding the documentation comments as well. I'm excited to see all that come together, lol. 

I've got a few suggested changes, but otherwise, great work.

ricfehr3 (changes requested):
Requesting changes from last comment.

tests/util/util_test.c
MeirGavish: Thanks!

include/graphic_utils.h
MeirGavish: Yeah, I added a to do for that there, I'll do it.

source/graphic_utils.c
MeirGavish: Oh, good point!

tests/util/Makefile
MeirGavish:
```suggestion
          -g -O3 -Wall -Werror -Wno-format
```

source/graphic_utils.c
MeirGavish:
```suggestion
void update_text_rect_to_right_align_str(Rect* rect, const char* str, int overflow_direction)
```

include/graphic_utils.h
```diff
@@ -142,19 +143,40 @@ void main_bg_se_move_rect_1_tile_vert(Rect se_rect, int direction);
...
  * Should be either OVERFLOW_LEFT or OVERFLOW_RIGHT.
  * 
- * The rect is in number of pixels but should be a multiple of TILE_SIZE
+ * The rect is in number of pixels but should be a multiple of TTE_CHAR_SIZE
  * so it's a whole number of tiles to fit TTE characters
  * 
  * Note that both rect->left and rect-right need to be defined, top and bottom don't matter
  */
-void update_text_rect_to_right_align_num(Rect* rect, int num, int overflow_direction);
+void update_text_rect_to_right_align_str(Rect* rect, char* str, int overflow_direction);
```
MeirGavish:
```suggestion
void update_text_rect_to_right_align_str(Rect* rect, const char* str, int overflow_direction);
```

source/graphic_utils.c
```diff
@@ -235,22 +236,41 @@ void tte_erase_rect_wrapper(Rect rect)
...
+        rect->left = max(0, rect->right - str_len * TTE_CHAR_SIZE);
     }
     else if (overflow_direction == OVERFLOW_RIGHT)
     {
-        int num_fitting_digits = rect_width(rect) / TILE_SIZE;
-        if (num_digits < num_fitting_digits)
-            rect->left += (num_fitting_digits - num_digits) * TILE_SIZE;
+        int num_fitting_chars = rect_width(rect) / TTE_CHAR_SIZE;
+        if (str_len < num_fitting_chars)
+            rect->left += (num_fitting_chars - str_len) * TTE_CHAR_SIZE;
```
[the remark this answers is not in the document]
MeirGavish: I'll do 2 things here.
1. Make sure `rect_width` and `rect_height` can't return a negative number by returning 0 if rect->right < rect->left
2. Use int for `str_len` and make the cast happen implicitly at assignment.

There's no reason to really fuss over signed/unsigned mismatch here.

include/util.h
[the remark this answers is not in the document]
MeirGavish: Meh, my comment seems clear enough to me...

MeirGavish:
Seems good to me now.

include/graphic_utils.h
MeirGavish: @ricfehr3 I ended up removing `SE_UP` and `SE_DOWN` and just used `SCREEN_UP` and `SCREEN_DOWN` instead.

ricfehr3 (approved):
@MeirGavish This looks good. Let's merge this one before the clang-format pr #246 . It's easier to reformat the files rather than make you merge the format changes in.

PR #276, opened 2025-12-04
Restored removed joker shop availability check function

MeirGavish:
It seems PR #259 removed `_get_shop_joker_avail()` and we haven't noticed since its only use is conditionally compiled when testing jokers. I restored it and renamed it `is_shop_joker_avail()` so the joker testing code can be used again.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -552,6 +552,11 @@ static int four_fingers_joker_count = 0;
 static int straight_and_flush_size = STRAIGHT_AND_FLUSH_SIZE_DEFAULT;
 
 GBLA_UNUSED
+static inline bool is_shop_joker_avail(int joker_id)
+{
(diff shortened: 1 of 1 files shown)


ricfehr3:
Opa, that was my bad. @MeirGavish for smaller PRs like this, would you like me to merge after I test/review? Or would you prefer to do it when you would like and I can just review?

MeirGavish:
@ricfehr3 Yeah, if it's a small PR I made and everything is fine, you can totally merge it.

MeirGavish:
@ricfehr3 If it's made by someone else I will usually want to go over it, I usually self-request a review if I want to review it before it goes in and if it's small it usually doesn't take me long.

ricfehr3:
Right on, I'll merge. Thanks!

PR #279, opened 2025-12-04
Fixed bug where reroll button is triggered by any key

MeirGavish:
This fix needs to be merged with or before #258 because it makes this bug worse.


MeirGavish:
Merging this, it's a small simple change.

PR #280, opened 2025-12-04
Swap Cards in Hand only

MathisMartin31:
Branching from my other PR #258 

This one is only about swapping Cards in hand while playing a round

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -1975,9 +1975,8 @@ static inline void process_user_card_movement(enum ScreenHorzDir move_dir)
     // swap cards around if A is held down when pressing D-pad keys
     if (key_is_down(SELECT_CARD) && can_move_card && !card_moved_too_fast)
     {
-        bool selection_not_at_border = (move_dir == SCREEN_LEFT)
-            ? selection_x < hand_top
-            : selection_x > 0;
+        bool selection_not_at_border =
+            (move_dir == SCREEN_LEFT) ? selection_x < hand_top : selection_x > 0;
 
         if (selection_not_at_border)
         {
--- a/source/game.c
+++ b/source/game.c
@@ -1961,12 +1961,13 @@ static inline void select_current_card(void)
 
 // card moving logic
 
+// true if and only if we are currently moving a card around
+static bool moving_card = false;
+
 // This will prevent us from moving cards around if we selected one
 // by moving too fast after pressing the A button
 static bool card_moved_too_fast = false;
-
-// true if and only if we are currently moving a card around
-static bool moving_card = false;
+static bool card_selected_instead_of_moved = false;
 
 // After pressing A, if we press Left/Right too fast, we should select the card
 // and change focus to the next one, instead of swapping them
@@ -1984,7 +1985,7 @@ static inline void game_playing_apply_card_movement_input(enum ScreenHorzDir mov
     card_moved_too_fast = (timer - selection_hit_timer) < card_swap_time_threshold;
 
     // swap cards around if A is held down when pressing D-pad keys
-    if (key_is_down(SELECT_CARD) && !card_moved_too_fast)
+    if (key_is_down(SELECT_CARD) && !card_moved_too_fast && !card_selected_instead_of_moved)
     {
         bool selection_not_at_border =
             (move_dir == SCREEN_LEFT) ? selection_x < hand_top : selection_x > 0;
@@ -2000,13 +2001,10 @@ static inline void game_playing_apply_card_movement_input(enum ScreenHorzDir mov
     else
     {
         // select current card if we tried moving it too fast
-        if (card_moved_too_fast && !moving_card)
+        if (key_released(SELECT_CARD) || (card_moved_too_fast && !moving_card))
         {
             select_current_card();
-            // raise this flag to prevent selecting the next card when releasing A
-            moving_card = true;
-            // and also this one so we don't drag the next card along if we don't release A
-            card_moved_too_fast = false;
+            card_selected_instead_of_moved = true;
         }
         hand_set_focus(next_card);
     }
@@ -2091,6 +2089,10 @@ static inline void game_playing_process_hand_select_input(void)
 
         if (selection_x > hand_top / 2)
         {
+            discard_button_highlighted = false; // Play button
+        }
+        else
+        {
             discard_button_highlighted = true; // Discard button
         }
     }
@@ -2127,12 +2129,13 @@ static inline void game_playing_process_hand_select_input(void)
         // select card if we were not moving it around
         else if (key_released(SELECT_CARD))
         {
-            if (!moving_card)
+            if (!moving_card && !card_selected_instead_of_moved)
             {
                 select_current_card();
             }
             moving_card = false;
             card_moved_too_fast = false;
+            card_selected_instead_of_moved = false;
             selection_hit_timer = TM_ZERO;
         }
 
--- a/source/game.c
+++ b/source/game.c
@@ -1110,7 +1110,11 @@ static void reorder_card_sprites_layers(void)
                 break;
             }
 
-            // if there is one, shift it and all the cards that follow
+            // If there is one, shift it and all the cards that follow
+
+            // Iterating up to `hand_top - non_null_card_idx + 1` should end up out of bounds
+            // but for some reason it doesn't pose any issue, and taking out the +1 breaks
+            // the code, so I'll be elaving it here until someone figures it out ^^'
             for (int j = 0; j <= hand_top - non_null_card_idx + 1; j++)
             {
                 hand[i + j] = hand[non_null_card_idx + j];
@@ -1949,14 +1953,14 @@ static inline void game_playing_process_hand_select_input(void)
     // This should fix inputs sometimes not registering when quickly selecting cards
     static const int card_swap_time_threshold = 5;
     static bool card_moved_too_fast = false;
-    static uint move_timer = TM_ZERO;
+    static uint selection_hit_timer = TM_ZERO;
 
     // Register timer when we hit A to pick a card
     // If we try to move the picked card before card_swap_time_threshold frames,
     // We will select it instead of moving it
     if (key_hit(SELECT_CARD))
     {
-        move_timer = timer;
+        selection_hit_timer = timer;
     }
 
     if (key_hit(KEY_LEFT))
@@ -1964,7 +1968,7 @@ static inline void game_playing_process_hand_select_input(void)
         if (selection_y == 0)
         {
             // Do not use FRAMES(x) here as we are counting real frames ignoring game speed
-            card_moved_too_fast = (timer - move_timer) < card_swap_time_threshold;
+            card_moved_too_fast = (timer - selection_hit_timer) < card_swap_time_threshold;
 
             // The reason why this adds 1 is because the hand is drawn from right to left.
             // There is no particular reason for this, it's just how I did it.
@@ -2004,7 +2008,7 @@ static inline void game_playing_process_hand_select_input(void)
     {
         if (selection_y == 0)
         {
-            card_moved_too_fast = (timer - move_timer) < card_swap_time_threshold;
+            card_moved_too_fast = (timer - selection_hit_timer) < card_swap_time_threshold;
 
             int right_card = selection_x - 1;
 
@@ -2109,7 +2113,7 @@ static inline void game_playing_process_hand_select_input(void)
             }
             moving_card = false;
             can_move_card = true;
-            move_timer = TM_ZERO;
+            selection_hit_timer = TM_ZERO;
         }
 
         if (key_hit(DESELECT_CARDS))
--- a/source/game.c
+++ b/source/game.c
@@ -1044,6 +1044,8 @@ void display_mult(void)
     check_flaming_score();
 }
 
+// idx_a and idx_b are assumed to be valid indexes within the hand array
+// no checks will be performed here for performance's sake
 static void swap_cards_in_hand(int idx_a, int idx_b)
 {
     CardObject* temp = hand[idx_a];
@@ -1051,7 +1053,7 @@ static void swap_cards_in_hand(int idx_a, int idx_b)
     hand[idx_b] = temp;
 }
 
-static void sort_hand_by_suit()
+static void sort_hand_by_suit(void)
 {
     for (int idx_a = 0; idx_a < hand_top; idx_a++)
     {
@@ -1068,7 +1070,7 @@ static void sort_hand_by_suit()
     }
 }
 
-static void sort_hand_by_rank()
+static void sort_hand_by_rank(void)
 {
     for (int idx_a = 0; idx_a < hand_top; idx_a++)
     {
@@ -1083,7 +1085,7 @@ static void sort_hand_by_rank()
     }
 }
 
-static void rearrange_card_sprites()
+static void rearrange_card_sprites(void)
 {
     // Update the sprites in the hand by destroying them and creating new ones in the correct order
     // (This feels like a diabolical solution but like literally how else would you do this)
@@ -1135,7 +1137,7 @@ static void rearrange_card_sprites()
     }
 }
 
-void sort_cards()
+static void sort_cards(void)
 {
     if (sort_by_suit)
     {
--- a/source/game.c
+++ b/source/game.c
@@ -2111,7 +2111,7 @@ static inline void game_playing_process_hand_select_input(void)
             }
         }
         else
-        {   
+        {
             game_playing_highlight_discard_btn();
             if (key_hit(SELECT_CARD) && discards > 0 && hand_can_discard())
             {
--- a/source/game.c
+++ b/source/game.c
@@ -1953,27 +1953,31 @@ static inline void select_current_card(void)
 }
 
 // card moving logic
-static bool can_move_card = true;
-static bool moving_card = false; // true if we are currently moving a card around
+
+// This will prevent us from moving cards around if we selected one
+// by moving too fast after pressing the A button
+static bool card_moved_too_fast = false;
+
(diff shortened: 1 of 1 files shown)


MeirGavish:
There's already a review in #258 from Copilot but I want to ask another one here too.

source/game.c
```diff
@@ -1030,61 +1030,79 @@ void display_mult(void)
...
+            }
+
+            // exit loop if there are no non-NULL cards left/there are no more sprites to destroy
+            if (non_null_card_idx > hand_top)
+            {
+                break;
+            }
+
+            // if there is one, shift it and all the cards that follow
+            for (int j = 0; j <= hand_top - non_null_card_idx + 1; j++)
```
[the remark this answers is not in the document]
MathisMartin31: As I said in the previous review, I haven't really understood yet why it works with the +1 and why it doesn't without. It's kind of weird and I have to dig deeper.

source/game.c
```diff
@@ -1030,61 +1030,79 @@ void display_mult(void)
...
 }
 
-static inline void sort_hand_by_suit(void)
+static void swap_cards_in_hand(int idx_a, int idx_b)
 {
-    for (int a = 0; a < hand_top; a++)
+    CardObject* temp = hand[idx_a];
+    hand[idx_a] = hand[idx_b];
+    hand[idx_b] = temp;
+}
```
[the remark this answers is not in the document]
MathisMartin31: Added a comment about the need to provide valid indexes to the function. Its purpose, however, seem plenty obvious to me :thinking:

source/game.c
```diff
@@ -1035,61 +1035,81 @@ void display_mult(void)
...
+// no checks will be performed here for performance's sake
+static void swap_cards_in_hand(int idx_a, int idx_b)
 {
-    for (int a = 0; a < hand_top; a++)
+    CardObject* temp = hand[idx_a];
+    hand[idx_a] = hand[idx_b];
+    hand[idx_b] = temp;
+}
+
+static void sort_hand_by_suit(void)
```
ricfehr3: Should remain `inline`, it's only used in `sort_cards`

source/game.c
```diff
@@ -1035,61 +1035,81 @@ void display_mult(void)
...
-                hand[a] = hand[b];
-                hand[b] = temp;
+                swap_cards_in_hand(idx_a, idx_b);
             }
         }
     }
 }
 
-static inline void sort_hand_by_rank(void)
+static void sort_hand_by_rank(void)
```
ricfehr3: Should remain `inline`, it's only used in `sort_cards`

source/game.c
```diff
@@ -1882,18 +1916,69 @@ static inline bool hand_play(void)
     return true;
 }
 
+// When holding A, if we press an arrow key too fast, we should select the card
+// and change focus to the next one, instead of swapping them
+// This should fix inputs sometimes not registering when quickly selecting cards
+#define CARD_SWAP_TIME_THRESHOLD 5
```
ricfehr3: Since this is only used in `game_playing_process_hand_select_input`, it should be a `static const int` in the scope of the function.

source/game.c
```diff
@@ -1882,18 +1916,69 @@ static inline bool hand_play(void)
...
+
+static inline void select_current_card(void)
+{
+    hand_toggle_card_selection();
+    set_hand();
+}
+
 static inline void game_playing_process_hand_select_input(void)
 {
     static bool discard_button_highlighted =
```
ricfehr3: not on you, but could you move the comment above the declaration, so it looks better when formatted.

source/game.c
```diff
@@ -1882,18 +1916,69 @@ static inline bool hand_play(void)
...
+                {
+                    select_current_card();
+                    // raise this flag to prevent selecting the next card when releasing A
+                    moving_card = true;
+                    // and also this one so we don't drag the next card along if we don't release A
+                    can_move_card = false;
+                }
+                // The reason why this adds 1 is because the hand is drawn from right to left.
+                // There is no particular reason for this, it's just how I did it.
+                hand_set_focus(selection_x + 1);
```
ricfehr3: I think it's fine to do this way, but I think it would make more sense if above you declared `left_card_idx = selection_x;` and `right_card_idx = selection_x + 1`.

source/game.c
```diff
@@ -1882,18 +1916,69 @@ static inline bool hand_play(void)
...
         if (selection_y == 0)
         {
-            // The reason why this adds 1 is because the hand is drawn from right to left. There is
-            // no particular reason for this, it's just how I did it.
-            hand_set_focus(selection_x + 1);
+            // Do not use FRAMES(x) here as we are counting real frames ignoring game speed
+            card_moved_too_fast = (timer - move_timer) < CARD_SWAP_TIME_THRESHOLD;
+
+            // swap cards around if A is held down when pressing D-pad keys
+            if (key_is_down(SELECT_CARD) && can_move_card && !card_moved_too_fast)
```
ricfehr3: This logic is repeated below for moving the card to the right. This logic should be broken out into it's own function, and have as much of the code repeat as possible. I figure the function would be something like: 

```c
enum Direction 
{
    DIR_LEFT;
    DIR_RIGHT;
};
process_user_card_movements(enum Direction dir);
```

Feel free to implement whatever you see fit though. 

I do see that below there is logic to check `key_released(SELECT_CARD)` that should change the `moving_card`, `can_move_card`, and `move_timer` variables. For this, I _think_ the logic would still be sound if instead of first checking if KEY_LEFT or KEY_RIGHT was hit, you instead first check if `selection_y == 0` then check for key_hit in there. Still, preserving the `discard_button_highlighted` above.

ricfehr3 (dismissed):
Sorry, it took me a minute to get to this. Once again, great addition and huge thank you! I have some suggestions below:

source/game.c
```diff
@@ -1035,61 +1035,81 @@ void display_mult(void)
...
+            if (non_null_card_idx > hand_top)
+            {
+                break;
+            }
+
+            // if there is one, shift it and all the cards that follow
+            for (int j = 0; j <= hand_top - non_null_card_idx + 1; j++)
+            {
+                hand[i + j] = hand[non_null_card_idx + j];
+            }
```
MeirGavish: I think this operation needs to be extracted to a function `shift_null_cards_to_end`, there's too much logic nested in this function.

source/game.c
```diff
@@ -1035,61 +1035,81 @@ void display_mult(void)
...
-                hand[a] = hand[b];
-                hand[b] = temp;
+                swap_cards_in_hand(idx_a, idx_b);
             }
         }
     }
 }
 
-static void sort_cards(void)
+static void rearrange_card_sprites(void)
```
MeirGavish: Also maybe rename the function `reorder_card_sprite_layers` to make it a little clearer to what it actually does.

source/game.c
MeirGavish: I think you can use the existing `enum ScreenHorzDir` for this, it's pretty generic. Also it's important to extract this logic from here, this function is very long and nested as it is.

source/game.c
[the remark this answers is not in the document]
MeirGavish: This is a little scary, at least add a comment for this.

source/game.c
MeirGavish: Yes, I think this will be a good idea, you can then move this comment to the `right_card_idx = selection_x + 1` assignment

source/game.c
```diff
@@ -1882,18 +1916,69 @@ static inline bool hand_play(void)
...
+    static bool moving_card = false;         // true if we are currently moving a card around
+    static bool card_moved_too_fast = false; // see define of CARD_SWAP_TIME_THRESHOLD
+    static uint move_timer = TM_ZERO;        // idem
+
+    // Register timer when we hit A to pick a card
+    // If we try to move the picked card before CARD_SWAP_TIME_THRESHOLD frames,
+    // We will select it instead of moving it
+    if (key_hit(SELECT_CARD))
+    {
+        move_timer = timer;
```
MeirGavish: Rename this variable to `selection_hit_timer`, saying "move" is incorrect since this will not necessarily be a move, in case it's too fast it will turn into a select.

source/game.c
```diff
@@ -1882,18 +1916,69 @@ static inline bool hand_play(void)
...
+                    select_current_card();
+                    // raise this flag to prevent selecting the next card when releasing A
+                    moving_card = true;
+                    // and also this one so we don't drag the next card along if we don't release A
+                    can_move_card = false;
+                }
+                // The reason why this adds 1 is because the hand is drawn from right to left.
+                // There is no particular reason for this, it's just how I did it.
+                hand_set_focus(selection_x + 1);
+            }
```
MeirGavish: Something really bothers me about this else clause and the corresponding if condition, it's very difficult to understand. I think there is a misuse of the boolean variables here, you set `moving_card` to true even though you are not really moving a card and `can_move_card` is only set to false in this occasion. I think the logic here is not completely sound. If you need to prevent selection on key  release maybe do it more explicitly with a different boolean `card_selected_instead_of_moved` or something of the sort. Also, since the if clause is checking so many variables, the else is entered on many occasions, I wonder if you need to change it to `else if (moved_too_fast)`? Because that's the precise case you are dealing with here.

MeirGavish (changes requested):
Good job! Very good feature to have! With this the gameplay experience will be pretty complete because the player will have full control of the order of activation.
I left a few comments, and also address @ricfehr3's review.

MathisMartin31:
Thank you both for the extensive reviews! I'm a little busy this evening but I'll try to address them later tonight or tomorrow :)

source/game.c
MathisMartin31: The difficulty would be the static variables defined earlier, and I don't think I can just pass a pointer to them as argument, can I?. I could move the definition outside the function

MeirGavish: You can move them out in my opinion.

ricfehr3:
> Thank you both for the extensive reviews! I'm a little busy this evening but I'll try to address them later tonight or tomorrow :)

No thank you for always working on stuff the people wanna see, haha. Happy to review it, and no rush at all.

source/game.c
ricfehr3: I was suggesting to encapsulate all the logic that uses the static variables into one function. Below it looks like all logic that has to do with the user moving the card rely on `selection_y == 0`. So instead of checking `key_hit(KEY_LEFT); if(selection_y == 0)` and `key_hit(KEY_RIGHT); if(selection_y == 0)` then the below `else if(selection_y == 0)` check could all be reversed so you first do a check for `selection_y == 0`, then perform the key_hit logic in a function.

Something like:

```c
void process_user_card_movements()
{
    static bool can_move_card = true;
    static bool moving_card = false; 

    if(key_hit(KEY_LEFT))
    {
        ...
    }
    else if(key_hit(KEY_RIGHT))
    {
        ...
    }
    else
    {
        ...
    }
}

...

static inline void game_playing_process_hand_select_input(void)
{
...
    if(selection_y == 0)
    {  
        process_user_card_movements();
    }
    ...
```

MeirGavish: Yeah, seems like it would be good to invert the logic like that.

MathisMartin31: I'll have to move more logic around because the code actually looks like this:
```
if (key_hit(KEY_LEFT))
{
    if (selection_y == 0)
    {
        // move/select card
    }
    else
    {
        // highlight discard/play buttons
    }
}
else if(key_hit(KEY_RIGHT))
{
    ...
}
else if (key_hit(KEY_UP))
{
    ...
}
else if (key_hit(KEY_DOWN))
{
    ...
}
```

The code is organized around taking into account only 1 D-pad input press, which could save us some bugs on emulators where you could end up pressing 2 on a keyboard or something.

Maybe the new `process_user_card_movement` can take the horizontal direction as argument?

Other issue is that the static booleans and counter are used later in the function, so they will have to be completely extracted outside either functions so I can use them in both

MathisMartin31: I moved it in its own function. How does it look now?

MeirGavish: I still think it would be better to invert the logic and have the row check in the outer if statement, the inner directional button checking in another function can still be `if (key_hit(KEY_LEFT)) else if (key_hit(KEY_RIGHT)) etc.` and respond to only a single key press. 
That said, I'm also okay with how it is now and I think this refactor can be done later if you don't want to do it now, I can do it too, but I'll wait to see if @ricfehr3 agrees.

ricfehr3: I'm on the fence on this one. Looking at the problem @MathisMartin31 was talking about, I'm not sure if I want to break that input `if`/`else` statement which leans me toward the "refactor later" idea. But ultimately, I would like to see the logic flow: `if(selection_y == 0) { update_input(); }` and in the `update_input()` function, doing that `if`/`else` as @MeirGavish described. This would allow the new static variables to be encapsulated in that function as well, which is more towards our goal of decoupling `game.c` and would simplify that `if`/`else` block in the input. @MathisMartin31 Would you be down to make the change in the way I described and letting me test with it? If it doesn't break anything I would like to see it that way.   

Expanding on my previous example, I think `selection_y == 0` should be the first check, and we can think of it as "We are on the cards row", and we can do all the logic mutually exclusive of the counter "We are on the play/discard button row." I think that will preserve the logic and read better, you could even move the logic for the play/discard buttons into it's own function. I a full example below, what do you think? :thinking: 

```c
void process_user_card_movements()
{
    static bool can_move_card = true;
    static bool moving_card = false; 

    if(key_hit(KEY_LEFT))
    {
        ...
    }
    else if(key_hit(KEY_RIGHT))
    {
        ...
    }
    else
    {
        ...
    }
}

...

void process_user_button_movements()
{
    if (key_hit(KEY_LEFT))
        // highlight discard/play buttons
    }
    else if(key_hit(KEY_RIGHT))
    {
        ...
    }
    else if (key_hit(KEY_UP))
    {
        ...
    }
    else if (key_hit(KEY_DOWN))
    {
    ...
    }
}

...
 
static inline void game_playing_process_hand_select_input(void)
{
    ...
    if(selection_y == 0) // on card row
    {  
        process_user_card_movements();
    }
    else // on button row
    {
        process_user_button_movements();
    }
    ...
}
```

MathisMartin31: Yes I see what you two mean. I'll take a crack at it as soon as I can ^^

source/game.c
MathisMartin31: I'll revisit these booleans once I'll have refactore all the card selection/movement logic into a single function. Maybe having everything closer to each other will help me take a better look

MeirGavish: @MathisMartin31 
Alright. Let me know if you need help, as I said before, if you want I can do the refactor.

MathisMartin31: You know what, I'll take your offer ^^

You can go ahead with the refactor if you have some time

MeirGavish: Alright, I'll give this a shot, I'll just branch out from your branch and create another PR.
I'll probably not have time to do this today, but probably tomorrow.

MeirGavish: @ricfehr3 @MathisMartin31 
This refactor is a bit big because it needs to redo the handling for vertical movement too, the code here is a bit spaghetti in that regard - subtly mixing handling for vertical and horizontal movement in many cases. I'm starting to think it might just be better to refactor this part to use the selection grid instead.
I think that refactor should go into another PR, I opened issue #320 for it.
In my opinion if the issue in this comment is resolved, I can approve this PR and get to refactoring once it's merged. For now I opened [a PR for this PR](https://github.com/MathisMartin31/gba-latro_MathisMartin/pull/1#issue-3734772436) just for a smaller refactor.

MathisMartin31: I was able to remove the `can_move` flag as, as you said, it was redundant with `card_moved_too_fast`. I tested a bit and found no difference in how the inputs behaved, which is good because I haven't broken anything, but also bad because I sometimes still fail to select a card when pressing buttons in really quick succession. 

As for the `else`, it should stay as an `else` and not become an `else if`, because we need to select the next card regardless of whether we need to select the current one or not. It goes something like this:

1. Do we move the card? If yes, hover the next one if we haven't hit the hand's edge
2. If not, do we select it instead? Then always hover the next one

MeirGavish: @MathisMartin31 
I still don't think this else clause should change `moving_card` and `card_moved_too_fast` just to control other statements because then their values are wrong - you are not moving a card and the card did move too fast. I think you should add a bool `card_selected_instead_of_moved` that will actually mean what happened and add checks for it in the corresponding ifs.
Other than that, it bothers me too that quick selections are not always registered, I tried looking at it and I can't figure out why this is happening either.
Also, can you merge [this refactor PR](https://github.com/MathisMartin31/gba-latro_MathisMartin/pull/1#issue-3734772436) I made?

MathisMartin31: @MeirGavish 

> I still don't think this else clause should change moving_card and card_moved_too_fast just to control other statements because then their values are wrong - you are not moving a card and the card did move too fast. I think you should add a bool card_selected_instead_of_moved that will actually mean what happened and add checks for it in the corresponding ifs.

Ok I see what you mean, I'll do that ^^

> Also, can you merge this refactor PR I made?

Done ^^

> Other than that, it bothers me too that quick selections are not always registered, I tried looking at it and I can't figure out why this is happening either.

After some testing I'm almost sure that it happens when we release A and press Left/Right on the same frame, which may lead to either not doing anything, or selecting the card twice (which, in turn, amounts to not doing anything) :thinking:.

Maybe the timer trick to select if the card was moved too fast after pressing A can be extended to when we move too fast after releasing A

MathisMartin31: Okay so I added `key_released(SELECT_CARD)` to the condition for when we select a card instead of moving it and it feels much smoother. I don't know if I totally fixed it, but I think it's way better now.

MeirGavish: Interesting, good job! It does feel better but still not perfect. I guess it's good enough and I can approve this.

MeirGavish (approved):

Issue #281, opened 2025-12-05
Separate clang format test to its own workflow


MeirGavish:
Separate the clang-format to its own workflow and not skip the build if other workflows fail. If one workflow fails you should be able to see if others succeed or fail. The build is an especially important test that should always run.

_Originally posted by @MeirGavish in https://github.com/cellos51/balatro-gba/issues/258#issuecomment-3615712216_

PR #283, opened 2025-12-05
Separate clang-format test into its own workflow

ricfehr3:
For #281 

[Here](https://github.com/cellos51/balatro-gba/actions/runs/19975368645/job/57290116222) is an example of what a failed run shows.

### DIFF
--- a/.github/workflows/nightly_release_ci_workflow.yml
+++ b/.github/workflows/nightly_release_ci_workflow.yml
@@ -21,7 +21,23 @@ jobs:
     strategy:
       fail-fast: false
     runs-on: ubuntu-latest
-    needs: build-job
     steps:
+      - name: Download the build output
+        uses: actions/download-artifact@v5
+        with:
+          name: build-out
+          path: build
+
       - name: Package Release
         run: mv build/balatro-gba.gba build/balatro-gba_nightly_$(date +'%+4Y%m%d')_$(cat build/githash.txt).gba && zip -j release build/*.gba
+
+      - name: Deploy release
+        uses: WebFreak001/deploy-nightly@v3.2.0
+        with:
+          upload_url: https://uploads.github.com/repos/cellos51/balatro-gba/releases/258698380/assets{?name,label}
+          release_id: 258698380
+          asset_path: ./release.zip # path to archive to upload
+          asset_name: nightly-gbalatro-$$.zip # name to upload the release as, use $$ to insert date (YYYYMMDD) and 6 letter commit hash
+          asset_content_type: application/zip
+          max_releases: 7 
+          ignore_hash: false # Only upload unique versions


.github/workflows/clang_format_ci_workflow.yml
MeirGavish, line 1: How did you make git/GitHub realize this is extracted from the original file and not think it's deleted and created in a new one?

MeirGavish (commented):
Thanks! I see that in the example failed test the other workflows are still skipped when clang fails and I prefer it to be the other way around, can we do that?

.github/workflows/clang_format_ci_workflow.yml
ricfehr3: I have no idea, sometimes git is weird. I guess it got confused 🤷 , the files are all here though.

ricfehr3:
> Thanks! I see that in the example failed test the other workflows are still skipped when clang fails and I prefer it to be the other way around, can we do that?

I went ahead and made that change,~~but I also realize we can run all 3 simultaneously. What do you think of that? Get all the errors at once.~~

---

**Edit:**
I think we should actually leave it like this. I want unit tests to pass in the nightly before building the release. We should never have code that fails the format test in main, but if it fails that test, it isn't crucial to not publish the nightly. It's an unlikely situation too. I think we can merge this if you're happy.

.github/workflows/build_ci_workflow.yml
```diff
@@ -21,14 +21,14 @@ on:
...
-    uses: ./.github/workflows/run_tests_ci_workflow.yml
+  unit-tests-job:
+    name: Run unit tests
+    uses: ./.github/workflows/unit_tests_ci_workflow.yml
 
   build:
     name: Build ROM
     runs-on: ubuntu-latest
-    needs: run-tests-job
+    needs: unit-tests-job
```
MeirGavish: Actually I don't think the build should depend on the unit tests, they are separate tests that test different things and can fail independently. The unit tests check logic for the base tools while the build just tests build for the main target. They build separate targets so there's no dependency there.

.github/workflows/build_ci_workflow.yml
```diff
@@ -73,3 +73,9 @@ jobs:
           name: build-out
           path: build
           retention-days: 5
+
+  clang-format-job:
+    name: Run clang-format test
+    needs: [ unit-tests-job, build ]
```
MeirGavish: I think the dependency here is not critical but I think it's okay to leave it.

MeirGavish (changes requested):
Actually I think we should change the dependencies, see my comment for this.

I think since I see you can add a dependency on multiple workflows, the right thing to do is make the nightly depend on both the unit tests and build workflows.

ricfehr3:
@MeirGavish I tested the workflow a few time to make sure it behaves as expected, but in addition to the action from this PR, when you deploy the nightly it has this:

image

I think this is what we want, what do you think?

.github/workflows/nightly_release_ci_workflow.yml
```diff
@@ -17,10 +17,10 @@ jobs:
 
   nightly:
     name: Deploy nightly
+    needs: build-job
```
MeirGavish, line 20: @ricfehr3 I think so, does it mean the nightly will require the unit tests as well?

ricfehr3: Yes, this makes it so it has to pass the three in that workflow:, clang-format, build the rom, and unit tests. I think this is the best way.

Issue #277, opened 2025-12-04
[BUG] Joker shop price not properly erased on purchase


MeirGavish:
**Describe the bug**
When purchasing a joker whose cost requires 2 digits, the price is not properly erased and the last digit stays. Currently the only such jokers are Blueprint and Brainstorm costing $10.

**To Reproduce**
Buy Blueprint or Brainstorm from the shop. This can be easily done by setting the define
`#define TEST_JOKER_ID0 BLUEPRINT_JOKER_ID` so Brainstorm will surely appear in the shop (requires #276 to be merged)

**Expected behavior**
The last digit ("0") should be erased after purchase.

**Screenshots**

https://github.com/user-attachments/assets/bde5d611-c453-4881-bdef-23b23a23d58b

**What version**
Main branch with latest commit [Clang format and refactor game.h/c + update Bracket Alignment to BlockIndent (#259)](https://github.com/cellos51/balatro-gba/commit/291edaf4f255d44a8511f01c35eae21ba9f6404e)

**What Emulator or Platform were you playing on**
mgba

PR #286, opened 2025-12-06
[Fixed] Joker shop price not properly erased on purchase

blackcan1122:
regarding #277 

Implemented a way to calculate Sprite Size

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -3540,7 +3540,7 @@ static void erase_price_under_sprite_object(SpriteObject* sprite_object)
     {
         // fallback
         height = CARD_SPRITE_SIZE;
-        width = CARD_SPRITE_SIZE / 2;
+        width = CARD_SPRITE_SIZE;
     }
 
     Rect price_rect;


MeirGavish:
@blackcan1122 Thanks for fixing this! Can you add a screenshot/video showing it's fixed? Can you also check the prices under jokers when highlighting them for sale get displayed and erased properly and maybe add screenshots/video for that as well?

blackcan1122:
No problem i will provide Screenshots / Videos
is there a macro (#define) for forcing sales on jokers?

blackcan1122:
Here is a video:

https://youtu.be/aYmp1kqfAMk

Edit: sorry i misunderstood, i thought you meant when jokers are on Sale but you meant when i sell jokers
in the new video i show both

source/sprite.c
```diff
@@ -278,6 +278,36 @@ void sprite_object_set_focus(SpriteObject* sprite_object, bool focus)
...
+        return;
+    }
+
+    OBJ_ATTR* obj = sprite_object->sprite->obj;
+
+    int shape = (obj->attr0 >> 14) & 0x3;
+    int size = (obj->attr1 >> 14) & 0x3;
+
+    // todo: define as a global
+    int size_table[3][4][2] = {
```
ricfehr3: Since this is only here, we don't need to define it as a global. Rather `static const` should be sufficient to get the same benefits but only in the scope of this function.

source/sprite.c
```diff
@@ -278,6 +278,36 @@ void sprite_object_set_focus(SpriteObject* sprite_object, bool focus)
...
+    {
+        *height = 0;
+        *width = 0;
+        return;
+    }
+
+    OBJ_ATTR* obj = sprite_object->sprite->obj;
+
+    int shape = (obj->attr0 >> 14) & 0x3;
+    int size = (obj->attr1 >> 14) & 0x3;
```
ricfehr3: These should use the built in libtonc functions from `tonc_oam.h`:

```c
INLINE const u8 *obj_get_size(const OBJ_ATTR *obj);
INLINE int obj_get_width(const OBJ_ATTR *obj);
INLINE int obj_get_height(const OBJ_ATTR *obj);
```

source/game.c
```diff
@@ -3517,22 +3517,29 @@ static void game_round_end_dismiss_round_end_panel()
...
+
     Rect price_rect;
     price_rect.left = fx2int(sprite_object->tx);
-    price_rect.top = fx2int(sprite_object->ty) + CARD_SPRITE_SIZE + TILE_SIZE;
-    // TODO: Should probably extract the size from the sprite
-
-    price_rect.right = price_rect.left + TILE_SIZE * 3;
-    // Taking 2 tiles down so the highlighted case is also covered
-    price_rect.bottom = price_rect.top + 2 * TILE_SIZE;
+    // todo: fix the + 4 offset, don't know why its needed
```
ricfehr3: Please don't add nebulous magic fixes. The value 4 needs some context to it's purpose in either a variable/macro name or a comment.

ricfehr3 (changes requested):
Hey, thank you for this PR. I left a couple comments below. Once they are good, I think I'm good to approve. Thanks!

MeirGavish:
Requesting a review  from Copilot as well

MeirGavish:
Note that copilot reviews should be taken with a grain of salt, it's more meant to find things we missed.

source/sprite.c
blackcan1122: with the switch to use 
```c
INLINE const u8 *obj_get_size(const OBJ_ATTR *obj);
INLINE int obj_get_width(const OBJ_ATTR *obj);
INLINE int obj_get_height(const OBJ_ATTR *obj);
```

we could get rid of this table entirely

blackcan1122:
https://github.com/user-attachments/assets/4674e1fa-e397-4cb3-afab-821353895940

I figured out where the mysterious “+4” magic number in the erase logic came from.

As a small test, I stored the card’s Y value globally when the shop was generated and then used that Y value to define the erase rectangle. This worked, so it seems that the floating/hover animation also changes the sprite’s Y position when it runs.

To make the erase logic robust, I added the full maximum selection height (10 px) to the bottom of the rectangle. That way, no matter when this function is called (during the animation or after it finishes), we always erase the text correctly.

Edit:

i moved the define macros from sprite.c to sprite.h so they are accesseable in other files, i hope you guys are fine with it

source/sprite.c
```diff
@@ -278,6 +275,44 @@ void sprite_object_set_focus(SpriteObject* sprite_object, bool focus)
...
+        sprite_object->sprite->obj == NULL || width == NULL || height == NULL)
+    {
+        return false;
+    }
+
+    const u8* size = obj_get_size(sprite_object->sprite->obj);
+    *width = size[0];
+    *height = size[1];
+    return true;
+}
```
MeirGavish, line 341: Can you add another level of wrapping and add `sprite_get_dimensions` etc. functions to the sprite struct and delegate the functions there? It just makes sense since the dimensions are not exclusive to the sprite objects and any sprite should be able to use them.

source/game.c
```diff
@@ -3517,22 +3517,38 @@ static void game_round_end_dismiss_round_end_panel()
...
 static void erase_price_under_sprite_object(SpriteObject* sprite_object)
 {
+    int width = 0;
+    int height = 0;
+
+    if (sprite_object_get_dimensions(sprite_object, &width, &height) == false)
+    {
+        // fallback
+        height = CARD_SPRITE_SIZE;
+        width = CARD_SPRITE_SIZE / 2;
```
MeirGavish: Why fallback width to `CARD_SPRITE_SIZE / 2` and not `CARD_SPRITE_SIZE`? It should be the number returned by the get dimensions function.

MeirGavish:
> I figured out where the mysterious “+4” magic number in the erase logic came from.
> 
> As a small test, I stored the card’s Y value globally when the shop was generated and then used that Y value to define the erase rectangle. This worked, so it seems that the floating/hover animation also changes the sprite’s Y position when it runs.
> 
> To make the erase logic robust, I added the full maximum selection height (10 px) to the bottom of the rectangle. That way, no matter when this function is called (during the animation or after it finishes), we always erase the text correctly.


Yes, there was a comment there about covering the case where the sprite is raised.

source/sprite.c
blackcan1122: Sure no problem! :)

source/game.c
blackcan1122: I haven't actually tested the fallbacks specifically yet, but due to the name I expected that CARD_SPRITE_SIZE would be square dimensions, so I roughly approximated that half the width of the square would be the width of the card. If you want i can test the Fallbacks too :)

I also don't 100% know when this fallback would trigger in a real scenario, but I like to at least have some visual feedback that something may be broken or unhandled. Since there is no log or anything, I can't print an error to it.

MeirGavish: No need to test the fallback, it is not likely to trigger but it still needs to be correct if it's in the code. The sprite dimensions returned by the function will be 32x32 actually (the card sprite itself has a width of 24 but it's not one of the GBA sprite sizes so we use 32x32 with transparency). Just change it to `CARD_SPRITE_SIZE` since that is the value returned by the function.

blackcan1122: I see that makes sense. Its changed
Yes i also when this case triggers something deeper down the game broke or went wrong, but i like to try to handle edge cases when somehow possible :)

MeirGavish (approved):

Issue #275, opened 2025-12-03
Add a C standard to the makefile


MeirGavish:
I never noticed we don't have a standard in the makefile, I think it makes sense to add one so the project is explicit about which one it conforms to make sure the compiler doesn't default to a standard we don't support. 
My gut feeling tells me to have the latest so we support all the latest features and conform to the latest practices.
However, we need to also make sure the libraries we're using (namely tonc) conforms to the standard and that the standard is supported by the arm GCC compiler we're using.

From [this discussion](https://github.com/cellos51/balatro-gba/pull/259#discussion_r2581372875)

ricfehr3:
It looks like the default c standard provided by devkit pro is C23:

```sh
echo | /opt/devkitpro/devkitARM/bin/arm-none-eabi-gcc -dM -E -x c - | grep STDC_VERSION
#define __STDC_VERSION__ 202311L
```

Since DevkitPro provides the compiler and libtonc, I think it's safe to assume that C23 should be fine for this project.

MeirGavish:
Thanks!

Adding the flag `-std=c23` generates the following errors.

```
C:/devkitPro/libtonc/include/tonc_tte.h:472:20: error: implicit declaration of function 'iprintf'; did you mean 'sprintf'? [-Wimplicit-function-declaration]
  472 | #define tte_printf iprintf
      |                    ^~~~~~~
D:/Data/Code/GBA/Balatro_main_repo/balatro-gba/source/game.c:988:5: note: in expansion of macro 'tte_printf'
  988 |     tte_printf(
      |     ^~~~~~~~~~
D:/Data/Code/GBA/Balatro_main_repo/balatro-gba/source/game.c: In function 'game_shop_get_rand_available_joker_id':
D:/Data/Code/GBA/Balatro_main_repo/balatro-gba/source/game.c:3551:31: error: implicit declaration of function 'random'; did you mean 'rand'? [-Wimplicit-function-declaration]
 3551 |     int fallback_random_idx = random() % jokers_avail_size;
      |                               ^~~~~~
      |                               rand
```

I did wonder about the usage of `random()` instead of `rand()`, however the `iprintf()` error is from tonc and we can't (or at least shouldn't) override it.

From https://gcc.gnu.org/onlinedocs/gcc/Standards.html:

[...]

PR #288, opened 2025-12-07
Added C standard gnu23 to the makefile

MeirGavish:
See conversation in related issue #275


MeirGavish:
Converting to draft for now, maybe it should be added for all the makefiles for the test modules.

ricfehr3:
> Converting to draft for now, maybe it should be added for all the makefiles for the test modules.

I think we should add it to all the makefiles and see if there are any errors with the test. The differences in architecture shouldn't effect anything if they adhere to the same standard. If there is an issue, then we can see what's up.

MeirGavish:
> > Converting to draft for now, maybe it should be added for all the makefiles for the test modules.
> 
> I think we should add it to all the makefiles and see if there are any errors with the test. The differences in architecture shouldn't effect anything if they adhere to the same standard. If there is an issue, then we can see what's up.


@ricfehr3 Okay, so I just tried it and the tests fail, maybe the test image is not running the latest version of gcc?

ricfehr3:
> > > Converting to draft for now, maybe it should be added for all the makefiles for the test modules.
> > 
> > 
> > I think we should add it to all the makefiles and see if there are any errors with the test. The differences in architecture shouldn't effect anything if they adhere to the same standard. If there is an issue, then we can see what's up.
> 
> @ricfehr3 Okay, so I just tried it and the tests fail, maybe the test image is not running the latest version of gcc?

It looks like ubuntu 24.04.3 still uses gcc 13.3.0, where `gnu23` was experimental, so they called it `gnu2x`, and that works. So the question there is, do we introduce this experimental gnu23 flag? I say yes, it's a trivial upgrade once the gcc version is updated.

MeirGavish:
> > > > Converting to draft for now, maybe it should be added for all the makefiles for the test modules.
> > > 
> > > 
> > > I think we should add it to all the makefiles and see if there are any errors with the test. The differences in architecture shouldn't effect anything if they adhere to the same standard. If there is an issue, then we can see what's up.
> > 
> > 
> > @ricfehr3 Okay, so I just tried it and the tests fail, maybe the test image is not running the latest version of gcc?
> 
> It looks like ubuntu 24.04.3 still uses gcc 13.3.0, where `gnu23` was experimental, so they called it `gnu2x`, and that works. So the question there is, do we introduce this experimental gnu23 flag? I say yes, it's a trivial upgrade once the gcc version is updated.

I'm trying to maybe add the latest version to the workflow so it explicitly matches the the one  in the main makefile.

```
sudo apt-get install -y gcc-14
sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 100
```

I just thought the experimental version flag could break once it's upgraded. The thing is that now we are also explicit about the GCC version used and that would mean we don't always necessarily use the latest and it may happen that someday `ubuntu-latest` doesn't support gcc-14 anymore so I'm not sure about this.

MeirGavish:
I think the proper thing to do will be to make a script that checks the version and if it's less than 14, install 14 and switch to it.

MeirGavish:
> I think the proper thing to do will be to make a script that checks the version and if it's less than 14, install 14 and switch to it.

Done!

tests/run_tests.sh
```diff
@@ -2,6 +2,20 @@
 
 set -eu
 
+# Check GCC version and install GCC 14 if needed
+check_gcc_version() {
```
ricfehr3: Ideally we would like to run this locally and not mess with developers existing configurations. I think we should do one of two things.

1. Set an environment variable in the CI to run this function to install it. Otherwise, we should assume someone is running this on their host machine and instead it should have a friendly warning. 

2. We could also put it back in the CI, and remove it when the day comes that GCC is on the correct version. I would say I honestly prefer that to eliminate all possibility of someone running this on accident and causing headaches. I do think keeping this check and outputting a nice and clear error with a non-zero exit code and not running the tests would be the correct behavior.

MeirGavish: You're right, good point! Maybe separate this from the test script to another script and run that on the CI? We could still add an environment variable too in case it is run by a developer on a host machine. Otherwise, we could go with option 2.

ricfehr3: I like the idea of moving it into another script and running it in the CI. I think it would be the most stable option. Changing to `2x` would deprecate fast. Moving it into the workflow could work, but I think it's complexity deserves a script that can be called in the CI. 

No need for an environment variable, rather just move that script into a CI directory, should be pretty obvious to not run anything from there (I hope).

The last thing to cover is the poor developer who is running the tests on the latest ubuntu version right now. I think a good piece of documentation would be fine for that. If you would like to here (or another PR), a README in the `tests` directory with this information would be sufficient I feel. Just let them know to install gcc-14.

MeirGavish: Done, added the README too, I made it pretty minimal.

ricfehr3: Awesome, looks great!

PR #308, opened 2025-12-11
Used proper text centering for price printing

MeirGavish:
Following #286 I changed the price text centering from `get_digits_even()` to use the proper `update_text_rect_to_center_str()` and finally removed `get_digits_even()`. This also consolidates the code for price printing and erasing rect to use the same function, making it more correct.
This also improves the centering for uneven number of characters, namely "$10" for Brainstorm and Blueprint

image

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -142,26 +142,4 @@ static inline int u32_get_digits(uint32_t n)
     return 10;
 }
 
-/**
- * @brief Get the number of digits closest to it's "even" value, i.e. ceil(digits/2)
- *
- * Useful for centering text on tiles
- *
- * @param n value to find the number of even decimal digits of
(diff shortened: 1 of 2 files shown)


include/util.h
```diff
@@ -142,26 +142,4 @@ static inline int u32_get_digits(uint32_t n)
...
-    if (n < 100)
-        return 1;
-    if (n < 10000)
-        return 2;
-    if (n < 1000000)
-        return 3;
-    if (n < 100000000)
-        return 4;
-    return 5;
-}
```
ricfehr3, line 165: Rip to a real one ✋ ☹️

MeirGavish: Lol

PR #312, opened 2025-12-12
Button sound effect

MeirGavish:
Added a button sound effect that plays on all button presses.
I also refactored `play_sfx()` to have a volume parameter so I can change the effects volume. It sounded too strong and felt like it needed to be more subtle.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -1678,8 +1678,8 @@ static void hand_set_focus(int index)
     }
 
     play_sfx(
-        SFX_CARD_FOCUS, 
-        MM_BASE_PITCH_RATE + rand() % CARD_FOCUS_SFX_PITCH_OFFSET_RANGE, 
+        SFX_CARD_FOCUS,
+        MM_BASE_PITCH_RATE + rand() % CARD_FOCUS_SFX_PITCH_OFFSET_RANGE,
         SFX_DEFAULT_VOLUME
     );
 }
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -307,8 +307,8 @@ void sprite_object_set_focus(SpriteObject* sprite_object, bool focus)
     sprite_object->focused = focus;
 
     play_sfx(
-        SFX_CARD_FOCUS, 
-        MM_BASE_PITCH_RATE + rand() % CARD_FOCUS_SFX_PITCH_OFFSET_RANGE, 
+        SFX_CARD_FOCUS,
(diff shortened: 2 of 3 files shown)


source/sprite.c
```diff
@@ -306,7 +306,7 @@ void sprite_object_set_focus(SpriteObject* sprite_object, bool focus)
     }
     sprite_object->focused = focus;
 
-    play_sfx(SFX_CARD_FOCUS, MM_BASE_PITCH_RATE + rand() % 512);
+    play_sfx(SFX_CARD_FOCUS, MM_BASE_PITCH_RATE + rand() % 512, SFX_DEFAULT_VOLUME);
```
ricfehr3: I think the 512 should be named.

source/game.c
```diff
@@ -4426,6 +4454,15 @@ static void game_over_anim_frame(void)
     main_bg_se_move_rect_1_tile_vert(GAME_OVER_ANIM_RECT, SCREEN_UP);
 }
 
+static void game_over_process_user_input()
```
ricfehr3: Nitpick, `inline` this function.

source/game.c
```diff
@@ -1677,7 +1678,7 @@ static void hand_set_focus(int index)
         selection_x = index;
     }
 
-    play_sfx(SFX_CARD_FOCUS, MM_BASE_PITCH_RATE + rand() % 512);
+    play_sfx(SFX_CARD_FOCUS, MM_BASE_PITCH_RATE + rand() % 512, SFX_DEFAULT_VOLUME);
```
ricfehr3: Same as this [comment](https://github.com/GBALATRO/balatro-gba/pull/312/changes#r2615979655), name the value `512`.

source/game.c
```diff
@@ -199,6 +199,7 @@ static void game_blind_select_on_update(void);
 static void game_blind_select_on_exit(void);
 static void game_lose_on_init(void);
 static void game_lose_on_update(void);
+static void game_over_process_user_input();
```
ricfehr3: If you chose to inline this from the other [comment](https://github.com/GBALATRO/balatro-gba/pull/312/changes#r2615981463), then this should be removed.

source/sprite.c
MeirGavish: I'll add a name for it

source/game.c
MeirGavish: Sure, just want to confirm the reasoning - it's a small function with not many call sites and this will save function call overhead, anything else?

ricfehr3: Exactly that. I figure we should take as many opportunities as possible to optimize.

Issue #231, opened 2025-11-15
Sound effects for accumulating chips and menu popup 


MeirGavish:
Add sound effects for the menu pop outs and when accumulating chips during the round from the played hand to the total.

MeirGavish:
For the menu popup it seems like in Balatro there is a sound effect for blind select but not for round end.

Issue #230, opened 2025-11-15
Money Sound Effects


MeirGavish:
Add money related sound effects for the following events
1. Earning money during the round (from jokers etc.)
2. Accumulating money during the round end animation
3. The final cashout at the round end animation
4. Selling jokers
5. Buying jokers

Effects can be reused for some events like in Balatro, 3 coin effects should do the trick.
The effects from Balatro have high frequencies in the coin sounds and when I tried importing them to the GBA  it seems like they didn't sound very good because those frequencies were lost.

MeirGavish:
I tried exporting the coin effects in a higher sample rate but Maxmod seems to still downsamples them and I couldn't find a setting to change that. I also tried pitch-shifting them but it just sounds bad.

MeirGavish:
I tried doing the pitch-shift trick from #313, by maybe trying to pitch shift 3x instead of 2x, but it's just too much, it sounds bad.
I'm removing my assignment for now unless I come up with some new solution.

poweredbygamespy:
Hey :) I uploaded edited versions of the coin1, coin3, and coin6 SFX in #131 with the high frequency components extracted and transposed down to avoid aliasing.

Alternatively, I tried increasing the Maxmod sample rate at run time, but that yielded little improvement in the audible frequency range and used ~20% more RAM for audio mixing buffers...

I'd be happy to turn this into a PR if the SFX is acceptable.

poweredbygamespy:
[balatro-coin.zip](https://github.com/user-attachments/files/29672370/balatro-coin.zip)

I could also try the GBA's square wave generator, that might sound more like a coin but it would be hard to make it sound similar to the original

Mega-Evolved-Ziquinho:
> I could also try the GBA's square wave generator, that might sound more like a coin but it would be hard to make it sound similar to the original

I was thinking exactly this for a while. We only use 1 square wave channel for the music, and the GBA has 3 more, but it might be impossible to recreate the original game's SFX with just them. I'm not a music artist by any means, I just help to see the project go further

MeirGavish:
> [balatro-coin.zip](https://github.com/user-attachments/files/29672370/balatro-coin.zip)
> 
> I could also try the GBA's square wave generator, that might sound more like a coin but it would be hard to make it sound similar to the original

Thanks! Actually I think these don't sound too bad, is that what they sound like in the game?
[...]

PR #313, opened 2025-12-12
Sound effects for score accumulation and blind menu pop-up

MeirGavish:
🛑 Based on PR #312

For issue #231 

Added sound effects for score calculation, score accumulation (lerping), and blind select menu pop-up.
For the score accumulation I once again had the issue of the sound having lots of high frequency content and downsampling ruining it. I managed to greatly improve it by pitch shifting it in Audacity and then pitch shifting it back in the code. It's not perfect but I think it's acceptable. I will try doing this for money sound effects for issue #230.

I tested it on original hardware and it does seem like some of the sound effects are not audible on the speaker but when plugging in earphones everything sounds good.

In Balatro's sound effects,

- chips_accum is chips2

- cancel is an unchanged name, maybe I should give it a different name such as "pop" since it's never used in any cancel...
Edit: renamed it "pop"


source/game.c
```diff
@@ -54,14 +55,13 @@
...
+#define BUTTON_SFX_VOLUME 154 // 60% of MM_FULL_VOLUME
+
 #define STARTING_ROUND 0
 #define STARTING_ANTE  1
 #define STARTING_MONEY 4
 #define STARTING_SCORE 0
 
-#define TEN_K 10000
-#define ONE_K 1000
-
```
MeirGavish: These were unused leftovers

source/game.c
```diff
@@ -195,10 +195,12 @@ static void game_round_end_on_update(void);
 static void game_round_end_on_exit(void);
 static void game_shop_on_update(void);
 static void game_shop_on_exit(void);
+static void game_blind_select_on_init(void);
 static void game_blind_select_on_update(void);
 static void game_blind_select_on_exit(void);
 static void game_lose_on_init(void);
 static void game_lose_on_update(void);
+static void game_over_process_user_input();
```
```suggestion
```

This was a merge error

MeirGavish:
> Looks and sounds good. My only comment would be changing the names of "SFX_CANCEL" and "SFX_BUTTON", however with the workflow with maxmod, I think what you have here is fine since it's commented. Merge when you're ready o7.

I renamed "cancel" to "pop" but for the button I don't want to rename it since it is the same effect and it's used more for buttons.

PR #282, opened 2025-12-05
Added blind score requirement and reward in blind select screen

MeirGavish:
I added score requirement and reward display in the blind select screen. I used the score truncation to squeeze them into 3 tiles there.
I also did a small refactor to rename the `blinds` array into `blind_states` to make it a little clearer.

https://github.com/user-attachments/assets/5be6c95a-c973-4cb7-be56-4fe11eea9e48

Since the truncation is so extreme and rounded down, it's sometimes a little awkward like on ante 2 where it looks like both the Big Blind and Boss Blind have the same score requirement of 1K when in fact it's 1200 and 1600.

image

Another screenshot of the infamous ante 5
image

And on ante 8 there is inevitable overflow but it's handled as well as possible.

https://github.com/user-attachments/assets/08073de4-1400-48a7-82e6-f12313e9f346

PR #314, opened 2025-12-13
Introduce decimal point font

### LINKED ISSUE #295: Add custom font to allow compactly displaying decimal point
Opening an issue for a proposal in a comment for PR #282.
This will also solve issue #177.

_Originally posted by @ricfehr3 in https://github.com/cellos51/balatro-gba/issues/282#issuecomment-3618404295_

The lack of making `1.2K` for example bugs me a bit. I thought of a (__maybe unrealistic__) idea. Loading a custom font actually isn't too difficult. We could have a copy of the standard font, then map unused characters to custom, squished versions of the numbers text with decimals: "1.","2.",..."9.","0." or the decimal first ".0",".1"... etc. So, for example, we could map the following (Looking at the [tonc tutorial](https://gbadev.net/tonc/tte.html#ssec-test-se4), I don't think we use these characters):

1. '&' <-> '.0'
2. '^' <-> '.1'
3. '{' <-> '.2'
4. '}' <-> '.3'
5. '|' <-> '.4'
6. '`' <-> '.5'
7. '<' <-> '.6'
8. '>' <-> '.7'
9. '_' <-> '.8'
10.';' <-> '.9'

I have an [example of modifying the font](https://github.com/ricfehr3/balatro-gba/commit/faa4948daa297726be5740ee233f1a7619e4a5bf). It's a commit branched from this branch right now, so if you want to apply as a [patch](https://github.com/ricfehr3/balatro-gba/commit/faa4948daa297726be5740ee233f1a7619e4a5bf.patch). 

This would require:

- [ ] Artwork for all the numbers.
- [ ] Some code mapping the characters.

ricfehr3:
Closes #295 

This adds custom font with special characters replaced with ".0" to ".9". 

I've moved the font into it's own file since it seemed appropriate and decoupled `tonc.h` from `util.h`'s tests.

### DIFF
--- a/include/font.h
+++ b/include/font.h
@@ -23,7 +23,7 @@ extern const unsigned int gbalatro_sys8Glyphs[192];
  *
  *  For example:
  *  ```c
- *  tte_printf("Testing " XSTR(FP0) " Something!");
+ *  tte_printf("Testing " XSTR(FP0_CHAR) " Something!");
  *  ```
  *  prints "Testing .0 Something!"
  *


include/font.h
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @brief Get the decimal point string equivalent of 0-9 from integer param.
+ *
+ * @param val Value between 0-9 to get ".0" to ".9" equivalents. Note that
+ *            if a value outside of range is passed it will perform '% 9'
+ *            on that value.
+ *
+ * @return A char* to the string values of '.0' to '.9'
+ */
+const char* get_font_point_str(int val);
```
ricfehr3, line 68: The reason for having both the XSTR method and this function is to hopefully simplify the work required to complete #282 using `snprintf`, but also give the flexibility of building it in at compile time.

MeirGavish:
Great job! I think this could really solve our decimal point problems.

So, I guess I should have pointed it out when first seeing the image is that what worries me is that it is inconsistent with the existing font numbers as well as letters. The position of the decimal point reduces 2 whole pixels from the width of the characters making them so very thin in comparison to the original font which has quite the thick lines and it doesn't look very good.

Also, in this example it seems like .2 and .3 are swapped.

image

So what I noticed is that almost all characters in the original font don't touch the right most pixel column of the tile, so I figured we could give some more space to the digits themselves by pushing the decimal point one pixel to the left and it won't merge with any other preceding character. Here is a draft for a resulting font,

image

And enlarged
image

I'm still not fully happy with it, I made most digits thinner but 1, and 9 managed to escape it by not touching that pixel. I think I will give them the same treatment for consistency, so 6 is an inverted version of 9 etc.
Other than that, I think the other digit characters also need to be modified to look the same as the decimal point ones but without the decimal point so it looks fully consistent.

Is the tool for creating the font binary a command line too? Can you add it to the makefile chain so it generates sys8_custom.s from sys8_custom.png and you can just make changes to the image, make and see the results in the compiled rom? Then I think the .s file should be in the build directory too.

This is what stuck out to me first but I will also give a more in-depth review to the code.

source/font.c
```diff
@@ -0,0 +1,22 @@
...
+    XSTR(FP5),
+    XSTR(FP6),
+    XSTR(FP7),
+    XSTR(FP8),
+    XSTR(FP9),
+};
+
+const char* get_font_point_str(int val)
+{
+    val = val & 9;
```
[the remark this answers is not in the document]
MeirGavish: And since it's good to be defensive, I'm also wondering about negative values here

ricfehr3:
> Great job! I think this could really solve our decimal point problems.
> 
> So, I guess I should have pointed it out when first seeing the image is that what worries me is that it is inconsistent with the existing font numbers as well as letters. The position of the decimal point reduces 2 whole pixels from the width of the characters making them so very thin in comparison to the original font which has quite the thick lines and it doesn't look very good.
> 
> Also, in this example it seems like .2 and .3 are swapped.
> image
> 
> So what I noticed is that almost all characters in the original font don't touch the right most pixel column of the tile, so I figured we could give some more space to the digits themselves by pushing the decimal point one pixel to the left and it won't merge with any other preceding character. Here is a draft for a resulting font,
> image
> 
> And enlarged image
> 
> I'm still not fully happy with it, I made most digits thinner but 1, and 9 managed to escape it by not touching that pixel. I think I will give them the same treatment for consistency, so 6 is an inverted version of 9 etc. Other than that, I think the other digit characters also need to be modified to look the same as the decimal point ones but without the decimal point so it looks fully consistent.
> 
> Is the tool for creating the font binary a command line too? Can you add it to the makefile chain so it generates sys8_custom.s from sys8_custom.png and you can just make changes to the image, make and see the results in the compiled rom? Then I think the .s file should be in the build directory too.
> 
> This is what stuck out to me first but I will also give a more in-depth review to the code.

I definitely agree, it doesn't look that good with the font. It was more of a test, but I think your text already looks much better. I wrote this wiki page so if you wanted to generate and test different fonts you can: https://github.com/GBALATRO/balatro-gba/wiki/How-To:-Generate-Custom-Font . Thanks for providing a draft btw!

I think only the usenti tool is able to do the text output. The workflow is a bit more cumbersome, but it probably will be pretty static once we settle on something. We can totally do an entire custom font at this point now as well, if we wanted.

I'll add that fix for the order of the values and check out what copilot said.

MathisMartin31:
How about taking this opportunity to make the font look a bit more like Balatro's? Here's my attempt, including both digits and characters

image
<br>
image

The digits with and without the dot are also the same at 5x7 px

MeirGavish:
> How about taking this opportunity to make the font look a bit more like Balatro's? Here's my attempt, including both digits and characters

> The digits with and without the dot are also the same at 5x7 px

I think that's a good idea, I think we should just start with the digits here and then do the rest of the letters in another PR.
Your font looks great! From a quick look only the B and 8 need a bit of rounding in the middle in my opinion (just bump a pixel in), the 8 looks like a stroked 0 like that.

MathisMartin31:
> From a quick look only the B and 8 need a bit of rounding in the middle in my opinion (just bump a pixel in), the 8 looks like a stroked 0 like that.

Yeah, I did that to match Balatro's font but at this scale it doesn't look good ><

image
<br>
image

MeirGavish:
> I definitely agree, it doesn't look that good with the font. It was more of a test, but I think your text already looks much better. I wrote this wiki page so if you wanted to generate and test different fonts you can: https://github.com/GBALATRO/balatro-gba/wiki/How-To:-Generate-Custom-Font . Thanks for providing a draft btw!
> 
> I think only the usenti tool is able to do the text output. The workflow is a bit more cumbersome, but it probably will be pretty static once we settle on something. We can totally do an entire custom font at this point now as well, if we wanted.
> 
> I'll add that fix for the order of the values and check out what copilot said.

Thanks! I missed the wiki page. 😅
Really annoying you can't export the font with a command line tool but I guess that's what we have to work with for now...
So I played around with the new consistent digits in-game (I also changed 5 a little bit to look better and be more consistent with Balatro's m6x11 font)

The spacing between digits does feel a little bit awkward but I think it's still good enough and acceptable for the benefit of being able to accurately display large numbers.

image
image
image
image

source/main.c
```diff
@@ -18,6 +18,9 @@
 #include "soundbank.h"
 #include "soundbank_bin.h"
 
+extern const TFont sys8_customFont;
+extern const unsigned int sys8_customGlyphs[192];
```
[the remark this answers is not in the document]
MeirGavish: It seems like usenti is exporting a header, it would be good to leave it in that header.

MathisMartin31:
Yeah the digits are a bit thin but the alternative would be them touching the dot, which already small enough by itself that touching the digits, even if only by a corner, would make it practically invisible ^^'

source/font.c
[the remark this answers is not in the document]
ricfehr3: Oops, yeah that's an oopsie on the modulo. @MeirGavish negative numbers are a good call. I'll ABS the value.

source/main.c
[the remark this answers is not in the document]
ricfehr3: That makes sense. We would need to enforce that the file is named the same so the headers match when generating a new font... actually, I think I'm just going to make a tool to do this. It won't be too flexible of a tool, but it seems pretty straightforward how the bitmap is setup.

ricfehr3:
@MeirGavish I have added a script to autogenerate font. It isn't very flexible, neither is the addition to the Makefile, however, it does streamline the font generation. However... it's not working well with the CI right now. I'm going to come back to this one, but feel free to check out the script and come to me with any ideas on how to fix the build in the CI. It requires a python library (PIL), that isn't available in the compose container. I'll need to think on this one. I did choose python for compatibility, though. It could be better suited as a devtool to make the file, I'm thinking. Sort of ruins the automation approach though... Not sure, brains a little fried. haha. 

@MathisMartin31 @MeirGavish Just wanted to say amazing work on getting the font looking nice. That was really quick and it looks good. I'm using the most recent version of that in the new template for the build.

MeirGavish:
For my testing I used this font with just the digits altered and updated 5,

sys8_custom

I guess you can also use this font with the rest of the characters altered to be similar to Balatro's m6x11 font, I guess we can merge it and give it a closer look later.
balatro-gba-font

MathisMartin31:
> I guess you can also use this font with the rest of the characters altered to be similar to Balatro's m6x11 font, I guess we can merge it and give it a closer look later.
> sys8_custom

It seems you posted the same font twice ? ^^'

MeirGavish:
> > I guess you can also use this font with the rest of the characters altered to be similar to Balatro's m6x11 font, I guess we can merge it and give it a closer look later.
> > sys8_custom
> 
> It seems you posted the same font twice ? ^^'

Right, good catch, thanks! I edited it with your font.

MeirGavish:
> @MeirGavish I have added a script to autogenerate font. It isn't very flexible, neither is the addition to the Makefile, however, it does streamline the font generation. However... it's not working well with the CI right now. I'm going to come back to this one, but feel free to check out the script and come to me with any ideas on how to fix the build in the CI. It requires a python library (PIL), that isn't available in the compose container. I'll need to think on this one. I did choose python for compatibility, though. It could be better suited as a devtool to make the file, I'm thinking. Sort of ruins the automation approach though... Not sure, brains a little fried. haha.

Oh wow, you made our own font generator script from scratch that completely bypasses usenti, that's really cool!
It seems like the issue with the script in the CI is that it's missing PIL/Pillow, I'll try adding a pip install.

MeirGavish:
> > @MeirGavish I have added a script to autogenerate font. It isn't very flexible, neither is the addition to the Makefile, however, it does streamline the font generation. However... it's not working well with the CI right now. I'm going to come back to this one, but feel free to check out the script and come to me with any ideas on how to fix the build in the CI. It requires a python library (PIL), that isn't available in the compose container. I'll need to think on this one. I did choose python for compatibility, though. It could be better suited as a devtool to make the file, I'm thinking. Sort of ruins the automation approach though... Not sure, brains a little fried. haha.
> 
> Oh wow, you made our own font generator script from scratch that completely bypasses usenti, that's really cool! It seems like the issue with the script in the CI is that it's missing PIL/Pillow, I'll try adding a pip install.

No pip, tried adding ensure-pip... this is getting a bit out of hand and I don't know if I want to continue trying to trash your PR... also these should probably be extracted to another script rather than stacked into a big command like I did.

font/gbalatro_sys8.png
MeirGavish, line 1: Is this the large version of the image? It should be the original size one.

ricfehr3: Oops, yes that should be the smaller one.

ricfehr3:
@MeirGavish Okay, so! Came back to this with a fresh(er) mind. What I ended up doing to fix that pillow issue was to add a custom Dockerfile that just pulls down the devkitpro image and adds on python3-willow. I chose python3-willow because it has all the dependencies needed for pillow, but at a package level (with apt) and not using pip. It's good too, because the docker image is cached anyway, so it only installs once per image, and that image only changes if the compose file changes.

I also have the makefile almost where it needs to be. I'm not going to lie, it's a bit of a headache, so I'm going to fix it once I can come back to it. However, it isn't required for this PR, it just limits the font to be generated ONLY from the `font/gbalatro_sys8.png`.

@MathisMartin31 Thank you again for coming in clutch with the font. It looks good! And we can also with this PR we can just drop the font into the same file and rebuild easily.

font/gbalatro_sys8.png
MeirGavish, line 1: This is not the version with the updated "5", also I see a stray pixel in the ' character that I missed and snuck in if you could remove that too.

.gitignore
```diff
@@ -1,15 +1,15 @@
...
+*.sln
+*.vcxproj*
+/balatro-gba/x64/**
+/tests/bitset/build
+/tests/list/build
+/tests/pool/build
+doc
+tests/util/.vscode
+tests/util/build
+/font/*.s
```
MeirGavish: You removed the audio/*.bak line I added, please add it back

```suggestion
audio/*.bak
/font/*.s
```

MeirGavish: Also, I think it would be better to put the .s file in the build directory along with the rest of the .s files from grit etc.
I think this will also simplify your makefile.

include/font.h
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @brief Utility file to interact with custom font
+ */
+#ifndef FONT_H
+#define FONT_H
+
+#include <tonc.h>
+
+extern const TFont gbalatro_sys8Font;
+extern const unsigned int gbalatro_sys8Glyphs[192];
```
MeirGavish, line 12: So we're not generating a header too? Maybe we should make a static header for this then...

include/font.h
```diff
@@ -0,0 +1,68 @@
...
+#define FP8 _ // .8
+#define FP9 ; // .9
+/** @} */
+
+/**
+ * @brief Get the decimal point string equivalent of 0-9 from integer param.
+ *
+ * @param val Value between 0-9 to get ".0" to ".9" equivalents. Note that
+ *            if a value outside of range is passed it will perform '% 10'
+ *            on that value.
```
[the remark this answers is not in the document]
MeirGavish: Meh, this is an edge case, maybe just add "and the last decimal digit will be used." and call it a day.
Edit: linking [Copilot's comment](https://github.com/GBALATRO/balatro-gba/pull/314#discussion_r2625872840) because the comment appears on its own too.

font/gbalatro_sys8.png
MeirGavish: The more I look at the font, the more I'm thinking this should only have the digits and the rest of the characters should be the same as the sys8 for now and be made similar to Balatro's in another PR. I'm a little worried about the descenders, characters like "j" touch others below them in the splash screen.
image

MeirGavish (changes requested):
I'm giving this a review, didn't look at the scripts and code in depth

font/gbalatro_sys8.png
MathisMartin31: That's weird, the only problematic characters are the ones with a little tail (y, p, q, g and j), and they are not too different than the base sys8 font. Are you sure the problem isn't currently present ?

image

But yes this is probably an issue that isn't entirely related to this PR, let's postpone this

include/font.h
ricfehr3: I don't really see a point in separating it in it's own header. It will always be named the same thing, and it would simply take these two lines and move it to another file that would be included in `font.c`. I'm open for an argument on that though

.gitignore
ricfehr3: I was initially having problems with putting it in the build directory, but I got it working now. Thanks!

font/gbalatro_sys8.png
ricfehr3: If we have an updated font file I can add it to this PR? Otherwise I can keep it at the default sys8 and we can do another PR to update the font? Either way is fine for me. For now, I did remove that extra pixel in the '`' and I think I've got the right version this time too.

include/font.h
MeirGavish: Oh, I thought this was on main.c, I think it's fine if it's in the font.h header.

MeirGavish: Oh, I thought this was on main.c, I think it's fine if it's in the font.h header.

include/font.h
```diff
@@ -0,0 +1,68 @@
...
+#define FP0 & // .0
+#define FP1 ^ // .1
+#define FP2 { // .2
+#define FP3 } // .3
+#define FP4 | // .4
+#define FP5 ` // .5
+#define FP6 < // .6
+#define FP7 > // .7
+#define FP8 _ // .8
+#define FP9 ; // .9
```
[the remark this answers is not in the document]
MeirGavish: @ricfehr3 Is the reasoning for not just defining it as strings so that we can do something like `'FP3_CHAR'` to get a single char rather than string?

include/font.h
```diff
@@ -0,0 +1,70 @@
...
+ *
+ *  "FP#" -> Font "Point" "number"
+ *
+ *  This is used at compile time directly with the help of libtonc's XSTR() macro
+ *  **Note**: This should **ONLY** be used with the XSTR() macro, otherwise it
+ *  is undefined behavior.
+ *
+ *  For example:
+ *  ```c
+ *  tte_printf("Testing " XSTR(FP0) " Something!");
```
MeirGavish:
```suggestion
 *  tte_printf("Testing " XSTR(FP0_CHAR) " Something!");
```

Just update the example too

font/gbalatro_sys8.png
MeirGavish: > That's weird, the only problematic characters are the ones with a little tail (y, p, q, g and j), and they are not too different than the base sys8 font. Are you sure the problem isn't currently present ?
> But yes this is probably an issue that isn't entirely related to this PR, let's postpone this

Yeah, you're right, it's already like this and it's fine. I opened an [issue for this](https://github.com/GBALATRO/balatro-gba/issues/325#issue-3741422708) so we can stress about changing most  of the text in another PR.

PR #315, opened 2025-12-15
Fixed bug where the mult wasn't truncated

MeirGavish:
There is a bug where the mult isn't truncated, seems like the change from #269 wasn't properly merged/rebased.

image

I even have a picture on real hardware from a production build.

This PR fixes the issue.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -1022,7 +1022,7 @@ void display_chips(void)
     check_flaming_score();
 }
 
-void display_mult()
+void display_mult(void)
 {
     Rect mult_text_overflow_rect = MULT_TEXT_RECT;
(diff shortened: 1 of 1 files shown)


source/game.c
```diff
@@ -1022,16 +1022,24 @@ void display_chips(void)
     check_flaming_score();
 }
 
-void display_mult(void)
+void display_mult()
```
ricfehr3: Out of curiosity why remove the `void` here?

MeirGavish: Mistake, I think I imported the changes from another commit and missed this

```suggestion
void display_mult(void)
```

PR #321, opened 2025-12-16
Music - Fixed wrong G/D chord sample

MeirGavish:
Apparently I made a mistake with one of the chord samples. The chord is only used once in the progression so it's subtle and I missed it but now I can't unhear it.
I've also been switching to work with OpenMPT so I added its backup files to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ _codeql_detected_source_root
 doc
 tests/util/.vscode
 tests/util/build
+audio/*.bak


MeirGavish:
@ricfehr3 I'm merging this, I don't want to bother you with straining to hear the difference.

ricfehr3:
> @ricfehr3 I'm merging this, I don't want to bother you with straining to hear the difference.

LOL, sounds good.

PR #322, opened 2025-12-16
Music - up to measure 46

### LINKED ISSUE #382: Music - up to measure 46
Up to and including the "triad chords" where I think it's a good looping point.

MeirGavish:
So I finally managed to get some progress on the music for #110.
~~This is still not finished, it has 32 out of 62 measures so about half way there but I figured I will open a draft PR so I can document some of my changes and people can see/listen to my changes as I make progress.
This also shouldn't be merged unfinished because it doesn't end in a good place to loop so it sounds awkward.~~

Edit: Okay, so I added up to measure 45 (incl.) finishing with what I called the "triad chords" section. I made a bit more progress but I think this is actually a good looping point and it includes all the main recognizable musical features of the track so I put those in another branch and I'm marking this PR as ready.

I "expanded" the track so it still runs at the same tempo but now a single row is a 1/32 rather than a 16th note which allows for smaller subdivisions. I did that to try to make the triplets from measure 29, it's still just an approximation because I can't divide by 3 but I think it sounds good enough. I switched to OpenMPT for this since it was much more comfortable to do that there, there's literally a button for it. I think I'll continue with OpenMPT from here.
I also made a few small changes to the existing lead melody:
1. I added portamento so some parts of the melody slide smoothly into each other rather than change immediately.
2. I added a few note stops that were missing.
3. I added some fade-out to long notes to make it feel more natural and breathable.

More importantly, I added a new instrument and sample I called "Echo Quasi-Square" because the instrument envelope makes it echo and it sounds like a chiptune square wave even though it's a sample (how I wish the Gameboy PSGs were accessible in the MaxMod pipeline). I took it from FearOfDark's "Rolling Down The Street In My Katamari" xm module, I hope it's okay I'm using it. I used this instrument a lot and I think it adds a lot of color.

One thing I used it for is panned backing chords that come in when the first melody repeats. They exist in the original Balatro theme and they are quite iconic in my opinion but they're not in MrCrimit's transcription.

Also I added 2 more channels, so we're now up to 8 channels, this allowed making the aforementioned chords with a single sample/instrument on multiple channels rather than making whole samples for the chords like I did before. It's much better in terms of memory and workflow. I also needed these channels to better layer the instruments and have instruments enter and fade properly without conflicting on each other's channels.
From my testing it seems like adding the channels didn't add any performance or audio artifacts and I think this is very much worth the musical improvements.
I think it would be good to keep it at 8 channels, this is also the number of channels in the SNES.

The ROM is now up to 4.7MB.

I will appreciate if anybody has time to listen and give their opinions before I merge.

I'm uploading a zip with the compiled rom to make it easier.
[balatro-gba.zip](https://github.com/user-attachments/files/25747489/balatro-gba.zip)


MeirGavish:
At this point I'm thinking of adding more channels. I'm wondering whether to merge the current version before doing that in case it has performance impact.

MeirGavish:
Okay, I added more channels and it doesn't seem like performance is impacted.

MeirGavish:
I'm marking this ready for review, I'll appreciate if anybody has time to listen and give their opinion.

GreenNinja2525:
> I'm marking this ready for review, I'll appreciate if anybody has time to listen and give their opinion.

I think it sounds really good! I'm not an expert on music, but it definitely feels more impactful and I like it a lot. I think you did a great job with it! @MeirGavish

MeirGavish:
> > I'm marking this ready for review, I'll appreciate if anybody has time to listen and give their opinion.
> 
> I think it sounds really good! I'm not an expert on music, but it definitely feels more impactful and I like it a lot. I think you did a great job with it! @MeirGavish

@GreenNinja2525 
Thanks! Glad you liked it!

MeirGavish:
I'm merging this, I usually prefer to not squash the music commits because the changes are in a binary file that won't show in the differ. However, these are 31 commits and I don't want to spam the git history with so much non-code commits so I'll squash.

Issue #326, opened 2025-12-18
Refactor round hand selection to use the selection grid


MeirGavish:
This will also allow joker swapping during the round

MeirGavish:
Opened two issues by mistake, closing as duplicate of #320

PR #329, opened 2025-12-18
Change selection grid to maintain relative horizontal position when switching rows

MeirGavish:
For issue #326, the hand selection switches to the Play/Discard buttons according to your position in the hand so to maintain this behavior I made this change for the selection grid.
This isn't a perfect way to do this, a better way is to use the actual positions of the items on the screen, Balatro does this. But since I don't have that instead I assume they are somewhat evenly aligned and use the relative position within the row.

https://github.com/user-attachments/assets/044563be-ca36-4609-883d-26f3a60308dc

@ricfehr3 I want to test something with this PR, once it's ready, try merging it without approving just to see if you can bypass the approval.

### DIFF
--- a/source/selection_grid.c
+++ b/source/selection_grid.c
@@ -58,8 +58,13 @@ void selection_grid_move_selection_vert(SelectionGrid* selection_grid, int direc
             return;
 
         int old_row_size = selection_grid->rows[selection.y].get_size();
-        // TODO: Maybe some of the fixed conversions here are unnecessary
-        new_selection.x = fx2int(fxmul(int2fx(selection.x), (int2fx(new_row_size) / old_row_size)));
+
+        // Branchless set to 1 if 0 to avoid division by 0
+        old_row_size += (old_row_size == 0);
+
+        // Maintain relative horizontal position
+        // The operations are equivalent to fixed point if all the numbers were converted
+        new_selection.x = fx2int(selection.x * ((int2fx(new_row_size) / old_row_size)));
 
         if (selection.y >= 0 && selection.y < selection_grid->num_rows)
         {


source/selection_grid.c
```diff
@@ -57,12 +57,10 @@ void selection_grid_move_selection_vert(SelectionGrid* selection_grid, int direc
...
 
-        if (selection.x >= new_row_size)
-        {
-            // TODO: Maintain relative horizontal position
-            // Clip selection to row size
-            new_selection.x = new_row_size - 1;
-        }
+        int old_row_size = selection_grid->rows[selection.y].get_size();
+        // TODO: Maybe some of the fixed conversions here are unnecessary
+        new_selection.x = fx2int(fxmul(int2fx(selection.x), (int2fx(new_row_size) / old_row_size)));
```
ricfehr3: Here, if `old_row_size == 0`, then there will be a divide by 0 error. I think we should have a check for that being 0. Or (because I've been looking at branchless programming lately, lol). We could set the value to `1` if it's 0 without `if` statements with:

```c
old_row_size += (old_row_size == 0); 
```

We don't have to do that, but some sort of check is needed.

ricfehr3 (changes requested):

source/selection_grid.c
MeirGavish: You're right, good catch! It seems like the compiler gives 0 here as a result of division by 0 but it's bad to rely on that since it's undefined behavior.
I did it branchless as you suggested and added comments, I also removed some of the fixed point conversions/operations as suggested by my TODO here.

MeirGavish:
> Approving and merging now that I can. I won't merge anything without permission going forward.

@ricfehr3 
You could approve and merge my (and others') PRs beforehand too, just not your own because GitHub doesn't let you approve your own PRs.
I think it's good to always have someone else review your code before merging and it's good that GitHub encourages that but I also think admins/maintainers should be able to merge their own PRs. The checkbox makes it very obvious you're bypassing rules so you know not to abuse it but if there's something that shouldn't wait for a review since it's urgent etc. you should merge it.

As for other PRs, with time I want to be able to let go a bit and let you merge more without my review and relying just on yours, for now I am specifying those PRs explicitly.

Issue #327, opened 2025-12-18
Update the truncate number function to use the decimal point


MeirGavish:
This will allow us to accurately represent big numbers using the custom decimal point font.
This will also fix #177.

MeirGavish:
I have something functional but it's a little messy and inefficient and I still want to optimize and clean it up before making a PR.
Thinking about it, I think we might want to change the current implementation a bit. 
It uses up all the space to preserve accuracy so for example for 5 characters it will truncate 1234000 -> 1234K.
I think we would now want to truncate it to 1.234M since it now preserves accuracy and even though it uses the same number of characters it's more readable.

Other than that, I want to still be able to test the function and that introduces some issues when trying to make the test code independent from GBA code.
For one, the code in util.c/.h now relies on the code in font.c/.h which does break dependency a bit because the font is quite specific to our implementation so it doesn't feel very correct but I don't want to lose the ability to test this function.
To make util and font independent I added the following to util.h
```
#ifndef XSTR
#define STR(x)		#x
// Expand and then stringify
#define XSTR(x)		STR(x)
#endif // XSTR
```

I did this because using __XSTRING which I thought should be available from stdlib.h did not compile for me, I suppose the macro is not always available.

Also, since the extern definitions use `TFont` from tonc, I will need to move it from font.h to a dedicated header.
```
extern const TFont gbalatro_sys8Font;
extern const unsigned int gbalatro_sys8Glyphs[192];
```

Issue #177, opened 2025-10-25
You can lose despite getting the displayed required score due to rounding


MeirGavish:
When displaying high scores, the numbers are displayed with a suffix notation to preserve screen space and stay within the 1-tile per character limit.
The Big Blind score requirement is 1.5x ante base, in ante 5 it's 16500 but it rounds down and displays 16k, if you get exactly 16000 your score would also display 16k but you still wouldn't win.

_Originally posted by @MathisMartin31 in https://github.com/cellos51/balatro-gba/discussions/148#discussioncomment-14476770_

<img width="524" height="366" alt="Image" src="https://github.com/user-attachments/assets/6953e474-c8d8-4a4e-97bc-d6592a16ecbe" />

I'm trying to think what would be the correct thing to do about this. The options that come to mind are
1. Round the actual score down and then the big blind requirement would be exactly 16000
2. Round the displayed blind requirement up but the player score down, then both would not be accurate but at least you won't get a case where it looks like you win but you didn't. Instead this would trade it off for a state where you win even though it doesn't look like you did.
3. ~~Use proper integer rounding for the displayed score.~~ Actually I'm ruling this one out since rounding the player score will create a worse issue where the blind requirement is for example 10k and the player has 9500 but rounded up is displayed 10k.

Mega-Evolved-Ziquinho:
I'm not one of the people developing this game, but "bug" is not an accurate tag. It's still absolutely working as intended, it's just that this intended behaviour doesn't come off as such due to the GBA's rather limited sprite limit

Edit: 

> I'm not one of the people developing this game, [...]
Made a first draft of the Acrobat joker sprite if that counts

MeirGavish:
Yeah, you could still say it's a sort-of-kind-of bug, I don't know what other label would fit and it feels awkward to make a new overly specific label for this.

MeirGavish:
As of #269 this issue is slightly better. The score box now uses 5 character tiles so it can display the player's score accurately in this case but the blind requirement is still truncated since it only has 4 tiles.

<img width="480" height="320" alt="Image" src="https://github.com/user-attachments/assets/d1aadc29-6416-49c4-948c-9355ce107a01" />

PR #333, opened 2025-12-19
Update the truncate number function to use decimal point

MeirGavish:
For issue #327 (See my comments there)

This also finally solves issue #177 and it looks great!

image

The thing that's been bothering me is score lerping because it displays many numbers quickly and if there's not enough time the score isn't printed.

Here's what it looked like unoptimized,

https://github.com/user-attachments/assets/67ac02b2-7e67-4f9e-8e40-6afd461b5aa8

But after some optimizations,

https://github.com/user-attachments/assets/f231e2c5-ffcb-4bd8-98ea-1e573ce58743

### DIFF
--- a/source/util.c
+++ b/source/util.c
@@ -26,7 +26,7 @@ int int_arr_max(int int_arr[], int size)
  *        Currently static single use, may be unstaticed if needed but use with caution.
  *
  * @param num_str The string to remove trailing zeros from, modified in-place.
- * @param size The size of the string - strlen is not used and no checks are performed, 
+ * @param size The size of the string - strlen is not used and no checks are performed,
  *             the function relies on the caller to provide the correct size,
  *             if it's larger than the actual string length or negative it will result in
  *             an invalid write.


source/game.c
```diff
@@ -2986,7 +2985,8 @@ static inline void game_playing_process_input_and_state(void)
             );
         }
     }
-    else if (play_state == PLAY_ENDED)
+    // TODO: Remove magic number
```
```suggestion
```

Remove this TODO

include/font.h
MeirGavish, line 1: Seems like I messed merging and managing the git history... I'll try to resolve this...

tests/util/util_test.c
```diff
@@ -111,12 +198,12 @@ void test_truncate_uint_to_suffixed_str()
 
     suffixed_str_buff[0] = '\0';
     truncate_uint_to_suffixed_str(UINT32_MAX, 4, suffixed_str_buff);
-    assert(strcmp(suffixed_str_buff, "4B") == 0);
+    assert(strcmp(suffixed_str_buff, "4" XSTR(FP2_CHAR) "9B") == 0);
```
MeirGavish:
```suggestion
    assert(strcmp(suffixed_str_buff, "4" XSTR(FP2_CHAR) "9B") == 0); // "4.29B"
```
Add a clear number comment here

include/font.h
MeirGavish: Done, simple rebase

source/font.c
```diff
@@ -17,6 +19,11 @@ static const char* s_font_point_lookup[] = {
...
 {
-    val = ABS(val) % 10;
+    val = abs(val) % 10;
     return s_font_point_lookup[val];
 }
+
+char digit_char_to_font_point(char digit_char)
+{
+    return get_font_point_str(digit_char - '0')[0];
+}
```
MeirGavish, line 29: I added this function to convert chars, it's the case I ended up needing.
@ricfehr3 Considering I couldn't use the macro literals to define char constants, maybe it would be better to just avoid using XSTR altogether and just directly define the strings instead of literals?

```
#define FP0_STR "&" // .0
#define FP1_STR "^" // .1
#define FP2_STR "}" // .2
...
```

include/util.h
```diff
@@ -15,6 +15,12 @@
  */
 #define GBAL_UNUSED __attribute__((unused))
 
+#ifndef XSTR
+#define STR(x) #x
+// Expand and then stringify
+#define XSTR(x) STR(x)
+#endif // XSTR
```
MeirGavish: I did this to make util and font independent from tonc and because using __XSTRING which I thought should be available from stdlib.h did not compile for me, I suppose the macro is not always available.

Maybe we can avoid use of `XSTR` altogether and this can be removed. See my [other comment](https://github.com/GBALATRO/balatro-gba/pull/333#discussion_r2637790212).

source/game.c
```diff
@@ -118,7 +118,11 @@
 #define SHOP_BOTTOM_PANEL_BORDER_PID         26
 // Naming the stage where cards return from the discard pile to the deck "undiscard"
 
-#define NUM_SCORE_LERP_STEPS 32
+/* This needs to stay a power of 2 and small enough
+ * for the lerping to be done before the next hand is drawn.
+ */
+#define NUM_SCORE_LERP_STEPS   16
+#define TM_SCORE_LERP_INTERVAL 2
```
MeirGavish: Maybe these optimizations can be extracted to another PR?

source/game.c
```diff
@@ -2986,7 +2985,7 @@ static inline void game_playing_process_input_and_state(void)
             );
         }
     }
-    else if (play_state == PLAY_ENDED)
+    else if (play_state == PLAY_ENDED && timer % FRAMES(TM_SCORE_LERP_INTERVAL) == 0)
```
MeirGavish: Likewise here, see [other comment](https://github.com/GBALATRO/balatro-gba/pull/333#discussion_r2637790754)

source/game.c
```diff
@@ -2998,10 +2997,11 @@ static inline void game_playing_process_input_and_state(void)
...
         if (lerped_temp_score > 0)
         {
-            display_temp_score(fx2uint(lerped_temp_score));
+            // Set the score display first because it's more important
+            // in case there isn't enough time within the frame to display both
+            display_score(fx2uint(lerped_score));
 
-            // We actually don't need to erase this because the score only increases
-            display_score(fx2uint(lerped_score)); // Set the score display
+            display_temp_score(fx2uint(lerped_temp_score));
```
MeirGavish, line 3003: Another optimization I'm wondering about.
Edit: So switching the order here is what made the most difference in how it looks. I'm not fully sure why but this is the optimization that has to stay.

source/main.c
```diff
@@ -33,7 +34,10 @@ void init()
         CLR_WHITE,
         TTE_BIT_UNPACK_OFFSET,
         &gbalatro_sys8Font,
-        NULL
+
+        // Explicitly use 8x8 tile text drawing function to improve performance
+        // See https://gbadev.net/tonc/tte.html#ssec-map-reg
+        se_drawg_w8h8
```
MeirGavish, line 40: And this optimization too.
Edit: I think it's good to have this optimization regardless, we're pretty set on using 8x8 tiles for text for now and this should improve performance.

source/util.c
```diff
@@ -18,45 +21,131 @@ int int_arr_max(int int_arr[], int size)
...
+ *        as a NULL-terminated string; may be empty if nothing remains.
+ */
+static inline void truncate_num_get_remainder_string(
+    uint32_t decimal_remainder,
+    uint32_t truncated_num,
     int num_req_chars,
-    char out_str_buff[UINT_MAX_DIGITS + 1]
+    char suffix_char,
+    char remainder_str[UINT_MAX_DIGITS + 1]
 )
```
MeirGavish, line 67: This is a single use function but I felt like it needed some documentation because it has a lot of parameters and it might not be completely obvious what it does and how it behaves.

include/font.h
```diff
@@ -67,4 +62,15 @@ extern const unsigned int gbalatro_sys8Glyphs[192];
...
 const char* get_font_point_str(int val);
 
+/**
+ * @brief Get the decimal point char equivalent of 0-9 from char param.
+ *
+ * @param digit_char A char of a digit between '0'-'9' to get '.0' to '.9' equivalents. Note that
+ *                   if a char outside of range is passed it the return value will be within limits
+ *                   but is not defined with any relationship to the input.
+ *
+ * @return A char representing a value in font within the range of '.0' to '.9'
```
MeirGavish:
```suggestion
 * @return A char representing a value in the font within the range of '.0' to '.9'
```

source/game.c
MeirGavish: I removed this one from this PR.

include/gbalatro_sys8.h
MeirGavish, line 1: Since the extern definitions use TFont from tonc, I needed to move it from font.h to a dedicated header.

MeirGavish:
Pretty happy with this now, opening for reviews.

source/util.c
```diff
@@ -18,45 +21,131 @@ int int_arr_max(int int_arr[], int size)
...
+            remainder_str_format = "%lu";
     }
 
-    int num_digits = u32_get_digits(num);
-    int overflow_size = num_digits - num_req_chars;
-    char* suffix = "";
+    snprintf(remainder_str, UINT_MAX_DIGITS + 1, remainder_str_format, decimal_remainder);
+
+    // Truncate overflow
+    int remaining_chars = num_req_chars - u32_get_digits(truncated_num) - 1; // - 1 for suffix
```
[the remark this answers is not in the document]
MeirGavish: Good point, committing.

source/font.c
ricfehr3: That makes way more sense to do it that way. I gave it a quick test If you want to do it here, go for it. Otherwise I gave it a quick test. Works great:

```diff
diff --git a/include/font.h b/include/font.h
index 23c97a5..41761f1 100644
--- a/include/font.h
+++ b/include/font.h
@@ -44,16 +44,16 @@ extern const unsigned int gbalatro_sys8Glyphs[192];
  *
  * @{
  */
-#define FP0_CHAR & // .0
-#define FP1_CHAR ^ // .1
-#define FP2_CHAR } // .2
-#define FP3_CHAR { // .3
-#define FP4_CHAR | // .4
-#define FP5_CHAR ` // .5
-#define FP6_CHAR < // .6
-#define FP7_CHAR > // .7
-#define FP8_CHAR _ // .8
-#define FP9_CHAR ; // .9
+#define FP0_CHAR "&" // .0
+#define FP1_CHAR "^" // .1
+#define FP2_CHAR "}" // .2
+#define FP3_CHAR "{" // .3
+#define FP4_CHAR "|" // .4
+#define FP5_CHAR "`" // .5
+#define FP6_CHAR "<" // .6
+#define FP7_CHAR ">" // .7
+#define FP8_CHAR "_" // .8
+#define FP9_CHAR ";" // .9
 /** @} */
 
 /**
diff --git a/source/font.c b/source/font.c
index 7220357..519e5b4 100644
--- a/source/font.c
+++ b/source/font.c
@@ -3,16 +3,16 @@
 #include "tonc.h"
 
 static const char* s_font_point_lookup[] = {
-    XSTR(FP0_CHAR),
-    XSTR(FP1_CHAR),
-    XSTR(FP2_CHAR),
-    XSTR(FP3_CHAR),
-    XSTR(FP4_CHAR),
-    XSTR(FP5_CHAR),
-    XSTR(FP6_CHAR),
-    XSTR(FP7_CHAR),
-    XSTR(FP8_CHAR),
-    XSTR(FP9_CHAR),
+    FP0_CHAR,
+    FP1_CHAR,
+    FP2_CHAR,
+    FP3_CHAR,
+    FP4_CHAR,
+    FP5_CHAR,
+    FP6_CHAR,
+    FP7_CHAR,
+    FP8_CHAR,
+    FP9_CHAR,
 };
 
 const char* get_font_point_str(int val)

```

include/util.h
ricfehr3: I agree, let's get rid of it.

source/util.c
ricfehr3: Always great to have some help understanding these complex functions. Thanks for the docs!

include/gbalatro_sys8.h
ricfehr3: All good. I like only having files that absolutely need the `tonc.h` header as well.

include/font.h
```diff
@@ -67,4 +62,15 @@ extern const unsigned int gbalatro_sys8Glyphs[192];
...
+/**
+ * @brief Get the decimal point char equivalent of 0-9 from char param.
+ *
+ * @param digit_char A char of a digit between '0'-'9' to get '.0' to '.9' equivalents. Note that
+ *                   if a char outside of range is passed it the return value will be within limits
+ *                   but is not defined with any relationship to the input.
+ *
+ * @return A char representing a value in the font within the range of '.0' to '.9'
+ */
+char digit_char_to_font_point(char digit_char);
```
ricfehr3, line 71: Good rename here 👍

source/main.c
ricfehr3: This is a great catch, and yeah, looking at the webpage it's a just good to have it here. Nice!

tests/util/util_test.c
```diff
@@ -1,3 +1,4 @@
+#include <font.h>
```
ricfehr3, line 1: Thanks for updating the test! You can remove all the XSTR stuff if you decide to add that change to remove the need for it

source/util.c
```diff
@@ -18,45 +21,144 @@ int int_arr_max(int int_arr[], int size)
     return max;
 }
 
-void truncate_uint_to_suffixed_str(
-    uint32_t num,
+// Can be unstaticed if needed
+static inline void num_str_truncate_trailing_zeros(char* num_str, int size)
```
ricfehr3, line 34: Since it's an inline function for a specific call, I think it's fine to keep this way, but if `size` is passes as anything greater than `strlen(num_str) + 1` it will dereferenced memory out-of-bounds of `num_str`. Maybe just a comment warning it isn't doing a check would be fine, or you can add a check. Your call.

source/font.c
MeirGavish: Done.

include/font.h
MeirGavish: @ricfehr3 Thanks! Just notice this isn't a rename, I kept `get_font_point_str()` as an int-to-string function and added this as a char-to-char function.

tests/util/util_test.c
MeirGavish: Done! It was important to update this test and add cases, it helped me catch some bugs while working on this.

source/util.c
MeirGavish: I'll add full documentation in case this is needed elsewhere

MeirGavish: Done.

include/font.h
ricfehr3: Ah, I missed that.

source/game.c
```diff
@@ -2998,10 +2996,11 @@ static inline void game_playing_process_input_and_state(void)
...
```
ricfehr3, line 3003: I'm a little confused at the reason behind switching these around. Does it matter?

MeirGavish: @ricfehr3 
Me too, I had [another comment](https://github.com/GBALATRO/balatro-gba/pull/333#discussion_r2637791239) and this is what made the most difference. My original rationale was that if there isn't enough time to display both the total score and the temp score, the total score is more important and it would look better if the temp score is the one who isn't displayed this frame. I added a comment about why I'm doing this, I guess it's not clear enough?
Either way, I'm not sure why it made such a big difference but it completely removed the issue.

ricfehr3: Okay, it took me a minute to get, but I see now. lol, that's my bad. I think you can leave the comment as is.

PR #336, opened 2025-12-21
[Optimization] Slow down score lerping from 60 to 30 FPS

MeirGavish:
Extracted and separated from #333.

Slowed score lerping so it happens every other frame instead of every frame. The idea was to give the CPU more time to render the score and avoid blank scores mid-lerping but this doesn't seem to make such a big difference in that regard.
I think it still looks good at this rate and maybe even better because you can better see the score accumulating?

Some videos to show the difference.

**Main**

https://github.com/user-attachments/assets/801b42df-1a23-48fe-b996-2d0b688d4c96

https://github.com/user-attachments/assets/28580cda-209c-4df6-9348-15a4010739aa

**PR**

https://github.com/user-attachments/assets/9854906b-abb0-447b-b940-f523de644e20

https://github.com/user-attachments/assets/7b4ffab6-15bd-45fc-b66b-7d24e49f7477

https://github.com/user-attachments/assets/8692b2a0-3f40-4e66-a373-ad3161d850f1

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -121,7 +121,8 @@
 /* This needs to stay a power of 2 and small enough
  * for the lerping to be done before the next hand is drawn.
  */
-#define NUM_SCORE_LERP_STEPS 32
+#define NUM_SCORE_LERP_STEPS   16
+#define TM_SCORE_LERP_INTERVAL 2
 
 // Shop
 #define REROLL_BASE_COST 5 // Base cost for rerolling the shop items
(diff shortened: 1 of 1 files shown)


MathisMartin31:
Maybe it's just me but... I really can't tell the difference/what the issue was ><

MeirGavish:
> Maybe it's just me but... I really can't tell the difference/what the issue was ><

The difference is subtle, the numbers change faster currently and slower with the PR, I think it's a good thing that it's hard to tell. It's easier to see the difference on the higher scores.
This was originally a part of #333 to fix the issues shown in the video there, but I ended up fixing them without this so I separated it to another PR in case we still want this optimization, I think it could be nice to have.

MathisMartin31:
Oh I do see that it changes more slowly, but I couldn't spot the issue that you are fixing with this PR ^^'. But if it's fixed that's good :)

PR #338, opened 2025-12-22
Introduce the game variables struct

ricfehr3:
@MeirGavish I have been messing with work on #156 . It's been a "give a mouse a cookie" situation where one change bubbles up to a huge list of commits. 

It would be ridiculous to introduce all these changes at once, but I think I can break the commits down to make it a lot more digestible. 

This PR doesn't do anything special besides add the `GameVariables` struct and supplying a pointer to to all callbacks from the `STATE_TABLE`. The reason this is chosen first is because:

1. It doesn't effect any logic
2. A good percentage of these callbacks will need this struct, so it's best to just pass a pointer it to all callbacks rather waste cycles with a `getter()` function. 

The immediate next plan is to move a small subsystem out (I was working on blinds, but it becomes a mess, so I'll try to find something smaller first) and iterate from there. This change will help get the process started. Anyway, let me know what you think.


include/game.h
```diff
@@ -30,6 +30,10 @@ typedef struct CardObject CardObject;
 typedef struct Card Card;
 typedef struct JokerObject JokerObject;
 
+typedef struct
+{
+} GameVariables;
```
MeirGavish: Just add a comment stating this struct is meant to contain global variables that will be put into it in a future refactor.

include/game.h
```diff
@@ -98,9 +102,9 @@ enum HandType
 typedef struct
 {
     int substate;
-    void (*on_init)();
-    void (*on_update)();
-    void (*on_exit)();
+    void (*on_init)(GameVariables* vars);
+    void (*on_update)(GameVariables* vars);
+    void (*on_exit)(GameVariables* vars);
```
MeirGavish: Maybe it's time to typedef the function prototype?

source/game.c
```diff
@@ -184,7 +184,7 @@ typedef struct
 // efficient. Well, this is the answer.
 // Thanks!
 // https://github.com/cellos51/balatro-gba/issues/137#issuecomment-3322485129
-static void noop(void)
+static void noop(GameVariables* vars)
```
MeirGavish, line 192: And maybe rename this `noop_state_func` or something else that would fit the typedef name.

ricfehr3:
> I think it makes sense because the state functions are very generic. I'm wondering whether we should avoid passing this variable to too many functions because I think the idea should be to decouple so for example functions that do things specific to the round like scoring a hand don't have to rely on variables related to the shop for example. That doesn't necessarily mean the variable shouldn't be passed, it's just something to keep in mind. Another example for decoupling would be to allow a hand to exist without a deck to draw cards from, then we could create a hand and score it with jokers for testing without having to initialize unrelated things such as the deck. I'm rambling a bit because this PR has very little changes and has more to do with design.
> 
> I'm requesting changes mostly for documentation.

Yeah, this is the foundational step in this refactor, so it's great to have discussions about this.

My reason for passing the entire GameVariables struct is because:
1. The only alternative I can think of is a `getter()` function for it, and this has less overhead.
2. With that, it would be easier to ignore the argument for functions that don't need it, verses functions that do.
3. We could pass a `void* ptr` instead and keep a generic callback type as well. This may have some benefits for expanding on state machines, making them more generic, but that is becoming **too** generic for this project. Passing a `void* ptr` is pretty standard: for example, [freertos tasks](https://www.freertos.org/Documentation/02-Kernel/02-Kernel-features/01-Tasks-and-co-routines/05-Implementing-a-task). But if it's going to be casted to `GameVariables*` anyway, I think it makes sense to just keep the type instead of generalizing it to `void*`.

There definitely is some room to decouple systems more, and I'm open to any suggestions. With the GameVariables struct, it would be functionally the same as having all those static vars in the top of `game.c`. Because of that, I feel comfortable replacing those variables since the logic is already sound. Maybe afterwards we can clean up further?

include/game.h
ricfehr3: Absolutely!

source/game.c
ricfehr3: Done, that name makes sense.

MeirGavish:
> > I think it makes sense because the state functions are very generic. I'm wondering whether we should avoid passing this variable to too many functions because I think the idea should be to decouple so for example functions that do things specific to the round like scoring a hand don't have to rely on variables related to the shop for example. That doesn't necessarily mean the variable shouldn't be passed, it's just something to keep in mind. Another example for decoupling would be to allow a hand to exist without a deck to draw cards from, then we could create a hand and score it with jokers for testing without having to initialize unrelated things such as the deck. I'm rambling a bit because this PR has very little changes and has more to do with design.
> > I'm requesting changes mostly for documentation.
> 
> Yeah, this is the foundational step in this refactor, so it's great to have discussions about this.
> 
> My reason for passing the entire GameVariables struct is because:
> 
> 1. The only alternative I can think of is a `getter()` function for it, and this has less overhead.
> 2. With that, it would be easier to ignore the argument for functions that don't need it, verses functions that do.
> 3. We could pass a `void* ptr` instead and keep a generic callback type as well. This may have some benefits for expanding on state machines, making them more generic, but that is becoming **too** generic for this project. Passing a `void* ptr` is pretty standard: for example, [freertos tasks](https://www.freertos.org/Documentation/02-Kernel/02-Kernel-features/01-Tasks-and-co-routines/05-Implementing-a-task). But if it's going to be casted to `GameVariables*` anyway, I think it makes sense to just keep the type instead of generalizing it to `void*`.
> 
> There definitely is some room to decouple systems more, and I'm open to any suggestions. With the GameVariables struct, it would be functionally the same as having all those static vars in the top of `game.c`. Because of that, I feel comfortable replacing those variables since the logic is already sound. Maybe afterwards we can clean up further?

Yeah, I also think `void*` doesn't quite fit here. I guess it's mainly a question of what we put in this struct and how it will be passed around so I'll write some thoughts. 
I think it makes sense to put things global to the run - money, round, ante, jokers, blinds. There are global variables for each state that shouldn't be there, they should be moved to their respective files.
For other things I am not sure - for the hand for example, I think it makes sense to have getters. What about score, chips and mult? They are only valid for the round itself, it doesn't make sense to pass them to every state. Other states could call `display_chips() and `display_mult()` if they want to just update the display to 0 since it's always there. Though `display_score()` must receive an argument, we can split it to `display_curr_score()` that would be a wrapper to `display_score()`.

I'm trying to look at how would a refactor would look for an example function - `game_playing_handle_round_over()`, it makes sense that this function would have this parameter because it accesses the blind and the ante. So this means it will need it passed from the caller `game_playing_discarded_cards_loop()` which will then need it passed from `game_playing_on_update()`. Does it mean all functions would need this parameter? I think not because that would be a lot to change and I'm sure some functions don't need it.

MeirGavish:
Okay, so after all this, I just want to add a more practical conclusion for right now. I think it would be better to make it a getter for this despite the overhead because it will be simpler to code, you don't need to add parameters to so many functions. But I guess making it a parameter would also be okay...

ricfehr3:
> Okay, so after all this, I just want to add a more practical conclusion for right now. I think it would be better to make it a getter for this despite the overhead because it will be simpler to code, you don't need to add parameters to so many functions. But I guess making it a parameter would also be okay...

Do you still feel strongly about the getter @MeirGavish ? I still think it makes more sense to pass it as a parameter, especially after seeing it in use [here](https://github.com/GBALATRO/balatro-gba/pull/363). I hold my opinion because most every function will be using the getter, which cost cycles, as opposed to other functions ignoring it, which is free.

MeirGavish:
> > Okay, so after all this, I just want to add a more practical conclusion for right now. I think it would be better to make it a getter for this despite the overhead because it will be simpler to code, you don't need to add parameters to so many functions. But I guess making it a parameter would also be okay...
> 
> Do you still feel strongly about the getter @MeirGavish ? I still think it makes more sense to pass it as a parameter, especially after seeing it in use [here](https://github.com/GBALATRO/balatro-gba/pull/363). I hold my opinion because most every function will be using the getter, which cost cycles, as opposed to other functions ignoring it, which is free.

No, I don't feel strongly about it, if you still think it's better to pass a parameter that's okay.

include/game.h
```diff
@@ -1,6 +1,8 @@
 #ifndef GAME_H
 #define GAME_H
 
+#include "game_variables.h"
```
ricfehr3, line 4: This include will be moved out with the state machine code below.

ricfehr3:
@MeirGavish This one is ready, can you just approve it? Then I can add @emiyl as a coauthor in the squash merge?

```
Co-authored-by: Name <email@example.com>
```

Issue #341, opened 2025-12-22
[BUG] Possible Four Fingers Bug


MeirGavish:
**Describe the bug**
When playing a 4 card flush with Four Fingers, and an unscored card completes a pair, "hand contains" jokers for the pair, namely The Duo, activate even though the card is non-scoring. See video

**To Reproduce**
Have Four Fingers and the Duo and play described scenario.
Can be easily tested with `TEST_JOKER_ID0` and `TEST_JOKER_ID1`.

**Expected behavior**
I'm pretty sure The Duo shouldn't activate.

**Screenshots**

https://github.com/user-attachments/assets/f7ed24e5-9b4b-4bd4-bb1e-4a561135fa9c

**What version**
Main branch with latest commit "Change selection grid to maintain relative horizontal position when switching rows (https://github.com/GBALATRO/balatro-gba/pull/329)" https://github.com/GBALATRO/balatro-gba/commit/f76c25614ce0bc690be422eda0af7f6615dcc89c

**What Emulator or Platform were you playing on**
mGBA

**Additional context**
Possibly an issue in the hand-contains code. I'm not 100% sure this is a bug but I think it is.

MathisMartin31:
When I come back from vacation I could look at this and try to fix in my PR for contained hands optimization? Maybe there's something I can do there since I'm already doing a big refactor of this part of the code

MeirGavish:
> When I come back from vacation I could look at this and try to fix in my PR for contained hands optimization? Maybe there's something I can do there since I'm already doing a big refactor of this part of the code

I opened a PR for this issue, I requested your review for now in case you have time to take a look once you're back from vacation so you can give your input and whether you think this integrates well with your refactor PR.

MeirGavish:
So I had to make sure and apparently this is... not a bug?

https://github.com/user-attachments/assets/870a58c8-49a0-485b-b744-95c57987be0a

Edit: seed used for reference 94U5EXC6

MathisMartin31:
> So I had to make sure and apparently this is... not a bug?

Oh. That's unexpected, did we end up coding this whole thing the exact same way as Balatro?

But even if this is present in the base game it *feels* like a bug to me, it at least doesn't behave how the Jokers say it should, so I think we should still try to fix it. What do you think?

MathisMartin31:
> But even if this is present in the base game it *feels* like a bug to me, it at least doesn't behave how the Jokers say it should, so I think we should still try to fix it. What do you think?

[...]

PR #342, opened 2025-12-23
Fixed Four Fingers bug #341, may be incorrect behavior though

MeirGavish:
This is a possible fix for the possible bug #341.
It seems like `cards_in_hand_update_loop()` in the `HAND_PLAY` section calls functions to set the scored cards in the played hand as `selected` so the fix would be to check if the cards are selected in the hand distribution function.
This does fix the bug and I didn't see any other errors so far but I need to re-review this fix and give it more testing for more hand types and cases since it touches a central function for hand detection. It also feels like there is a bit of an abuse of the `selected` flag but I guess a bigger refactor would be out of the scope for this fix.

https://github.com/user-attachments/assets/4cb942ca-ffb3-41fa-bef2-0055926805d6

Edit: so I extracted the seemingly duplicate code and did more testing and it seems to work as expected but apparently this behavior is [not in line with Balatro](https://github.com/GBALATRO/balatro-gba/issues/341#issuecomment-3691363488).
I might extract the documentation to another PR and not merge this one.

There's only one hand played here, I forgot the recording on, sorry about the video length.

https://github.com/user-attachments/assets/360025ac-8690-414d-b404-2b3742dd0564

https://github.com/user-attachments/assets/21b648de-e028-4a67-abea-32f74c6f57cd

https://github.com/user-attachments/assets/baa1af59-b71c-4ddf-aa61-1829c4b4605d

### DIFF
--- a/source/hand_analysis.c
+++ b/source/hand_analysis.c
@@ -3,18 +3,39 @@
 #include "card.h"
 #include "game.h"
 
-void get_hand_distribution(u8* ranks_out, u8* suits_out)
+
+/**
+ * @brief Outputs the distribution of ranks and suits in an input card array.
+ *        Can be unstaticed if needed (and then this comment moved to hand_analysis.h)
+ *        Can also be refactored to use list but then both the hand and played stack would
+ *        need to use list as well.
+ * @param cards     An array of CardObject pointers
+ * @param num_cards The number of elements in \p cards.
+ * @param ranks_out output - updated such as ranks_out[rank] is the number of cards of rank in
+ *                  \p cards. Must be of size NUM_RANKS.
+ * @param suits_out output - updated such as suits_out[suit] is the number of cards if suit in the
+ *                  \p cards. Must be of size NUM_SUITS.
+ */
+static inline void get_card_distribution(
+    CardObject** cards,
+    int num_cards,
+    u8 ranks_out[NUM_RANKS],
+    u8 suits_out[NUM_SUITS]
+)
 {
+    if (cards == NULL || ranks_out == NULL || suits_out == NULL)
+    {
+        return;
+    }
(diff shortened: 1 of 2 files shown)


MeirGavish:
> The change aligns the behavior of `get_played_distribution()` with the existing `get_hand_distribution()` function, ensuring that only scored cards in a played hand are included in the distribution calculation.

This makes me more certain of this fix, these functions are basically the same, they just check different arrays, I think these should be consolidated into a single function and the checked array should be passed...

MathisMartin31:
Oh wow so the fix was a single line, I didn't expect that xD

I'll double check but I think you can go on with this PR since it looks like the bug is fixed, and this is code I didn't touch in my own optimization PR, I only changed `get_hand_type()`. Even if you delete one of the two mentioned functions I'll be easy for me so don't worry about that

source/hand_analysis.c
```diff
@@ -33,7 +33,7 @@ void get_played_distribution(u8* ranks_out, u8* suits_out)
     int top = get_played_top();
     for (int i = 0; i <= top; i++)
     {
-        if (!played[i])
+        if (!played[i] || !card_object_is_selected(played[i]))
```
MathisMartin31: There aren't many changes to review here, but as you said yourself, the `get_played/hand_distribution` are now basically the same so they are either duplicates, or they were used interchangeably before when they were not.

I'll have to look deeper into it when I get my computer back but since you were on tracks you can go ahead and do it too, I'll look into the changes you come up with :)

MeirGavish:
Closing. See [comment on corresponding issue](https://github.com/GBALATRO/balatro-gba/issues/341#issuecomment-3691363488) and further discussion there.

Issue #344, opened 2025-12-26
Change toggle sort key to SELECT


MeirGavish:
I think the SELECT button makes more sense to toggle the sorting than the R button. In Balatro it's mapped to "Run Info", not sure we would implement that and even if we did, we could make it an on-screen button, or even the sort toggle on-screen buttons like Balatro does.

[Related discussion](https://github.com/GBALATRO/balatro-gba/issues/339#issuecomment-3692607037)

MeirGavish:
See this [comment](https://github.com/GBALATRO/balatro-gba/issues/339#issuecomment-3692659810). Letting this cook for now to get input from others, might forego this change and close issue and PR.

MeirGavish:
Also if #346 is done, we can forego this and remove the toggle sorting key altogether.

MeirGavish:
Closing this as of PR #351, it doesn't really make sense.

Issue #339, opened 2025-12-22
Improve card selection UX


MeirGavish:
When adding card swapping in PR #280, effort was made to maintain the flow of easily selecting cards but it still feels a little off. I think we need to re-look at this and try to improve it to be closer to how it was before the change because now selecting cards can become a little frustrating and the game feels less enjoyable now.
I think this should be done after/during the refactor to use the selection grid in #320.

ricfehr3:
I have been messing with the `card_swap_time_threshold` variable, it's helpful but still feels a bit off. I'm not sure what would make it a better experience... I think what it may need is better feedback that you're in "`move`" mode?  Maybe a loading bar with a sprite? Or making the card a different color or bigger? 

If we can solve it without that, it would be good, those are some large additions. However, if it's a good path, then maybe we should try some simple indicator?

The idea I have in my mind is a small sprite of a couple of a horizontal red line. When holding A for more than an instant the sprite would be scaled according to the timer with `card_swap_time_threshold`. Then it could either disappear or stay above the card.

Something simpler would totally be fine too.

MeirGavish:
Here's how Balatro does that when you play with a controller, it's similar to how we do it but with a few differences.

1. It gives feedback when you're in move mode by placing the card sprite you're moving above the others. We can do that too, make it the top sprite layer in the z-order.
2. If you're in move mode, releasing the selection button does not select the card anymore. That is not how we do this and maybe doing that can help us but I'm not sure.
3. The indicator for the card focus is a little different, we raise it, Balatro flashes it white.

It seems like pressing the selection button immediately raises the card to the top layer even if it's selected and not moved and move mode still has a threshold before it's entered.

https://github.com/user-attachments/assets/43422047-0ce1-4b00-9a81-8d8c0da82369

MathisMartin31:
Yes I played some rounds of Balatro to try and figure out how it worked. The whole "selected instead of moved" when pressing left/right is also there. I still couldn't figure out how to make it smooth, but when playing reaaaally quickly Balatro doesn't feel *that* smooth either imo, so any improvement on our end would actually be a huge thing.

As for visual flair, we could flash the card white by displaying a texture on top of it with an animation, so that's not really an issue, but I think that to keep track of what card is currently hovered we still need to raise it above the others because we can't really outline it or anything since we don't have that much room

Mega-Evolved-Ziquinho:
Just so you know @MathisMartin31, you sent the message twice

[...]

PR #345, opened 2025-12-26
Changed toggle sort key to SELECT from R

MeirGavish:
Tested on real hardware and I think it feels good enough. For issue #344.


MeirGavish:
Converted to draft for now since I'm not sure if we want to do this, see [this comment](https://github.com/GBALATRO/balatro-gba/issues/339#issuecomment-3692659810)

MeirGavish:
Closing this as of PR #351, it doesn't really make sense.

PR #348, opened 2025-12-27
Refactored to use selection grid during round

MeirGavish:
I refactored the round to use the selection grid now.
I also added button.c/.h to generalize button logic, especially highlighting as well as pressing.
This refactor includes moves and renames of code and it seems like the differe is not handling it well so I will try to comment what happened to some of the functions.
I wanted to remove the globals `selection_x` and `selection_y` but they're still used by the main menu and blind select so next time.
The global variables `moving_card`, `card_moved_too_fast` etc. couldn't have been moved to a single function and made local since they are used by both the selection callback and movement callback.

Also, I think this might solve #339 too. The selection feels better to me now, I think that maybe handling directional and selection input separately is what solved it. I need to test it again some more and compare to main but I would like input from others too.

Edit: Okay, after testing and comparing to main, it really feels like sometimes selection is just not registered in main but this PR fixes it so I'm linking #339 too.

https://github.com/user-attachments/assets/7e8b3ec3-8443-4f62-886d-bf6e54809e7b

https://github.com/user-attachments/assets/0a66c3a5-8bcc-45f8-a542-b3887117ca07

https://github.com/user-attachments/assets/3b4869e9-c305-4a78-8af7-4a5b3e2687a7

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -2143,7 +2143,7 @@ static bool game_playing_button_row_on_selection_changed(
 {
     // The selection grid system only guarantees that the new selection is within bounds
     // but not the previous one...
-    // As of writing (PR #348), this check is not strictly needed for this row but it is 
+    // As of writing (PR #348), this check is not strictly needed for this row but it is
     // left in, in case that ever changes. It can be reconsidered and removed.
     if (prev_selection->y == row_idx && prev_selection->x >= 0 &&
         prev_selection->x < game_playing_button_row_get_size())


include/selection_grid.h
```diff
@@ -37,9 +37,10 @@ typedef struct SelectionGrid SelectionGrid;
  *                       Contains the previous cursor position
  * @param new_selection Pointer to the Selection state after the change occurred.
  *                      Contains the new cursor position
+ * @return false if the selection change needs to be aborted, true if can proceed.
  *
  */
-typedef void (*RowOnSelectionChangedFunc)(
+typedef bool (*RowOnSelectionChangedFunc)(
```
MeirGavish, line 43: I had to make some changes to the selection grid to be able to support this. This return value was specifically added to avoid wrapping while swapping cards.

include/selection_grid.h
```diff
@@ -71,6 +72,18 @@ typedef int (*RowGetSizeFunc)();
...
+ * @brief A set of attributes to the selection grid row affecting selection grid behavior.
+ */
+typedef struct
+{
+
+    /**
+     * @brief Whether to wrap selection when it passes the end of the row.
+     */
+    bool wrap;
+} SelGridRowAttributes;
```
MeirGavish, line 85: I decided to add it as a struct containing all the special attributes/options of the row in case we want to add anything else.

source/game.c
```diff
@@ -128,12 +126,11 @@
...
-#define GAME_PLAYING_BUTTONS_SEL_Y     1
-#define GAME_PLAYING_PLAY_BTN_SEL_X    0
-#define GAME_PLAYING_DISCARD_BTN_SEL_X 1
-#define GAME_PLAYING_NUM_SEL_ROWS      2
-#define GAME_PLAYING_NUM_BOTTOM_BTNS   2
+#define GAME_PLAYING_HAND_SEL_Y    1
+#define GAME_PLAYING_BUTTONS_SEL_Y 2
+
+#define GAME_PLAYING_NUM_SEL_ROWS    2
+#define GAME_PLAYING_NUM_BOTTOM_BTNS 2
```
```suggestion
#define GAME_PLAYING_HAND_SEL_Y      1
#define GAME_PLAYING_BUTTONS_SEL_Y   2
#define GAME_PLAYING_NUM_SEL_ROWS    2
#define GAME_PLAYING_NUM_BOTTOM_BTNS 2
```

Maybe remove the space between the defines since they're related.

source/game.c
```diff
@@ -399,10 +429,31 @@ static StateInfo state_info[] = {
 #undef DEF_STATE_INFO
 };
 
+// TODO: Format for line length
+SelectionGridRow game_playing_selection_rows[] = {
+    {0, jokers_sel_row_get_size,          jokers_sel_row_on_selection_changed,          jokers_sel_row_on_key_transit,        {.wrap = false}},
+    {1, game_playing_hand_row_get_size,   game_playing_hand_row_on_selection_changed,   game_playing_hand_row_on_key_transit, {.wrap = true} },
+    {2, game_playing_button_row_get_size, game_playing_button_row_on_selection_changed, game_playing_button_row_on_key_hit,   {.wrap = true} }
+};
```
MeirGavish: I think the lines have gotten a little too long and I would have liked to break this down but right now this is the only format clang-format accepts.

source/game.c
```diff
@@ -239,8 +236,6 @@ static void display_round(int value);
 static void display_hands(int value);
 static void display_discards(int value);
 static void set_hand(void);
-static void hand_set_focus(int index);
-static bool hand_can_discard(void);
```
MeirGavish, line 243: This was a duplicate forward declaration

source/game.c
```diff
@@ -1728,37 +1791,9 @@ static void set_hand(void)
...
-    {
-        selection_x = index;
-    }
-
-    play_sfx(
-        SFX_CARD_FOCUS,
-        MM_BASE_PITCH_RATE + rand() % CARD_FOCUS_SFX_PITCH_OFFSET_RANGE,
-        SFX_DEFAULT_VOLUME
-    );
-}
```
MeirGavish, line 1755: This function was completely removed, not moved

source/game.c
```diff
@@ -1905,61 +1946,42 @@ static inline void set_seed(int seed)
...
-        hand_selections--;
-        play_sfx(SFX_CARD_DESELECT, MM_BASE_PITCH_RATE, SFX_DEFAULT_VOLUME);
-    }
-    else if (hand_selections < MAX_SELECTION_SIZE)
-    {
-        card_object_set_selected(hand[selection_x], true);
-        hand_selections++;
-        play_sfx(SFX_CARD_SELECT, MM_BASE_PITCH_RATE, SFX_DEFAULT_VOLUME);
-    }
-}
```
MeirGavish, line 1925: This function was turned into `hand_select_card()`

source/game.c
```diff
@@ -1905,61 +1946,42 @@ static inline void set_seed(int seed)
...
+        DISCARDS_TEXT_RECT.left,
+        DISCARDS_TEXT_RECT.top,
+        TTE_RED_PB,
+        discards
+    );
 
-    if (any_cards_deselected)
-    {
-        play_sfx(SFX_CARD_DESELECT, MM_BASE_PITCH_RATE, SFX_DEFAULT_VOLUME);
-    }
```
MeirGavish, line 1943: Just moved

source/game.c
```diff
@@ -3090,8 +3139,53 @@ static inline void game_playing_discarded_cards_loop(void)
...
+            /* FALL THROUGH */
+        case FIVE_OF_A_KIND:
+            /* FALL THROUGH */
+        case FLUSH_HOUSE:
+            /* FALL THROUGH */
+        case FLUSH_FIVE: // Select all played cards in the hand
+            select_all_five_cards_in_played_hand();
+            break;
+    }
+}
```
MeirGavish, line 3201: This is an extra function extraction refactor I did, not directly related to this PR.

source/game.c
```diff
@@ -3108,7 +3202,9 @@ static inline void cards_in_hand_update_loop(void)
                         hand_x + (int2fx(i) - int2fx(hand_top) / 2) * -HAND_SPACING_LUT[hand_top];
                     break;
                 case HAND_SELECT:
-                    bool is_focused = (i == selection_x && selection_y == 0);
+                    bool is_focused =
+                        (i == selected_card_idx &&
+                         game_playing_selection_grid.selection.y == GAME_PLAYING_HAND_SEL_Y);
```
MeirGavish, line 3225: So card focus (highlight) is managed by this function `cards_in_hand_update_loop()`,previously it used `selection_x` and `selection_y` so I just let it directly check the selection grid selection that now replaces it.

source/game.c
```diff
@@ -1905,61 +1946,42 @@ static inline void set_seed(int seed)
...
-{
-    if (hand_state != HAND_SELECT || hand_selections == 0)
-        return false;
-    return true;
+    hand_state = HAND_PLAY;
+    display_hands(--hands);
+
+    // Move back to hand selection
+    selection_grid_move_selection_vert(&game_playing_selection_grid, -1);
 }
```
MeirGavish, line 1997: These were just moved

source/game.c
```diff
@@ -1978,185 +2000,223 @@ static bool card_selected_instead_of_moved = false;
...
+            play_sfx(
+                SFX_CARD_FOCUS,
+                MM_BASE_PITCH_RATE + rand() % CARD_FOCUS_SFX_PITCH_OFFSET_RANGE,
+                SFX_DEFAULT_VOLUME
+            );
+        }
     }
-}
 
-static inline void game_playing_highlight_play_btn(void)
```
MeirGavish, line 2016: The highlight functions were removed as made redundant by button.c/.h

source/game.c
```diff
@@ -1978,185 +2000,223 @@ static bool card_selected_instead_of_moved = false;
...
 
-static inline void game_playing_unhighlight_buttons(void)
+static int game_playing_button_row_get_size(void)
 {
-    memcpy16(&pal_bg_mem[PLAY_HAND_BTN_BORDER_PID], &pal_bg_mem[PLAY_HAND_BTN_PID], 1);
-    memcpy16(&pal_bg_mem[DISCARD_BTN_BORDER_PID], &pal_bg_mem[DISCARD_BTN_PID], 1);
+    return NUM_ELEM_IN_ARR(game_playing_buttons);
 }
 
-static inline void game_playing_execute_hand_discard(void)
```
MeirGavish, line 2034: Turned into `game_playing_discard_on_pressed()`

source/game.c
```diff
@@ -1978,185 +2000,223 @@ static bool card_selected_instead_of_moved = false;
...
-    display_hands(--hands);
+    if (new_selection->y == row_idx)
+    {
+        game_playing_button_set_highlight(new_selection->x, true);
+    }
+
+    return true;
 }
 
-static inline void game_playing_process_hand_select_input(void)
```
MeirGavish, line 2062: This function was gutted entirely, it exists but it's only a call to `selection_grid_process_input()`

source/game.c
```diff
@@ -2184,17 +2244,6 @@ static inline void card_draw(void)
...
-static inline void display_ante(int value)
-{
-    tte_printf(
-        "#{P:%d,%d; cx:0xC000}%d#{cx:0xF000}/%d",
-        ANTE_TEXT_RECT.left,
-        ANTE_TEXT_RECT.top,
-        value,
-        MAX_ANTE
-    );
-}
```
MeirGavish, line 2196: Moved

source/game.c
```diff
@@ -128,12 +126,10 @@
...
 
-#define GAME_PLAYING_HAND_SEL_Y        0
-#define GAME_PLAYING_BUTTONS_SEL_Y     1
-#define GAME_PLAYING_PLAY_BTN_SEL_X    0
-#define GAME_PLAYING_DISCARD_BTN_SEL_X 1
-#define GAME_PLAYING_NUM_SEL_ROWS      2
-#define GAME_PLAYING_NUM_BOTTOM_BTNS   2
+#define GAME_PLAYING_HAND_SEL_Y      1
+#define GAME_PLAYING_BUTTONS_SEL_Y   2
+#define GAME_PLAYING_NUM_SEL_ROWS    2
```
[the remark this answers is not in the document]
MeirGavish: This is a leftover, will be removed

source/game.c
```diff
@@ -3090,8 +3138,53 @@ static inline void game_playing_discarded_cards_loop(void)
...
+        case STRAIGHT:
+            /* FALL THROUGH */
+        case FLUSH:
+            /* FALL THROUGH */
+        case STRAIGHT_FLUSH:
+            /* FALL THROUGH */
+        case ROYAL_FLUSH:
+            select_flush_and_straight_cards_in_played_hand();
+            break;
+            // ELSE FALL THROUGH
```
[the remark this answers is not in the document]
MeirGavish: Alright.

source/game.c
```diff
@@ -3090,8 +3138,53 @@ static inline void game_playing_discarded_cards_loop(void)
...
+        case FLUSH_FIVE: // Select all played cards in the hand
+            select_all_five_cards_in_played_hand();
+            break;
+    }
+}
+
 static inline void cards_in_hand_update_loop(void)
 {
+    int selected_card_idx = hand_sel_idx_to_card_idx(game_playing_selection_grid.selection.x);
+
```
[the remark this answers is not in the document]
MeirGavish: Since the iteration goes over the hand, the value should only be used when the hand has valid cards. 
I'm more concerned about the iteration starting from `hand_top + 1` and not from `hand_top` but not sure I should touch it in this PR...

source/game.c
```diff
@@ -1978,185 +1999,223 @@ static bool card_selected_instead_of_moved = false;
...
+{
+    // This is because the hand is drawn from right to left.
+    // There is no particular reason for why that was done, it's just how it was done.
+    // Maybe one day it can be reverted and made consistent so this conversion is not needed.
+    return hand_get_size() - selection_index - 1;
+}
+
+static void hand_select_card(int index)
+{
+    if (index < 0 || index > hand_get_size() || hand_state != HAND_SELECT || hand[index] == NULL)
```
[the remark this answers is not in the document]
MeirGavish: Yep.

source/selection_grid.c
```diff
@@ -28,17 +28,29 @@ void selection_grid_move_selection_horz(SelectionGrid* selection_grid, int direc
...
+    if (selection_grid->rows[new_selection.y].attributes.wrap)
+    {
+        int row_size = selection_grid->rows[new_selection.y].get_size();
+
+        new_selection.x = wrap(new_selection.x, 0, row_size);
+    }
+
     if (selection_grid->selection.y >= 0 &&
         selection_grid->selection.y < selection_grid->num_rows && new_selection.x >= 0 &&
         new_selection.x < selection_grid->rows[new_selection.y].get_size())
```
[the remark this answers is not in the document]
MeirGavish: Setting `wrap_enabled` to true/false is pretty redundant, it can be just set to the value of `attributes.wrap`.
Either way, the check would just be redundant in this case, I'm more concerned by accessing rows[new_selection.y] before bounds checking, I will fix that.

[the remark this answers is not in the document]
MeirGavish: Done.

MathisMartin31:
I tried selecting Jokers like shown in your videos but I couldn't, did you disable it temporarily, or did I miss something? Looks very nice and useful otherwise ^^

MeirGavish:
> I tried selecting Jokers like shown in your videos but I couldn't, did you disable it temporarily, or did I miss something? Looks very nice and useful otherwise ^^

@MathisMartin31 Thanks! No, I didn't disable it, I thought I may have made a mistake in one of my recent commits but I checked and it seems to be working. Are you sure you compiled the PR branch with the latest commit?
Also if/when you get a chance to check, can you say if the hand selection feels better than it is currently in the main branch? I'm trying to think whether this solves #339.
I also want to request a review from you just to make sure my changes to the card swapping code seem okay but I don't want to add so much while you're on holiday.

MathisMartin31:
@MeirGavish Oooooh you're right, I went and clicked on the link at the top of this page, and it took me to the right repo but when copying the url to clone with git, it doesn't add the branch you're on so I compiled main by mistake ><

I compiled the right branch this time and everything works like a charm! The Joker swapping during the round is very nice, and card selection is buttery smooth, way better than before I feel like it's even smoother than the base Balatro game itself. I didn't miss a single card when selecting quickly and the swapping is very responsive.

Don't worry about me being on holiday, I'll take a look when I can but whatever changes you made, the result is very good so my guess is your code can't be bad :laughing: 

Nice work! ^^

include/button.h
```diff
@@ -0,0 +1,70 @@
+#ifndef BUTTON_H
```
ricfehr3, line 7: This file needs a doxygen header.

include/selection_grid.h
ricfehr3: 💯 I agree with this.

source/button.c
```diff
@@ -0,0 +1,34 @@
...
+{
+    if (button == NULL)
+        return;
+
+    u16 set_color = BTN_HIGHLIGHT_COLOR;
+
+    if (!highlight)
+    {
+        set_color = pal_bg_mem[button->button_pal_idx];
+    }
```
ricfehr3:
```suggestion
    u16 set_color = highlight ? BTN_HIGHLIGHT_COLOR : pal_bg_mem[button->button_pal_idx];
```

Kind of a nitpick (there may be some performance improvements, but probably not much), but I think a ternary operator would work better here.

source/button.c
```diff
@@ -0,0 +1,34 @@
...
+
+    memset16(&pal_bg_mem[button->border_pal_idx], set_color, 1);
+}
+
+void button_press(Button* button)
+{
+    if (button == NULL || button->on_pressed == NULL ||
+        (button->can_be_pressed != NULL && !button->can_be_pressed()))
```
ricfehr3, line 21: if `button` is `NULL` here, the call to `can_be_pressed()`  is undefined behavior. I think the other checks are fine, but that call will need to happen after the initial checks. 

On that same note. It would also be good to check if the function pointer is NULL or not before calling.

source/selection_grid.c
```diff
@@ -66,14 +79,29 @@ void selection_grid_move_selection_vert(SelectionGrid* selection_grid, int direc
...
+        {
+            proceed_selection = selection_grid->rows[new_selection.y].on_selection_changed(
+                selection_grid,
+                new_selection.y,
+                &selection,
+                &new_selection
+            );
+        }
+
+        if (proceed_selection)
```
ricfehr3, line 101: I would move this into the above `if` statement. I think it would make more sense logically.

source/game.c
ricfehr3: Feel free to turn clang-format off if you want to make it more to your taste.

source/game.c
```diff
@@ -1978,185 +1998,223 @@ static bool card_selected_instead_of_moved = false;
...
+static bool game_playing_button_row_on_selection_changed(
+    SelectionGrid* selection_grid,
+    int row_idx,
+    const Selection* prev_selection,
+    const Selection* new_selection
+)
 {
-    play_sfx(SFX_BUTTON, MM_BASE_PITCH_RATE, BUTTON_SFX_VOLUME);
+    // The selection grid system only guarantees that the new selection is within bounds
+    // but not the previous one...
```
ricfehr3, line 2145: Is there any reason to add a check for the previous? If not, totally chill.

source/game.c
ricfehr3: I'm a huge fan of this, nice and compartmentalized :)

source/button.c
MeirGavish: Not sure why I didn't use the ternary operator here.

source/button.c
MeirGavish: This is a short-circuit evaluation, if `button` is `NULL`, the condition is `true` and the rest of the `||` is not evaluated. The same works for the `&&`, if the function is `NULL`, the condition is `false` and the call is not evaluated. The order matters.

source/game.c
MeirGavish: Actually I was basing this off of the joker row and there's the special case there where you sell a joker and then the selection moves, I this could cause the previous selection to be outside of bounds in case there was only one joker. For buttons I think there isn't much need for the check here since they are more static. I'll just remove the check and add a comment for it.

Edit: actually looking again, I see this is only in the shop row, I think the case was buying a joker then... I need to give it another look...

source/selection_grid.c
MeirGavish: The idea here was to allow the second function to still abort the selection so I need to check its return value again and only update according to that. I can put this in the previous `if` but then it increases nesting.

source/game.c
MeirGavish: I did that, I also see now that I left wrapping on for the buttons and I shouldn't have.

source/button.c
ricfehr3: Oh righteous! Thanks for teaching me something new :)

source/selection_grid.c
ricfehr3: Keeping the nesting down is totally fair. Ultimately the logic is the same, and it seems like a reasonable optimization.

source/game.c
MeirGavish: @ricfehr3 Okay, I gave this another look and figured out that I did this in the shop because I used a hack to highlight the initial selection, I added another commit with a comment to clarify it. The case for the joker sold is not relevant because there I call the move before the sell so everything is valid.

So, technically the check here is not needed and I'm considering removing it but I'm hesitating because I worry it could become relevant for some reason.

ricfehr3: @MeirGavish I think this is the only outstanding item for this PR. I'm more than happy to just merge it as is and leave a comment or make an issue to revisit the usefulness of the check. I think it would make sense to defer to later. The rest of this PR is gold.

source/game.c
```diff
@@ -1978,185 +2017,223 @@ static bool card_selected_instead_of_moved = false;
...
```
MeirGavish, line 2145:
```suggestion
    // but not the previous one...
    // As of writing (PR #348), this check is not strictly needed for this row but it is 
    // left in, in case that ever changes. It can be reconsidered and removed.
```

source/game.c
MeirGavish: @ricfehr3 I added a clarification comment.

MeirGavish:
Merging this, I will start testing before release.

Issue #170, opened 2025-10-14
Support multiple boss graphics for blinds


ricfehr3:
Following up with #133 . The current **blind** system only supports one type of boss. To add more bosses, each boss will need to load it's graphic data (tiles and palette) into the boss blind tid/pid locations. There is support for loading boss graphics data as of #133 . However, there is an issue that the blinds are hardcoded as indices into the palette array, to map background and highlight colors based on the blind. This currently makes it so grit is not applicable, as grit does not guarantee the order of the palette will be consistent to this mapping. What needs to be done here is decoupling the hardcoded indices in the palette array with some other mapping or mechanism. 

I have no idea how to do this yet. However, the only way may be to store the highlight/background colors in a struct that gets loaded per boss. Unfortunately that requires manual work, so maybe there's a better option.

MeirGavish:
Maybe there's a different tool we can use supplement/replace grit that would allow us to control the palette mapping.

ricfehr3:
I think we should revisit and think about this earlier issue on VRAM: #120 . I like the suggestion of marking cards by RGB colors to override that color. It gives something to search for at the very least, possibly able to be used for this issue as well.

I'm not sure how capable grit would be for these particular graphics. I'm thinking a new tool is needed. I'll come back to this when I can, but if anyone can chime in with ideas, it's welcome.

MeirGavish:
From [the regular backgrounds chapter](https://gbadev.net/tonc/regbg.html) in the tonc tutorial book:
_There are a few map editors out there that you can use. Two good ones are Nessie’s [MapEd](https://nessie.gbadev.org/) or [Mappy](https://www.tilemap.co.uk/mappy.php), both of which have a number of interesting features. I have my own map editor, [mirach](https://www.coranac.com/projects/#mirach), but it’s just a very basic thing._

I haven't looked at any of these tools, I don't know if they can do what we need here, and they may be a little outdated considering the tutorial was written over 20 years ago but it might be worth a look.

Also in the [tonc keypad demo](https://github.com/gbadev-org/libtonc-examples/tree/master/basic/key_demo) all the button highlight colors are mapped nicely sequentially corresponding to the key index and I'm not sure how he did that.

ricfehr3:
Noting: Related to #192 , related work is started here and it's a good reference point

MathisMartin31:
[...]

PR #285, opened 2025-12-06
Implement arbitrary-sized joker spritesheets

MathisMartin31:
From our discussion in #274 

I took the liberty to add the art for the Shoot the Moon Joker, since the new sprite art for the Face cards has been completed and it's basically just the Queen of Spades. Plus, it fits inside the new sprite sheet 0 ^^

I have a couple of big maps where I turned clang formatting off because it was really messing with the alignment. Also there are a loooot of changed lines in this PR because I formatted the files I changed so it added up pretty quickly ^^'

I'm putting it as a draft for now, see if anybody can squeeze more Jokers together.

PR #350, opened 2025-12-29
Encoded palette indexing in background pngs

MeirGavish:
So apparently image file formats [including PNG](https://en.wikipedia.org/wiki/Indexed_color#Image_file_formats_supporting_indexed_color) actually support palette indexing.
Then grit can use the palette index from the file instead of generating a new one every time.
What this means is that we can edit images while preserving the palette map and don't have to change the indices in the code.

What I did in this PR was open all the background images in usenti which creates the palette index the same way as grit and then used Image->Export... and chose png to overwrite the existing files. Now the same palette that was created with grit is encoded within the pngs.
When opening the files with LibreSprite you can see the same palette loaded and if you check Sprite->Color Mode you can see that it's "Indexed". Adding new colors means you have to manually update the palette which is a bit more cumbersome but it preserves the same mapping. 
To do this in LibreSprite I load the image with the color I want, sample the color, copy the RGB code, go back to the image I want to modify, edit color, paste the code, click "Add foreground color to palette", and rearrange the color in the palette to where I want it and within the 256 limit.

This should also allow us to modify the palette and control the color order which can be very useful for #170, I haven't touched any of the blind sprites to not interfere with any work @MathisMartin31 has done but I think this should be the way to go there and we should do away with all the hardcoded colors.
I also think this can be useful for the playing cards to implement high contrast. I don't know about the jokers, it could be nice for negatives but maybe this would make making joker art too cumbersome and I'm not touching them or any of the other sprites in this PR.
This is just for the backgrounds to make sure that any edit to them will not mess up the palette mapping and require changing the indices in the code.

@MathisMartin31 I'm requesting a review since you use GIMP so you can check how this works there, I'm not familiar with it at all. Either way, no rush with this, do it once you return from vacation.

Screenshots to show everything is still working.
balatro-gba-0
balatro-gba-1
balatro-gba-2
balatro-gba-3


GreenNinja2525:
This could be interesting. It really opens the door to more possibilities if you think about it. Like you said, high contrast cards could be feasible. It really sounds interesting, great work @MeirGavish

MathisMartin31:
Hi @MeirGavish, very useful stuff you have here!

Unfortunately for me, GIMP does not give me any message about importing/using the palette from the file and opens it in RGB mode (as opposed to indexed with a limited palette), and re-generating the palettes does not produce the same result, breaking our color swapping.

It's not that big of an issue though, most of my artwork is done anyway, and if I ever have to do more, I can do it in GIMP and copy-paste it in Usenti (which I'll have to download). Simplifying the palette process is more important than preserving 100% of my workflow.

As such, I will approve this PR since it's a very important QoL update for everyone ^^

MathisMartin31 (approved):

MeirGavish:
> Hi @MeirGavish, very useful stuff you have here!
> 
> Unfortunately for me, GIMP does not give me any message about importing/using the palette from the file and opens it in RGB mode (as opposed to indexed with a limited palette), and re-generating the palettes does not produce the same result, breaking our color swapping.
> 
> It's not that big of an issue though, most of my artwork is done anyway, and if I ever have to do more, I can do it in GIMP and copy-paste it in Usenti (which I'll have to download). Simplifying the palette process is more important than preserving 100% of my workflow.
> 
> As such, I will approve this PR since it's a very important QoL update for everyone ^^

Thanks! It is a shame this doesn't work in GIMP. I really recommend LibreSprite which is free and open source (and the also open source but the paid version Aseprite), I've been using it a lot for this project and it's a really great pixel art program. We'll find a workflow that works for you either way. For now we will not do this for jokers.

MathisMartin31:
> We'll find a workflow that works for you either way. For now we will not do this for jokers.

I think we should do it once #285 is merged since, as shown in the "changed files tab", indexing the colors seems to be reducing the file size by around 20% to 50% depending on the image.

This is very much not negligible when it comes to ROM size. We are miles away from the 32MB limit still but, just like with palettes, any easily achievable optimization can't be slept on.

MeirGavish:
> > We'll find a workflow that works for you either way. For now we will not do this for jokers.
> 
> I think we should do it once #285 is merged since, as shown in the "changed files tab", indexing the colors seems to be reducing the file size by around 20% to 50% depending on the image.
> 
> This is very much not negligible when it comes to ROM size. We are miles away from the 32MB limit still but, just like with palettes, any easily achievable optimization can't be slept on.

I don't think the file size has an effect on the ROM size because grit compiles it into the same data.

MathisMartin31:
> I don't think the file size has an effect on the ROM size because grit compiles it into the same data.

Oh you're right, I compared the ROM sizes with and without your changes, it's indeed the same. The more you know ^^

MeirGavish:
> > I don't think the file size has an effect on the ROM size because grit compiles it into the same data.
> 
> Oh you're right, I compared the ROM sizes with and without your changes, it's indeed the same. The more you know ^^

Yeah, most of the ROM is music anyway.

ricfehr3 (approved):
Nice. Going to merge this one!

Issue #354, opened 2026-01-03
Implement Boss Blind effects


MathisMartin31:
Related to #352. Since the Boss Blind graphics are well on the way, I think it's time to think about how to deal with their effects.

What I had in mind was something similar (if not entirely identical) to the Joker effects, with a single callback, event types and checks in game.c at different steps of the round.

Here's a list of all Blind effect event types just below. This will give us an idea of what we're dealing with but there clearly aren't as many as for the Jokers:

- No Effect : Wall, Violet Vessel
- On Round Start : Club, Fish, Goad, Water, Window, Manacle, Plant, Pillar, Needle, Head, Amber Acorn, Verdant Leaf, Crimson Heart, Cerulean Bell
- On Hand Played : Hook, Ox, Arm, Psychic, Eye, Mouth, Tooth, Flint
- On Hand Played End : Crimson Heart, Cerulean Bell
- On Cards Drawn : House, Wheel, Serpent, Mark
- Very Specific : Verdant Leaf (on card sold)

Most of these could correspond to a Joker Event, except Verdant Leaf but that's because we haven't implement the Campfire Joker yet, which also triggers when selling a card.

So we could either create a new, limited enum for the Blind effects, or rename the `JokerEvent` one to just `Event` or something and use it, expanding it if needed.

MeirGavish:
As I said, adding Boss Blind effects without Luchador, Director's Cut, or Chicot would make the game much more difficult. I wonder if it would make it near unbeatable since it seems like most builds that win the game now rely on face cards so The Plant would be a vicious inescapable run killer. I suppose we can add Luchador since it's a (non-legendary) joker and we already have jokers.
Other than that, there's the question of how much of the game we want to implement to not compete with the original Balatro. I think this would be a good time to start a discussion about the general scope of the project and make a decision about that.

Implementation-wise, we need to make sure boss blind effects can be disabled too, I think the logic for that should be done in the blind code and not at the game/round code so it might be a wrapper for the callback to check if the blind is disabled. But that also means the "No Effect" blinds such as The Wall and Violet Vessel which would need to be able to reset to the default boss blind score requirement.

As for how to implement the event system, I want to think about it a little more, it makes sense to have a similar system to the joker one and probably use the same event enum since they do correspond to the same events, but maybe it makes sense to have a different function for each type of event this time so we can pass the proper relevant arguments to it.

There are also more things that would need to be implemented for some blinds such as face down cards and debuffed cards.

[...]

PR #357, opened 2026-01-07
Add Luchador Joker to disable Boss Blind Effects

MathisMartin31:
Following our discussion in #354, I added Luchador which activates only when selling the Joker while playing a Boss Blind.

Since we can't currently sell a Joker while playing a round, this is blocked by #348 and should ideally be done after #285 because it would be easier to merge this way rather than rebasing with a lot of changes.

For now it doesn't actually do anything since Boss Blind effects aren't a thing, but the logic is there


MeirGavish:
I think we should add boss blind effects before or with Luchador, otherwise it's just adding a joker that doesn't do anything but can appear in the shop. I guess we can keep him disabled for now.

MathisMartin31:
> I think we should add boss blind effects before or with Luchador, otherwise it's just adding a joker that doesn't do anything but can appear in the shop. I guess we can keep him disabled for now.

Yes that's why I put this up as a draft, but maybe it was wayyy too soon ><. I'll cancel this PR, use this branch for the blind effects and come back later with a proper PR

PR #359, opened 2026-01-09
Fix Four Fingers joker effect persisting after game over

STADD:
When a run ended with Four Fingers owned, starting a new game would keep the straight/flush size at 4 instead of resetting to the default value of 5. This caused the Four Fingers effect to remain active even without the joker card.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -564,7 +564,6 @@ static int discard_top = -1;
 static int shortcut_joker_count = 0;
 
 static int four_fingers_joker_count = 0;
-static int straight_and_flush_size = STRAIGHT_AND_FLUSH_SIZE_DEFAULT;
 
 GBAL_UNUSED
 static inline bool is_shop_joker_avail(int joker_id)
@@ -879,7 +878,8 @@ bool is_shortcut_joker_active(void)
 
 int get_straight_and_flush_size(void)
 {
-    return straight_and_flush_size;
+    return four_fingers_joker_count > 0 ? STRAIGHT_AND_FLUSH_SIZE_FOUR_FINGERS
+                                        : STRAIGHT_AND_FLUSH_SIZE_DEFAULT;
 }
 
 static void add_joker(JokerObject* joker_object)
@@ -891,10 +891,6 @@ static void add_joker(JokerObject* joker_object)
     // only change size when the first one is added
     if (joker_object->joker->id == FOUR_FINGERS_JOKER_ID)
     {
-        if (four_fingers_joker_count == 0)
-        {
-            straight_and_flush_size = STRAIGHT_AND_FLUSH_SIZE_FOUR_FINGERS;
-        }
         four_fingers_joker_count++;
     }
 
@@ -913,10 +909,6 @@ static void remove_owned_joker(int owned_joker_idx)
     if (joker_object->joker->id == FOUR_FINGERS_JOKER_ID)
     {
         four_fingers_joker_count--;
-        if (four_fingers_joker_count == 0)
-        {
-            straight_and_flush_size = STRAIGHT_AND_FLUSH_SIZE_DEFAULT;
-        }
     }
 
     if (joker_object->joker->id == SHORTCUT_JOKER_ID)
@@ -4791,8 +4783,8 @@ static void game_over_on_exit()
     while (list_get_len(&_owned_jokers_list) > 0)
     {
         JokerObject* joker_object = list_get_at_idx(&_owned_jokers_list, 0);
-        joker_object_destroy(&joker_object);
         remove_owned_joker(0);
+        joker_object_destroy(&joker_object);
     }
 
     tte_erase_screen();


MathisMartin31:
Oooh that's a good catch, nice ^^

@MeirGavish I just tested our main and I can confirm the bug is there, and that STADD's PR fixes it. This looks like a very straightforward bug fix, maybe this can be moved to the Stable Release milestone?

MathisMartin31:
> The thing is, that this is not the correct fix. Even with it, there is still a bug that if you restart the game and get Four Fingers again, it then doesn't have any effect and the Flush and Straight size is 5.
> 
> The issue is that the code destroys the joker objects but doesn't call the removal function `remove_owned_joker()` that would update the relevant data. This also means there is an issue with the shop and previously owned jokers may not appear in the run after restart.

Wooops, sorry I didn't test it that far ><

STADD:
@MathisMartin31 and @MeirGavish, thank you for taking the time to look into this and explain in more detail why the previous fix wouldn’t fully resolve the issue.

One alternative approach could be something like the following. This ensures that owned jokers are properly removed from the list and that their corresponding JokerObjects are destroyed:

 ```diff
@@ -4788,17 +4788,13 @@ static void game_lose_on_update()
 // util we decide what we want to do after a game over.
 static void game_over_on_exit()
 {
-    ListItr itr = list_itr_create(&_owned_jokers_list);
-    JokerObject* joker_object;
-
-    while ((joker_object = list_itr_next(&itr)))
+    while (list_get_len(&_owned_jokers_list) > 0)
     {
+        JokerObject* joker_object = list_get_at_idx(&_owned_jokers_list, 0);
         joker_object_destroy(&joker_object);
+        remove_owned_joker(0);
     }
 ```

MeirGavish:
> @MathisMartin31 and @MeirGavish, thank you for taking the time to look into this and explain in more detail why the previous fix wouldn’t fully resolve the issue.
> 
> One alternative approach could be something like the following. This ensures that owned jokers are properly removed from the list and that their corresponding JokerObjects are destroyed:
> 
> ```diff
> @@ -4788,17 +4788,13 @@ static void game_lose_on_update()
> // util we decide what we want to do after a game over.
> static void game_over_on_exit()
> {
> -    ListItr itr = list_itr_create(&_owned_jokers_list);
> -    JokerObject* joker_object;
> -
> -    while ((joker_object = list_itr_next(&itr)))
> +    while (list_get_len(&_owned_jokers_list) > 0)
>     {
> +        JokerObject* joker_object = list_get_at_idx(&_owned_jokers_list, 0);
>         joker_object_destroy(&joker_object);
> +        remove_owned_joker(0);
>     }
> ```

Yes, it looks like it would work. Thank you for spotting this bug!

MeirGavish:
@STADD Once it seems fine to you, re-request a review and I will test it.

MeirGavish:
> Wooops, sorry I didn't test it that far ><

And it's okay, I only thought of testing it after giving it a closer look.

MeirGavish:
Looks like it's still bugged.

image

I think what I did was buy Four Fingers, reset, buy Four Fingers again, sell it, and reset again.
I'm not sure why this is happening, I can take another look tomorrow.

MathisMartin31:
@MeirGavish Maybe we could change a bit how Four Fingers work? Instead of updating a variable storing how long straights and flushes are, we could check if Four Fingers is owned when determining the hand type?

Or was this ruled out as too inefficient?

source/game.c
```diff
@@ -4788,12 +4788,11 @@ static void game_lose_on_update()
...
 {
-    ListItr itr = list_itr_create(&_owned_jokers_list);
-    JokerObject* joker_object;
-
-    while ((joker_object = list_itr_next(&itr)))
+    while (list_get_len(&_owned_jokers_list) > 0)
     {
+        JokerObject* joker_object = list_get_at_idx(&_owned_jokers_list, 0);
         joker_object_destroy(&joker_object);
+        remove_owned_joker(0);
```
MeirGavish, line 4786:
```suggestion
        JokerObject* joker_object = list_get_at_idx(&_owned_jokers_list, 0);
        remove_owned_joker(0);
        joker_object_destroy(&joker_object);
```

This is the issue, destroying needs to be done after removal so the data is still valid during removal.

MeirGavish:
> @MeirGavish Maybe we could change a bit how Four Fingers work? Instead of updating a variable storing how long straights and flushes are, we could check if Four Fingers is owned when determining the hand type?
> 
> Or was this ruled out as too inefficient?

That would ensure correctness but it would be a bit of a shame to lose the efficiency there. Also it's worth investigating since there are deeper issues here. I do think the variable `straight_and_flush_size` should be removed and instead `get_straight_and_flush_size` should be implemented like this.

```
int get_straight_and_flush_size(void)
{
    return four_fingers_joker_count > 0 ? 
        STRAIGHT_AND_FLUSH_SIZE_FOUR_FINGERS : STRAIGHT_AND_FLUSH_SIZE_DEFAULT;
}
```

The variable is redundant and can cause bugs if goes out of sync.

STADD:
@MeirGavish I’ve applied the changes thanks for helping me figure out why the size wasn’t updating. I also committed the improvement regarding straight/flush size handling by removing the redundant `straight_and_flush_size`.

MeirGavish:
Merging to have the fix in the next nightly build

Issue #175, opened 2025-10-19
Move game state functions into their own files


ricfehr3:
We currently have all the logic for every game state in `game.c`, however they are separated into functions by their gamestate. It would be useful to have these in each their own file in a sub-directory. 

Imo, these should be behind one include file that includes all the gamestates. Though the organization of how that would work requires some design work. Either xmacros or some callback-registration scheme, idk, if you think you have a good idea then go for it. 

Below is a current list of functions associated with gamestates. Currently this is blocked on #156 , but once the variables of a gamestate are encapsulated then they can be passed as a pointer to the callbacks.

```c
static void game_main_menu_on_init();
static void game_main_menu_on_update();
static void game_round_on_init();
static void game_playing_on_update();
static void game_round_end_on_update();
static void game_round_end_on_exit();
static void game_shop_on_update();
static void game_shop_on_exit();
static void game_blind_select_on_update();
static void game_blind_select_on_exit();
static void game_lose_on_init();
static void game_lose_on_update();
static void game_over_on_exit();
static void game_win_on_init();
static void game_win_on_update();
static void game_shop_intro();
static void game_shop_process_user_input();
static void game_shop_outro();
static void game_blind_select_start_anim_seq();
static void game_blind_select_handle_input();
static void game_blind_select_selected_anim_seq();
static void game_blind_select_display_blind_panel();
static void game_round_end_start();
static void game_round_end_start_expand_popup();
static void game_round_end_display_finished_blind();
static void game_round_end_display_score_min();
static void game_round_end_update_blind_reward();
static void game_round_end_panel_exit();
static void game_round_end_display_rewards();
static void game_round_end_display_cashout();
static void game_round_end_dismiss_round_end_panel();
```

ricfehr3:
Closing this, as the remaining refactor of `game.c` is larger than we thought and worth it's own tracking. We have done everything else up to this point in this issue as of #521 .

emiyl:
Yay!

Issue #124, opened 2025-09-16
Minimize malloc in favor of pools


ricfehr3:
Given the limitations of the GBA, it's good to apply best practices of embedded design. This includes preventing unnecessary dynamic allocation (malloc). It's preferred to have more fine-grained memory management via pools or similar. Below is the uses of malloc in the project:

```c
source/card.c
28:    Card *card = malloc(sizeof(Card));
64:    CardObject *card_object = malloc(sizeof(CardObject));

source/sprite.c
23:    Sprite *sprite = malloc(sizeof(Sprite));
117:    SpriteObject* sprite_object = (SpriteObject*)malloc(sizeof(SpriteObject));

source/joker.c
116:    joker_gfxTiles = (const unsigned int**)malloc((sizeof(unsigned int*) * num_spritesheets));
117:    joker_gfxPal = (const unsigned short**)malloc((sizeof(unsigned int*) * num_spritesheets));
118:    joker_spritesheet_pb_map = (int*)malloc(sizeof(int) * num_spritesheets);
136:    Joker *joker = (Joker*)malloc(sizeof(Joker));
176:    JokerObject *joker_object = malloc(sizeof(JokerObject));

source/list.c
8:    List *list = (List *)malloc(sizeof(List));
10:    list->_array = (void **)malloc(sizeof(void*) * init_size);
```

The reason to refactor this is malloc adds a lot of unpredictability with where memory is used, how much time it takes to alloc/free, and possible fragmentation. On the GBA, there really isn't much of a reason not to allocate a pool at the cost of a bit of ram (e.g. The `Sprite` struct can be pooled in 1.4k RAM with 128 sprites. I'd argue that if memory isn't an issue, then consistency of time to allocate and memory location outweighs the RAM usage). The first-class ADT pattern used in this project should make it trivial to implement pools, but there's some design considerations when dealing with the caveats of each data type. 

I'm opening this issue and will be opening a branch soon to address this.

ricfehr3:
Working on a branch in my fork here:

https://github.com/ricfehr3/balatro-gba/tree/feature/replace-malloc-with-pools

MeirGavish:
Awesome! While I do have quite some experience with C, I don't have experience with embedded development so dynamic allocation felt most natural. This seems like a good opportunity for optimization, thanks a lot!

ricfehr3:
No problem at all. I love this stuff. 

[...]

PR #363, opened 2026-01-15
Separate game logic into its own files

### LINKED ISSUE #304: Document sprite.h with Doxygen

emiyl:
When working on my NDS port of this project, I found the ~5000 line game.c difficult to navigate. This PR aims to separate all states of the game into their own files.

- [x] Separate all states in `DEF_STATE_INFO` to their own files
- [x] Move background changing code to each relevant file
- [x] Remove all externs and replace with getters and setters
- [x] Minimise use of getters and setters with structs to move the data around

Common UI elements used through multiple states are stored in `game/common_ui.c`.

`selection_grid.c` and `selection_grid.h` have been changed to add a `void* ctx`. This is so data like the jokers list can be passed to selection grid items.

```c
int jokers_sel_row_get_size(void* ctx)
{
    JokerSellProps* props = (JokerSellProps*)ctx;
    return list_get_len(props->owned_jokers_list);
}
```

I've also separated commonly used consts and definitions into the following files:
- `include/game/palette.h`
- `include/game/point.h`
- `include/game/timer.h`
- `include/game/rect.h`

Other changes:
- Removed unnecessary arguments to `display_hands()` and `display_discards()` that weren't used


MeirGavish:
Cool, this was definitely planned. Just notice issues #156, #175, and PR #338.

emiyl:
Nice nice, is it okay if I just make this PR a near 1:1 copy of the original code (just in different files ofc), and then any improvements like those issues and PRs can be made later?

MeirGavish:
> Nice nice, is it okay if I just make this PR a near 1:1 copy of the original code (just in different files ofc), and then any improvements like those issues and PRs can be made later?

I suppose so, I think the main question is what you do with the global variables that are accessed by several states, will they be moved to a general struct or accessed with getters and setters. I suppose they can be done with getters and setters for now and proper encapsulation can be done later but I'd also like @ricfehr3's input on this.

emiyl:
I was going to keep a "game.c" file that holds global variables, and just extern them into each file. I was going to restructure it into something else in a separate PR, but this one is just for moving things into separate files.

MeirGavish:
> I was going to keep a "game.c" file that holds global variables, and just extern them into each file. I was going to restructure it into something else in a separate PR, but this one is just for moving things into separate files.

Well, we're not very fond of "extern" so I think it's better to do with getters and setters. 😅
But I don't know, if it's temporary I guess it's not worse than it already is? I'll have to think about it and also wait for @ricfehr3's input.

emiyl:
I’m happy to see if I can put in a new system in this PR then so there’s no externs when it’s merged 😌

ricfehr3:
@emiyl Wow, this is awesome! We've been looking to do this for a while and it's nice to login and see this, so huge thank you!

This is going to be a very big PR. Initially we were going to try to decouple the `game.c` file in multiple PRs methodically into the main branch, but since you've put in a ton of work I think we need to rework the merging strategy to handle all these changes. I'm not looking to slow you down or interrupt your workflow at all, this is great. 

Right now, we don't have full testing coverage to a point where I'm comfortable doing large refactors like this. @MeirGavish I think we need a good way to incorporate large changes like this into project. I'm really not a fan of adding externs into the main branch and fixing it later. I feel that it encourages other developers to do the same and reverses some of the work we put in to avoid it so far. It would be beneficial to also have the global game state working with the changes in this PR (see [discussion](https://github.com/GBALATRO/balatro-gba/pull/338#issuecomment-3696894064) from related PR).

Maybe we should stage it in a `development` branch, or even it's own upstream branch until it's ready for the `development` branch. I think it makes sense to break up this PR as much as possible (we can remove the externs and add docs (every `.h` file needs to be documented with Doxygen) each in their own separate PRs). @MeirGavish We could even move the nightly to the `development` branch only and keep the main branch for stable releases. 

@emiyl Maybe when you have the "1:1 code" working prototype, we can upstream the branch and you can iterate with more changes in separate branches? I'll let @MeirGavish chime in too.

emiyl:
No problem! I have the 1:1 prototype working now and I'm cleaning it up from there. I'm making a load of getters and setters instead of externs.

emiyl:
Alright, everything is in its own file now and game.c is pretty much just filled with getters and setters.

emiyl:
Here's the idea I'm working on:

```c
void* ctx = NULL;

void set_game_state_ctx(enum GameState game_state)
{
    if (!ctx)
    {
        switch (game_state)
        {
            case GAME_STATE_MAIN_MENU:
                ctx = malloc(sizeof(MainMenuProps));
                break;
            case GAME_STATE_BLIND_SELECT:
                ctx = malloc(sizeof(BlindSelectProps));
                break;
            default:
                return;
        }
    }
    if (!ctx) // Check malloc success
        return;
    switch (game_state)
    {
        case GAME_STATE_MAIN_MENU:
        {
            MainMenuProps* props = (MainMenuProps*)ctx;
            props->timer = timer;
            props->rng_seed = rng_seed;
            break;
        }
        case GAME_STATE_BLIND_SELECT:
        {
            BlindSelectProps* props = (BlindSelectProps*)ctx;
            props->timer = timer;
            props->substate = state_info[game_state].substate;
            props->current_blind = current_blind;
            for (int i = 0; i < BLIND_TYPE_MAX; i++)
                props->blinds_states[i] = blinds_states[i];
            props->ante = ante;
            break;
        }
        default:
            break;
    }
    state_info[game_state].ctx = ctx;
}

void update_game_state_ctx(enum GameState game_state)
{
    void* ctx = state_info[game_state].ctx;
    if (!ctx)
        return;
    switch (game_state)
    {
        case GAME_STATE_MAIN_MENU:
        {
            struct MainMenuProps* props = (struct MainMenuProps*)ctx;
            rng_seed = props->rng_seed;
            break;
        }
        case GAME_STATE_BLIND_SELECT:
        {
            BlindSelectProps* props = (BlindSelectProps*)ctx;
            timer = props->timer;
            state_info[game_state].substate = props->substate;
            current_blind = props->current_blind;
            for (int i = 0; i < BLIND_TYPE_MAX; i++)
                blinds_states[i] = props->blinds_states[i];
            ante = props->ante;
            break;
        }
        default:
            break;
    }
}

void game_update()
{
    // ...
    set_game_state_ctx(game_state);
    state_info[game_state].on_update(state_info[game_state].ctx);
    update_game_state_ctx(game_state);
}

void game_change_state(enum GameState new_game_state)
{
    ctx = NULL;      // Reset context pointer
    // ...
}
```

```c
typedef struct
{
    int substate;
    void (*on_init)(void*);
    void (*on_update)(void*);
    void (*on_exit)(void*);
    void* ctx;
} StateInfo;

StateInfo state_info[] = {
#define DEF_STATE_INFO(stateEnum, init_fn, update_fn, exit_fn) \
    {.on_init = init_fn, .on_update = update_fn, .on_exit = exit_fn, .substate = 0, .ctx = NULL},
#include "../include/def_state_info_table.h"
#undef DEF_STATE_INFO
};
```

So each function has a "ctx" which is then cast into whatever struct is necessary for each function. A switch statement determines what data goes out and what data goes back in again.

emiyl:
this is probably nearly done

source/game.c
```diff
@@ -29,4716 +38,937 @@
...
+                break;
+            case GAME_STATE_SHOP:
+                ctx = malloc(sizeof(ShopProps));
+                break;
+            case GAME_STATE_LOSE:
+            case GAME_STATE_WIN:
+                ctx = malloc(sizeof(GameOverProps));
+                break;
+            default:
+                return;
```
ricfehr3: [Related](https://github.com/GBALATRO/balatro-gba/pull/363#issuecomment-3765622423)

I do like this idea. Keeping properties for each screen in a container is both flexible and isolates the variables well. @MeirGavish do you agree? It would be a good hybrid solution to our [discussion](https://github.com/GBALATRO/balatro-gba/pull/338#issuecomment-3696894064) in #338 . 

I would like to suggest a change. We don't want to use malloc at all, and I think we can avoid it and simplify the logic a bit. Let me know what you think @emiyl : 

Since the `ctx` here is only going to be the structs: `MainMenuProps`, `BlindSelectProps`, `RoundProps`, `RoundEndProps`, `ShopProps`, and `GameOverProps`. I think it would be better to wrap these in a union:

```c
union Ctx {
    MainMenuProps main_menu;
    ...
    GameOverProps game_over;
};
```

This way, pointers aren't needed at all and the callback function will just read it as a union member: `ctx->game_over`, for example. I think a pointer to this context in the update callbacks should still be passed to keep the parameter small. So `state_info[game_state].ctx` would remain relatively intact, but it will be typed to `union Ctx`. It just will always be the size of the `union Ctx` union and always be allocated. 

---

P.S. Still reading this one. I'm doing a review, but I thought I'd comment on this first. Thanks!

emiyl: The union type seems fine to me. Out of curiosity, what's bad about malloc?

MeirGavish: > The union type seems fine to me. Out of curiosity, what's bad about malloc?

It's inefficient in the long run because it needs to be generic in the allocated memory type and can have fragmentation.
See issue #124 and corresponding PR #132 to remove malloc from the codebase which did improve performance.

I'm a little busy this week so I will chime in later to add my opinion regarding the question, probably tomorrow.

MeirGavish: This is a big PR, so it will take me a bit, I will try to find time write a few things tomorrow.

MeirGavish:
Sorry for the delay with this, it's been a little difficult for me to get to it, I will try to get to it soon and add my input.

include/game/point.h
```diff
@@ -0,0 +1,18 @@
...
+// Game points
+static const BG_POINT HELD_JOKERS_POS = {108, 10};
+static const BG_POINT JOKER_DISCARD_TARGET = {240, 30};
+
+// Round phase points
+static const BG_POINT CARD_DRAW_POS = {208, 110};
+static const BG_POINT CARD_DISCARD_PNT = {240, 70};
+static const BG_POINT HAND_START_POS = {120, 90};
+static const BG_POINT HAND_PLAY_POS = {120, 70};
+static const BG_POINT PLAYING_SCORE_FLAME_CHIPS_POS __attribute__((unused)) = {176, 80};
```
ricfehr3, line 15: In `util.h` we have: "#define GBAL_UNUSED __attribute__((unused))"

include/game/rect.h
```diff
@@ -0,0 +1,94 @@
...
+static const Rect ROUND_END_BLIND_REWARD_RECT = {168, 96, UNDEFINED, UNDEFINED};
+static const Rect CASHOUT_TEXT_RECT = {88, 72, UNDEFINED, UNDEFINED};
+static const Rect SHOP_REROLL_RECT = {88, 96, UNDEFINED, UNDEFINED};
+static const Rect GAME_LOSE_MSG_TEXT_RECT = {104, 72, UNDEFINED, UNDEFINED};
+static const Rect GAME_WIN_MSG_TEXT_RECT = {112, 72, UNDEFINED, UNDEFINED};
+
+// Round phase rects
+static const Rect PLAYED_CARDS_SCORES_RECT = {72, 48, 240, 56};
+static const Rect HELD_CARDS_SCORES_RECT = {72, 108, 240, 116};
+static const Rect PLAYING_TEMP_SCORE_RECT __attribute__((unused)) = {128, 144, 200, 152};
```
ricfehr3, line 32: replace `__attribute__((unused))` with `GBAL_UNUSED`

include/game/rect.h
```diff
@@ -0,0 +1,94 @@
...
+static const Rect HAND_SIZE_RECT_SELECT = {128, 128, 152, 136};
+static const Rect HAND_SIZE_RECT_PLAYING = {128, 152, 152, 160};
+static const Rect HAND_TYPE_RECT = {8, 64, 64, 72};
+static const Rect TEMP_SCORE_RECT = {8, 64, 64, 72};
+static const Rect SCORE_RECT = {24, 48, 64, 56};
+
+// Blind rects
+static const Rect BLIND_REWARD_RECT = {40, 32, 64, 40};
+
+// clang-format on
```
ricfehr3, line 92: The code between the `clang-format off/on` is clang-formatted. The previous format was done manually and is preferred. It needs to be formatted as it was originally in `game.c`.

include/game/round.h
```diff
@@ -0,0 +1,61 @@
+#ifndef __INCLUDE_GAME_ROUND_H__
+#define __INCLUDE_GAME_ROUND_H__
```
ricfehr3, line 2:
```suggestion
#ifndef GAME_ROUND_H
#define GAME_ROUND_H
```
Consistency with other headers

include/game.h
```diff
@@ -63,82 +61,180 @@ enum HandState
...
 {
     int substate;
-    void (*on_init)();
-    void (*on_update)();
-    void (*on_exit)();
+    void (*on_init)(GameStateCtx*);
+    void (*on_update)(GameStateCtx*);
+    void (*on_exit)(GameStateCtx*);
+    GameStateCtx ctx;
 } StateInfo;
```
ricfehr3, line 71: I've been thinking about the use of the GameStateCtx in this struct. I think it should be removed. 

Below, in `game_update()`:
1.  `set_game_state_ctx(game_state)` is called. 
2. `set_game_state_ctx(game_state)` 
	- First, it modifies the global variable `ctx`
	- Then it sets sets `state_info[game_state]`
3. The `state_info[game_state].on_update(&state_info[game_state].ctx)` update callback is ran, reading the copied `ctx`.
4. `update_game_state_ctx(game_state)`is called.
	- First it copies `GameStateCtx ctx = state_info[game_state].ctx` and reads the new `ctx` variable (not to be confused with the initial `ctx` global variable)

Similarly, the other callback functions `on_init(...)` and `on_exit(...)` also get passed `ctx` of the state info `state_info[game_state].on_exit(&state_info[game_state.ctx])`. There's a lot of copying going on and it's not so clear how the main `ctx` variable gets set because of the misdirection. 

There isn't a need to pass around the `ctx` at all. You can just pass the pointer to the static scoped variable `static GameStateCtx ctx` directly and modify it safely without copies at all. This removes the need for the `state_info` `ctx` variable. Additionally, for reading from the `ctx` variable to process further (such as in `update_game_state_ctx(...)`), I see full copies `ctx = state_info[game_state].ctx`. It's better to just get the pointer. The `ctx` is about 640 bytes long, so multiple copies per frame start to pile up.

Makefile
```diff
@@ -81,14 +81,15 @@ ifneq ($(BUILD),$(notdir $(CURDIR)))
 
 export OUTPUT	:=	$(CURDIR)/$(BUILD)/$(TARGET)
 
-export VPATH	:=	$(foreach dir,$(SOURCES),$(CURDIR)/$(dir)) \
+export VPATH	:=	$(CURDIR)/source \
```
ricfehr3, line 84: There is a better way to do the Makefile changes. `SOURCES` just needs to be appended to. You can apply the patch below:

```make
diff --git a/Makefile b/Makefile
index cee931b..fccfd8a 100644
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ LIBTONC := $(DEVKITPRO)/libtonc
 TARGET         := $(notdir $(CURDIR))
 BUILD          := build
-SOURCES	       := source
+SOURCES	       := source source/game
 INCLUDES       := include
 DATA           :=
 MUSIC          := audio
@@ -81,15 +81,14 @@ ifneq ($(BUILD),$(notdir $(CURDIR)))

 export OUTPUT	:=	$(CURDIR)/$(BUILD)/$(TARGET)

-export VPATH	:=	$(CURDIR)/source \
-			$(CURDIR)/source/game \
+export VPATH	:= $(foreach dir,$(SOURCES),$(CURDIR)/$(dir)) \
 			$(foreach dir,$(DATA),$(CURDIR)/$(dir)) \
 			$(foreach dir,$(GRAPHICS),$(CURDIR)/$(dir)) \
 			$(foreach dir,$(FONT),$(CURDIR)/$(dir)) \

 export DEPSDIR	:=	$(CURDIR)/$(BUILD)

-CFILES		:=	$(foreach dir,$(SOURCES),$(notdir $(wildcard $(dir)/*.c $(dir)/**/*.c)))
+CFILES		:=	$(foreach dir,$(SOURCES),$(notdir $(wildcard $(dir)/*.c)))
 CPPFILES	:=	$(foreach dir,$(SOURCES),$(notdir $(wildcard $(dir)/*.cpp)))
 SFILES		:=	$(foreach dir,$(SOURCES),$(notdir $(wildcard $(dir)/*.s)))
 PNGFILES	:=	$(foreach dir,$(GRAPHICS),$(notdir $(wildcard $(dir)/*.png)))
```

source/joker.c
```diff
@@ -1,6 +1,7 @@
 #include "joker.h"
 
 #include "card.h"
+#include "game/round.h"
```
ricfehr3, line 4: @MeirGavish , what do you think about including subdirs? I think I would prefer to just add the path to the include path. But it's just style really.

source/game.c
```diff
@@ -29,4716 +38,924 @@
...
-        memset16(&pal_bg_mem[BLIND_SKIP_BTN_SELECTED_BORDER_PID], 0xFFFF, 1);
-    }
+    state_info[game_state].ctx = ctx;
 }
 
-static void game_blind_select_selected_anim_seq()
+void update_game_state_ctx(enum GameState game_state)
 {
-    if (timer < 15)
+    GameStateCtx ctx = state_info[game_state].ctx;
```
ricfehr3, line 805: Here, you copy to a locally scoped variable named the same as the file-scoped variable above `ctx`. I'm not sure if that was intended.

source/game.c
```diff
@@ -29,4716 +38,924 @@
...
-        game_blind_select_erase_blind_reqs_and_rewards();
-
-        if (selection_y == 0) // Blind selected
+        case GAME_STATE_MAIN_MENU:
         {
-            play_sfx(SFX_BUTTON, MM_BASE_PITCH_RATE, BUTTON_SFX_VOLUME);
-            state_info[game_state].substate = BLIND_SELECTED_ANIM_SEQ;
-            timer = TM_ZERO;
-            display_round(++round);
+            MainMenuProps props = ctx.main_menu;
```
ricfehr3, line 709: I would operate on a pointer so there's no need to copy it back.

ricfehr3 (changes requested):
Sorry for being so late on this. Thanks again for doing this. It's a great deal of work, and I'll be back a few more times to further review. It is a lot to go through. This time through I do have some changes I'd like to see in this PR. I'm also getting a list of issues to create for further PRs on this refactor. 

Thanks!

ricfehr3:
@emiyl I would like to continue incorporating these changes, but it is a big PR to go through. I wanted to propose a path forward that retains your authorship and contributions but is more bite sized so we can do a better / faster review. It also doesn't require any work on your end, as long as you're okay with the plan. 

I split your branch into desired smaller branches that can be introduced incrementally in the existing related issue #156, so we can avoid a bigger headache of reviewing this PR and adding it as a massive change. 

What I would like to do is merge this PR into a separate upstream branch, and pick out changes to separate it into smaller feature branches. Each of those branches will get a separate PR, and be merged as one squash commit. I'll make sure each final squash commit is co-authored by you so you can get the credit.

If you'd like to separate the changes into the branches matching that checklist, that would be epic. But otherwise, no work on your end is required and I can handle it. Thanks!

emiyl:
Unfortunately I’m swamped with deadlines so I don’t have time to devote to the project at the moment :( But thank you for considering and crediting me!

ricfehr3:
> Unfortunately I’m swamped with deadlines so I don’t have time to devote to the project at the moment :( But thank you for considering and crediting me!

No problem on giving you credit, your work is appreciated! I'll continue on the plan then. Thanks!

ricfehr3:
Going to close this one since we've completed #175 . Thank you for this PR tho!

PR #366, opened 2026-01-17
Switch example gameplay gif to updated video in README

MeirGavish:
I've had a hard time with gifs and they get a little blurry because you can only save them in native resolution so I thought I'll just switch to a video.
I played a few rounds and just recorded a round with whatever jokers I got, I got Seltzer so it ended up being a nice hand with retriggers and lots of score increases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,9 @@ This **tech-demo/proof of concept** is strictly limited in content to a minimal 
 
 ---
 
-<!-- The Gif is a little blurry but I think it looks fine -->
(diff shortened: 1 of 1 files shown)


GreenNinja2525 (approved):
Looks good, nice round too. It doesn't need to be a gif, this looks better quality too

MathisMartin31:
Yes, very nice indeed! We have retriggers, lots of Jokers and the little score flames :)

MeirGavish:
There seems to be general approval for this so I'm merging.

PR #369, opened 2026-01-22
Fix selecting a card by accident when moving too fast on round start

### LINKED ISSUE #364: [BUG] If you move the card selection cursor as soon as you can, the first card gets selected
**Describe the bug**
If you move the card selection cursor as soon as you can, the first card gets selected even though you didn't press the selection button "A".

**To Reproduce**
Start a round, as soon as the cards are drawn, or right after a played hand or discard, move the selection cursor, the card at the default selection position will be selected even the selection button wasn't pressed.

**Expected behavior**
The card should not be selected.

**Screenshots**

https://github.com/user-attachments/assets/12514db8-df66-48a9-a65f-305ec5c0cb1e

**What version**
Main branch with latest commit "Refactored to use selection grid during round (https://github.com/GBALATRO/balatro-gba/pull/348)" 5300846

mGBA, as well as original hardware with EZ-flash flashcart.

**Additional context**
This is obviously from the code that detects quick drag and changes it to selection, the question is why is it activated with no selection.
This is a small bug that doesn't have a big impact so I'm wondering whether to delay the release or just release anyway and solve this later.

MathisMartin31:
Pretty much the title ^^

Related to our discussion in #364 

The `selection_hit_timer` is equal to UNDEFINED until we actually hit A, and the check has been modified to reflect that

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -2012,7 +2012,7 @@ static bool card_selected_instead_of_moved = false;
 // and change focus to the next one, instead of swapping them
 // This should fix inputs sometimes not registering when quickly selecting cards
 static const int card_swap_time_threshold = 6;
-static uint selection_hit_timer = TM_ZERO;
+static uint selection_hit_timer = UNDEFINED;
 
 static bool game_playing_hand_row_on_selection_changed(
     SelectionGrid* selection_grid,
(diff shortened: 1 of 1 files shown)


source/game.c
```diff
@@ -2012,7 +2012,7 @@ static bool card_selected_instead_of_moved = false;
 // and change focus to the next one, instead of swapping them
 // This should fix inputs sometimes not registering when quickly selecting cards
 static const int card_swap_time_threshold = 6;
-static uint selection_hit_timer = TM_ZERO;
+static uint selection_hit_timer = UNDEFINED;
```
MeirGavish: Oh, I thought the timer was an int, not a uint, not too happy putting UNDEFINED in an unsigned value... Need to give it some thought...

MathisMartin31: We could always revert back to setting the timer to `TM_ZERO` and add a `A_pressed` bool, this functionally does the same thing

MeirGavish: Actually, I think `timer` and `selection_hit_timer` should be changed to int.

MathisMartin31: Should be good now

MeirGavish (approved):
Looks good! Merging

PR #370, opened 2026-01-25
Fix cards in deck count shifting when too few cards remain

### LINKED ISSUE #181: Card counter displays 522 instead of 52 sometimes
I couldn't replicate the glitch, as far as I know, it's just random, though if that helps I had the Blue Joker on me when it happened, so maybe it has to do with that, but I'm too lazy to reroll runs until I find one

MathisMartin31:
Fix for issue #181 

The fix is assuming we will have no more than a 2 digits number of cards in our deck. This will need refactoring when deck fixing will be possible with Consumables but for now the deck is fixed at 52 cards

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -404,7 +404,7 @@ static const Rect SINGLE_BLIND_SEL_REQ_SCORE_RECT = {80, 120,    104,     128  }
 // to be used with tte_erase_rect_wrapper()
 static const Rect HANDS_TEXT_RECT           = {16,      104,    UNDEFINED, UNDEFINED };
 static const Rect DISCARDS_TEXT_RECT        = {48,      104,    UNDEFINED, UNDEFINED };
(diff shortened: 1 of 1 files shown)


MeirGavish (changes requested):
Actually, I think what's missing here is a call to `tte_erase_rect_wrapper(DECK_SIZE_RECT);`.
DECK_SIZE_RECT needs to be fully defined with the bottom coordinate, 240, 160 for the edge of the screen should do it.
The text will then be left aligned but I think it wouldn't look too bad even if the full deck size is in the single digits. If there are more than 99 cards in the deck it will overflow outside the screen of course so maybe we should add a TODO in the code for that but for now I think it's a good enough fix.

MathisMartin31:
Should be good now ^^

MeirGavish (approved):
Cool, merging

PR #380, opened 2026-02-08
add profiler demo

ricfehr3:
@MathisMartin31 @MeirGavish Sup guys. I stumbled across this simple profiler that comes built into tonc. `profile_start();` Starts the timer, `u32 time_passed = profile_stop();` stops the timer and returns the raw cycle count between calling `profile_start();` and `profile_stop()`. In the example below printing on the screen is very slow, so slow it makes the `frame_diff` variable unusable. However, if you use the function `profile_timer_breakpoint();` as a breakpoint, you can easily get the frame time differences with gdb. 

Not sure where this will be useful, but I thought I'd just let y'all and anyone else know.

This PR is just for discussion / example. Not going to be merged, at least not in this state.

[Doc link in tonc](https://www.coranac.com/man/tonclib/group__grpTimer.htm#ge8cb66eb4453b22b5eb31262c19517cc)

### DIFF
--- a/source/main.c
+++ b/source/main.c
@@ -21,6 +21,7 @@
 #include "audio_utils.h"
 #include "soundbank.h"
 #include "soundbank_bin.h"
+#include "tonc_core.h"
 
 void init()
 {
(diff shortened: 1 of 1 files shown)


MeirGavish:
Cool! I think this could be useful to test optimizations or maybe check performance impact of changes

MeirGavish:
I think this should be merged as conditionally compiled code in one form or another so we have an integrated method to test performance that doesn't go out of sync. It could even be better if we could add a workflow so we can trigger performance testing on PRs so we can easily know if they impact performance, but I do realize it would take a lot of work, maybe it can be integrated with #407.
I made some modifications to this script so it works for me when debugging with VSCode, I have them in a branch somewhere, I will try to make a draft PR in the next few days so you can see the changes.

ricfehr3:
> I think this should be merged as conditionally compiled code in one form or another so we have an integrated method to test performance that doesn't go out of sync. It could even be better if we could add a workflow so we can trigger performance testing on PRs so we can easily know if they impact performance, but I do realize it would take a lot of work, maybe it can be integrated with #407. I made some modifications to this script so it works for me when debugging with VSCode, I have them in a branch somewhere, I will try to make a draft PR in the next few days so you can see the changes.

Sweet! That sounds great. I have also been thinking of a way to profile the game. I was thinking of adding breaks in the state machine code. I'll wait for what you have with your VSCode script and we can go from there. I'll make an issue to track this as well.

PR #379, opened 2026-02-08
Document `sprite.h` w/ minor optimizations / fixes

ricfehr3:
Document `sprite.h` and remove unused functions and variables.

Closes #304

PR #381, opened 2026-02-11
Sprite documentation addendum note

MeirGavish:
Addendum to #379


MeirGavish:
> Nice catch! Sorry I missed this one.
Nah, it's okay, I just thought I should add this after my initial review

PR #387, opened 2026-02-24
I added 4 cards

FuriousTroller:
Golden Joker
Gros Michel
Cavendish
Supernova

Known bug: If Golden Joker and Gros Michel happened to trigger its effects at the same time at the end of the round, "Extinct!" will appear as "inct!" due to them overlapping each other (visual bug)
image


FuriousTroller:
> Golden Joker Gros Michel Cavendish Supernova
> 
> Known bug: If Golden Joker and Gros Michel happened to trigger its effects at the same time at the end of the round, "Extinct!" will appear as "inct!" due to them overlapping each other (visual bug) image

I recolored the Joker to yellow, and made sure that each sprite file does not contain more than 16 color, the two bananas were disastrous, but it's true human effort (hence the quality)

MeirGavish:
Cool! Though unfortunately we want to limit the scope of the game to not compete with the original Balatro so the project is not taken down by Playstack. We limited the number of jokers in the scope and we already reached the limit of 52 implemented jokers so we currently don't plan to add any more.
See [the scope discussion](https://github.com/GBALATRO/balatro-gba/discussions/355).

FuriousTroller:
> Cool! Though unfortunately we want to limit the scope of the game to not compete with the original Balatro so the project is not taken down by Playstack. We limited the number of jokers in the scope and we already reached the limit of 52 implemented jokers so we currently don't plan to add any more. See [the scope discussion](https://github.com/GBALATRO/balatro-gba/discussions/355).

Thanks Gavish, for the hasty response. I just checked, and it totally makes sense to limit to 52 Jokers. I happened to know another Balatro project taken down this way, too. Guess I'd commit it to my fork then. Also, I do have a question for my friend who is also modding his version of this very fork. He is adding an AI versus mode to GBAlatro, and this is not an original feature of Balatro, would it be violating any rules of LocalThunk regarding this?

bzxgcs:
Amazing work! New content updates are always exciting.

bzxgcs:
I tried your version, but the ante doesn’t seem to increase — it always stays at 1. How can I submit an issue for your version?

FuriousTroller:
> I tried your version, but the ante doesn’t seem to increase — it always stays at 1. How can I submit an issue for your version?

Oh I just checked, I think it's a visual bug on my side, messing up with the game.c a bit. Thank you for your feedback, I am maintaining this fork with a friend too, I hope he can come up with a solution quickly.

MeirGavish:
I admit it's interesting to see this fork and as before I'm a little sad to give up on jokers but since this fork seems to continue in a different path of adding versus mode and custom jokers, I'll close this PR. Good luck!

Issue #216, opened 2025-11-08
Develop Joker Test Plan


ricfehr3:
With any changes, we need to ensure that the jokers operate correctly. We should add some sort of test or write a procedure to ensure they always are operating correctly. 

This could be a deceptively difficult task and I haven't put much thought into it. It is relatively low priority, but noting it here to track.

MeirGavish:
100% agree with this, it's been on my mind a lot. It will definitely be difficult because in general jokers have wildly differing effects that can be activated at different events and affect different aspects of the game. Our current jokers only affect the round itself so if we encapsulate the round variables maybe we can make unit tests for the current set of jokers. There will be other difficulties for this too...

MathisMartin31:
What I've been doing the past couple of weeks is having a list of Jokers to add from the start. It's not automated, but not entirely by hand either.

```
// Joker testing: define up to MAX_JOKERS_HELD_SIZE Joker IDs to have from the start to test them
static u8 test_jokers[MAX_JOKERS_HELD_SIZE] = {ID_1, ID_2, ID_3, ID_4, ID_5};
static u8 nb_test_jokers = 5;

...

void game_init()
{
    ...
    // Joker testing
    // Fill Jokers List with desired Jokers if nb_test_jokers is not 0
    for (u8 i = 0; i < nb_test_jokers; i++)
    {
        JokerObject *joker_object = joker_object_new(joker_new(test_jokers[i]));
        add_to_held_jokers(joker_object);
    }
    ...
}
```

It's kind of ugly, but I have a git stash with it that I apply when I want to test Joker combinations.

I agree with @MeirGavish, encapsulating game/round variables could make it easier to test Jokers. We could activate the test suite from the main menu for example, with a button combo. We could then have defined Joker lists and Hands that could replace what we would normally draw. For instance, first hand could be all the Sin Jokers and a hand with a suite made of cards from all 4 suits, etc

ricfehr3:
Assigning myself so we don't double up on work. But after #168 is merged, I'll be focusing on this. My immediate plan to  complete this ticket is:
- [ ] Write out a test plan that tries test all the jokers fully in the fewest steps possible. Jokers will be tested as a series of hands with the following information:
  - [ ] Detail all jokers in a hand
  - [ ] Detail all cards in a hand
  - [ ] Detail the sequence of cards played and the expected results
- [ ] Add code to deal custom hands including cards and jokers
[...]

PR #407, opened 2026-03-28
Add e2e tests

macabeus:
As discussed on https://github.com/GBALATRO/balatro-gba/discussions/405, this PR wires [`gba-kit`](https://github.com/macabeus/gba-kit) to add e2e tests here.

How it's set up:

- Added a new nodejs project on `./tests-2e`, using [vitest](https://vitest.dev/) as the test suite
- When starting the tests, it automatically calls `./tests-e2e/helpers/extract-addresses.sh` to extract the addresses for relevant variables used in the assertion. The extraction uses `arm-none-eabi-nm`
- There are 6 tests: 2 for the splash, 2 for the blind, and 2 for the hand. They are useful as a foundation when you need to add new tests.
- You can run the tests locally using `cd tests-e2e && npm run test` or, in watch mode: `npm run test:watch`
- Test runs automatically at the end of the `build` workflow from GitHub Actions

> Note: `gba-kit` is a new package still at `0.1.0`, so it might have some bugs still to be found and fixed. Please, let me know if you encounter any issue 🙏 

image


.github/workflows/build_ci_workflow.yml
```diff
@@ -69,6 +69,29 @@ jobs:
...
+        working-directory: tests-e2e
+        run: npx vitest run
+
+      - name: Upload E2E test output
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-output
+          path: tests-e2e/output/
+          retention-days: 5
```
macabeus, line 97: For now, there is no output. Still, I'm keeping it here since it might be useful later.

ricfehr3:
Hey, this is awesome. I'm going to need to figure out my way around what all is going on here. Is there a way to watch the tests being ran in-browser at all?

macabeus:
> Is there a way to watch the tests being ran in-browser at all?

Not yet. We have the "Recorder" tab on the webapp, but it doesn't work for these tests, since they call helpers and constant variables written by `extract-addresses.sh`.

It's possible to extend it to receive commands from the dev server, but I didn't implement it yet.
That's a good idea btw

MeirGavish:
I want to look at this but it seems like I won't have time in the upcoming weeks so I'm removing my review request, don't wait for my approval to merge.

tests-e2e/helpers/extract-addresses.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
```
ricfehr3: The indent width in this file should be changed to 4-spaces to be consistent with other scripts.

tests-e2e/helpers/extract-addresses.sh
```diff
@@ -0,0 +1,64 @@
...
+# Extract symbol addresses from the ELF file into addresses.json.
+# Tries local arm-none-eabi-nm first, falls back to Docker.
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="${SCRIPT_DIR}/../.."
+ELF="${REPO_ROOT}/build/balatro-gba.elf"
```
ricfehr3: Same exact issue seen [here](https://github.com/GBALATRO/balatro-gba/pull/420#discussion_r3025342852). 

Could you also follow the pattern in `get_memory_map.sh`

tests-e2e/helpers/extract-addresses.sh
ricfehr3, line 1: This file should be renamed `extract_addresses.sh` to be consistent with other scripts.

tests-e2e/helpers/extract-addresses.sh
```diff
@@ -0,0 +1,64 @@
...
+    exit 1
+  fi
+}
+
+# Run nm once and reuse the output
+NM_OUTPUT=$(get_nm_output)
+
+# Write to a temp file, then atomically rename.
+# This prevents corruption when multiple processes run concurrently.
+TMP="${OUT}.$$"
```
ricfehr3: Clever!

tests-e2e/helpers/extract-addresses.sh
```diff
@@ -0,0 +1,64 @@
...
+NM_FLAGS="--print-size"
+
+get_nm_output() {
+  if command -v arm-none-eabi-nm &>/dev/null; then
+    arm-none-eabi-nm $NM_FLAGS "$ELF"
+  elif [ -x "/opt/devkitpro/devkitARM/bin/arm-none-eabi-nm" ]; then
+    /opt/devkitpro/devkitARM/bin/arm-none-eabi-nm $NM_FLAGS "$ELF"
+  elif command -v docker &>/dev/null; then
+    docker run --rm -v "${REPO_ROOT}:/balatro-gba" gbalatro:dev \
+      /opt/devkitpro/devkitARM/bin/arm-none-eabi-nm $NM_FLAGS /balatro-gba/build/balatro-gba.elf
```
ricfehr3:
```suggestion
  if [ -x "$DEVKITARM/bin/arm-none-eabi-nm" ]; then
    "$DEVKITARM/bin/arm-none-eabi-nm" $NM_FLAGS "$ELF"
  elif command -v arm-none-eabi-nm &>/dev/null; then
    arm-none-eabi-nm $NM_FLAGS "$ELF"
```

I think this check can be simplified and reworked a bit.

It should check for the devkitARM version of `arm-none-eabi-nm` first. If the user has a cross compiler already installed there could be some version disparity. However, it will still probably work, so you can run it anyway, just as the second option. DEVKITARM should also be an environment variable in the shell too. It's required to run `make`.

For Docker, rather than run docker in this script, I think it would make more sense to assume that this script can be ran in the docker container.

ricfehr3 (changes requested):
I have some comments for the shell script. I really have no idea how to review the typescript 😅. Once we discuss and work on those comments, I'm happy to merge this in. I'm very interested in seeing more tests with this. It would be great to achieve the dream test coverage required for #216 . Thanks!

MeirGavish:
I'll request a review from Copilot so it might help with the Typescript.

macabeus:
Thanks for the comments!
I'm focused now on another personal project, but I can take it on ~Friday.

macabeus:
Done!
I was busy with [another personal project](https://github.com/macabeus/transmuter), but I have some spare time now to handle the comments.

I think it's all done now.

cc @MeirGavish @ricfehr3

CONTRIBUTING.md
```diff
@@ -17,7 +17,8 @@ Backgrounds may use more colors but notice that their palette is encoded in thei
 On pull-requests, various checks will be performed:
 1. **Formatting**: `clang-format` will be ran on every `.c/.h` file with [this configuration](https://github.com/GBALATRO/balatro-gba/blob/main/.clang-format). Failures will cause the CI to fail.
 2. **Unit Tests**: Unit tests are required to pass and are located in the repo [here](https://github.com/GBALATRO/balatro-gba/tree/main/tests).
-3. **Rom Build**: The ROM must successfully build with the `make -j$(nproc)` command.
+3. **E2E tests**: Automated E2E tests run over the latest ROM build. Read more about it on [tests_e2e/README.md](./tests_e2e/README.md).
+4. **Rom Build**: The ROM must successfully build with the `make -j$(nproc)` command.
```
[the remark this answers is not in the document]
ricfehr3: Don't worry about this, lol.

tests_e2e/helpers/extract_addresses.sh
```diff
@@ -0,0 +1,65 @@
...
+# This prevents corruption when multiple processes run concurrently.
+TMP="${OUT}.$$"
+
+echo "{" > "$TMP"
+FIRST=true
+for sym in $SYMBOLS; do
+    # --print-size output: "address size type name"
+    LINE=$(echo "$NM_OUTPUT" | grep " ${sym}$")
+    if [ -n "$LINE" ]; then
+        ADDR=$(echo "$LINE" | awk '{print "0x"$1}')
```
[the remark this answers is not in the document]
ricfehr3: I agree here, appending `|| true` is what I would recommend as well.

ricfehr3 (changes requested):
Hey, just that one small change from copilot that I commented on in terms of the bash. If you agree with the other comments, feel free to add. Once that's finished, I'll merge. Thank you!

macabeus:
@ricfehr3 Done!

ricfehr3 (changes requested):
Hey, thanks for updating this again. Sorry for the late response and sorry to ask for changes again. I noticed that a lot of variables are hardcoded into the tests. I was wondering, how difficult would it be to provide a way of mapping the actual variable names to variables specifically for the scripts? The variable names may change, and are changing now, so it would be nice to have a central place to map that. Great work tho!

macabeus:
@ricfehr3 Do you mean the variables defined on [`addresses.ts`](https://github.com/macabeus/balatro-gba/blob/bef42250fd6c76ebcdbab639be046290d5a3c651/tests_e2e/helpers/addresses.ts) and symbols on [`extract_addresses.sh`](https://github.com/macabeus/balatro-gba/blob/bef42250fd6c76ebcdbab639be046290d5a3c651/tests_e2e/helpers/extract_addresses.sh#L18)?

I think it's not feasible to automatically map the variable names.
If the variable name changed in the source code, it would be necessary to rename it in the e2e testing too.

Although it's not hard to keep them (manually) in sync because the test will fail.

ricfehr3:
> @ricfehr3 Do you mean the variables defined on [`addresses.ts`](https://github.com/macabeus/balatro-gba/blob/bef42250fd6c76ebcdbab639be046290d5a3c651/tests_e2e/helpers/addresses.ts) and symbols on [`extract_addresses.sh`](https://github.com/macabeus/balatro-gba/blob/bef42250fd6c76ebcdbab639be046290d5a3c651/tests_e2e/helpers/extract_addresses.sh#L18)?
> 
> I think it's not feasible to automatically map the variable names. If the variable name changed in the source code, it would be necessary to rename it in the e2e testing too.
> 
> Although it's not hard to keep them (manually) in sync because the test will fail.

I agree, automatic naming is a lot of work. I was hoping we could use a central configuration file like: `var_names.json`, that can map the real variable names with the ones handled in these e2e-tests. So the code here never changes and all the variable names can be updated in a central place, rather than modify the files directly. What do you think?

macabeus:
@ricfehr3 Done!

ricfehr3:
@macabeus Nice! Thanks for the changes :) I haven't forgotten this PR, I'll be reviewing it in the next few days. Thanks!

.github/workflows/build_ci_workflow.yml
```diff
@@ -67,11 +67,34 @@ jobs:
           echo "GID=$(id -g)" >> "$GITHUB_ENV"
 
       - name: Build the project
-        run: docker compose -f docker-compose.yml run --rm gbalatro sh -c "make -j$(nproc) && ./scripts/get_memory_map.sh"
+        run: docker compose -f docker-compose.yml run --rm gbalatro sh -c "make -j$(nproc) && ./scripts/get_memory_map.sh && ./tests_e2e/helpers/extract_addresses.sh"
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
```
ricfehr3: It looks like this is a couple versions behind as `setup-node` is on version 6 now. https://github.com/marketplace/actions/setup-node-js-environment

Is there a reason to keep it on version 4 or should we consider bumping that version.

.github/workflows/build_ci_workflow.yml
```diff
@@ -67,11 +67,34 @@ jobs:
...
+        with:
+          node-version: 22
```
ricfehr3: Is it possible to move this to `24` since it's the most recent LTS version of node? https://nodejs.org/en/about/previous-releases

.github/workflows/build_ci_workflow.yml
```diff
@@ -67,11 +67,34 @@ jobs:
...
+        working-directory: tests_e2e
+        run: npm ci
+
+      - name: Run E2E tests
+        working-directory: tests_e2e
+        run: npx vitest run
+
+      - name: Upload E2E test output
+        if: failure()
+        uses: actions/upload-artifact@v4
```
ricfehr3: This action is on version 7 now: https://github.com/marketplace/actions/upload-a-build-artifact

Should we update it to the most recent version?

tests_e2e/helpers/addresses.ts
```diff
@@ -0,0 +1,81 @@
...
+export const GameState = {
+  SPLASH_SCREEN: 0,
+  MAIN_MENU: 1,
+  PLAYING: 2,
+  ROUND_END: 3,
+  SHOP: 4,
+  BLIND_SELECT: 5,
+  LOSE: 6,
+  WIN: 7,
+} as const;
```
ricfehr3: This will need to be updated with the new OPTIONS, and RUN_SETUP. Is there a way to extract this and the other types below from the files directly? It's okay if it's difficult, I doubt these values will change too much.

ricfehr3 (changes requested):
Hey again! I wanted to wait after the release to merge this, but I think we are ready to bring it in. I left some comment on some things I noticed, but otherwise once rebased and the variable comment is addressed. I think it should be ready.

macabeus:
@ricfehr3 Done!
The PR is rebased, and the dependencies are updated.

I added support for reading DWARF on `gba-kit`. It enabled a lot of simplification and better maintenance on the tests.
Now, it can read the right memory address and variable size from the debug symbols.
There are no more hardcoded maps such as `GameState` or shell script to export the symbols using `nm`.

ricfehr3:
@macabeus Thank again for updating this. I'm encountering a bug but otherwise this is good to go for me. 

When following the README you added,I get this failure on `npm test`:

```
➜  tests_e2e git:(e2e-test) ✗ npm test

> test
> vitest run


 RUN  v4.1.9 /home/rf3/projects/balatro-gba/tests_e2e

 ❯ tests/splash.test.ts (0 test)
 ❯ tests/hand.test.ts (0 test)
 ❯ tests/blind.test.ts (0 test)

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Suites 3 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  tests/blind.test.ts [ tests/blind.test.ts ]
 FAIL  tests/hand.test.ts [ tests/hand.test.ts ]
 FAIL  tests/splash.test.ts [ tests/splash.test.ts ]
RangeError: Offset is outside the bounds of the DataView
 ❯ Cursor.u8 node_modules/@gba-kit/debug-info/src/reader.ts:33:21
 ❯ parseUnit node_modules/@gba-kit/debug-info/src/debug-line.ts:78:33
 ❯ parseDebugLine node_modules/@gba-kit/debug-info/src/debug-line.ts:43:4
 ❯ DebugInfo.fromElf node_modules/@gba-kit/debug-info/src/debug-info.ts:44:30
 ❯ helpers/addresses.ts:12:25
     10|   throw new Error(`ELF not found at ${ELF_PATH} — build it first with: UID=$(id -u) GID=$(id -g) docker compose up`);
     11| }
     12| const debug = DebugInfo.fromElf(new Uint8Array(readFileSync(ELF_PATH)));
       |                         ^
     13|
     14| /**

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/3]⎯


 Test Files  3 failed (3)
      Tests  no tests
   Start at  14:25:32
   Duration  187ms (transform 79ms, setup 0ms, import 0ms, tests 0ms, environment 0ms)

➜  tests_e2e git:(e2e-test) ✗
```

I'm made sure it is seeing the properly as well, Is there another step that the README didn't catch?

ricfehr3 (changes requested):
Requesting changes based on other comment on testing failure.

PR #409, opened 2026-03-29
Increased number of MaxMod audio channels to fix sound effects conflict

MeirGavish:
There was an issue I missed in my previous music commit where I increased the number of music channels where there were not enough sound channels for the music and sound effects when there were many rapid sound effects so some of them were not audible. This was evident in the end of the round where the cards from the hand are discarded and all the discarded cards are returned to the deck.

https://github.com/user-attachments/assets/77bea054-127b-4666-90d6-12b965dd24b3

https://github.com/user-attachments/assets/9c056091-75bd-428a-9a56-e77e12268bce

To fix this I increased the number of channels allocated to MaxMod and also replaced that magic number with a define.


ricfehr3:
Oh nice catch. Going to merge this one. Sounds really good btw!

ricfehr3 (approved):

PR #411, opened 2026-03-29
Fixed Shortcut joker ID for #410

### LINKED ISSUE #410: [BUG] Shortcut not working and its effect moved to The Order
**Describe the bug**
Shortcut Joker's effect doesn't activate, you still need a full straight without skips.
Its effect does activate with The Order, so having it you can make a straight with skips. It still retains its own effect of x3 mult if played hand contains a straight, making it quite OP.

**To Reproduce**
Get the Shortcut Joker, you can give yourself a lot of `STARTING_MONEY` and compile a version where you can get it with a bunch of rerolls. Try playing a straight with skips, it isn't detected as a straight.
Get the joker ID for The Order and set it as the `TEST_JOKER_ID0` and compile a version where you get it in the shop. Obviously the joker IDs have been mixed up so using `SHORTCUT_JOKER_ID` will actually give you The Order.

**Expected behavior**
Each of these jokers should have their own effects and only them.

**Screenshots**

Image
Image
Image

**What version**
Main branch with latest commit [Implement arbitrary-sized joker spritesheets (#285)](https://github.com/GBALATRO/balatro-gba/commit/5cf9e2674d267dacf134839cce70feb0689187f5) which seems to be the PR that caused the issue.

mgba

**Additional context**
The joker IDs have been mixed up, so other jokers may be affected but I haven't tested. Four Fingers is likely affected.

MeirGavish:
closes #410
It seems `SHORTCUT_JOKER_ID` was just not up to date.
I also removed some unused joker IDs.

balatro-gba-11
balatro-gba-12
balatro-gba-13

### DIFF
--- a/include/joker.h
+++ b/include/joker.h
@@ -84,10 +84,8 @@ enum JokerEvent
 #define MAX_JOKER_OBJECTS 32 // The maximum number of joker objects that can be created at once
 
 // Jokers in the game
-#define DEFAULT_JOKER_ID      0
 #define STENCIL_JOKER_ID      15
-#define GREEDY_JOKER_ID       18
-#define SHORTCUT_JOKER_ID     38
+#define SHORTCUT_JOKER_ID     48
 #define BRAINSTORM_JOKER_ID   41
 #define PAREIDOLIA_JOKER_ID   46
 #define FOUR_FINGERS_JOKER_ID 50


ricfehr3 (approved):
Nice, good catch and cleanup. I double checked and these special joker IDs match up now. I'll merge. Thanks!

Issue #335, opened 2025-12-21
Build and host Doxygen docs on merges to main


ricfehr3:
We want to host the docs built with Doxygen on github pages. Here is one option to consider: https://github.com/marketplace/actions/doxygen-and-github-pages

To complete this task:
- [ ] Add a github action workflow to build the documents with Doxygen
- [ ] Create the required [github pages](https://docs.github.com/en/pages/quickstart)

ricfehr3:
Closed with #417

PR #414, opened 2026-03-30
Make doxygen look good!

SuperbMuffin:
The default Doxygen layout can feel a bit dated and uninspired. This setup aims to modernize the look by integrating the Doxygen-Awesome theme by jothepro.

### Key Changes:
- Updated font and layout using the Sidebar-only theme from Doxygen-Awesome.
- Added a brief description for the wiki.
- Included a custom icon for the wiki.
- Adjusted the theme styling slightly to better suit the project’s branding and readability.
### Potential Improvements:
- Adding some of the Doxygen-Awesome JS extensions could enhance interactivity and navigation. (done)
- Including a GitHub corner would provide quick access to the repository and improve visibility. (done)
-Host the docs on a github site and remove the wiki from the repo

### **Before**
20260330_03h33m13s_grim

### **After**
20260330_03h30m05s_grim


ricfehr3:
Yo thanks for another PR. I like how clean this looks. I'll need some time to familiarize myself with what Doxygen Awesome is and put some thought into this. However, right away I see that the recommended install of the css file is [via a git submodule](https://jothepro.github.io/doxygen-awesome-css/index.html#autotoc_md12). If it's added, it should be added as a submodule, but introducing submodules can be a bit of a headache. @MeirGavish I wanted to start the conversation on that. Do you have any thoughts/opinions on submodules? 

Also, I wanted to note this would be related to #335 .

SuperbMuffin:
I felt adding it as a git submodule would add unnecessary complicity, If you guys felt that it would be a better way to do I'm cool with migrating it.I think submodules aren't that bad it just needs to be clear that u need to initialize it after cloning. I wonder if you can initialize it from the Doxyfile, probably not.

I also think that the js "extensions" from doxygen awesome should be added. I'll do that now. 

I also wanted to mention that the MIT license inside doxygen awesome isn't technically required because of the comments at the start of each css file. ( I'm not a lawyer I just asked chatgpt )

SuperbMuffin:
It looks even better now, the color tweaks actually come through. I added the light/dark js extension.

Im going to keep adding js extensions.

SuperbMuffin:
It looks pretty different now:

20260330_16h29m42s_grim

ricfehr3:
Okay, I thought about this a bit. I would really rather not introduce submodules for this, so I'm good to just bring it in as-is. I'm going to go over it more, but as long as it doesn't rely on any external scripts, then I'm good to bring in just a snapshot. 

@SuperbMuffin Can you add a `VERSION` to `doxygen-awesome` that credits the repo, has a link to the repo, and has a version/githash that we are using?

Thanks!

doxygen/doxygen-awesome/VERSION
```diff
@@ -0,0 +1,13 @@
...
+Version: snapshot
+Commit: d52eafe
+
+Imported-by: SuperbMuffin
+Date: 2026-04-02
+
+Notes:
+
+- Imported as a snapshot (no git submodule)
+- No external scripts required
```
ricfehr3:
```suggestion
- There are no dependencies on external resources besides images. Posing no security issue that would require consistent updating.
```

Just wanted to note why it's important it doesn't have external dependencies.

ricfehr3 (approved):
Just a quick suggestion, but otherwise it looks good to merge for me 👍

SuperbMuffin:
Awesome, I added the suggestion, I think it's ready to merge!

ricfehr3:
Great! Thank you for working through the PR and thanks for another contribution! 

~~One additional change I wanted to make, but I will make an issue for instead is moving the `icon.png` to the doxygen directory. However, we also have the image for the example gif in the same directory, so I'm just going to make note of it to clean up.~~

I change my mind, I like it as it is now. Thank you!

PR #415, opened 2026-03-30
Add basic common_ui background gfx functions

ricfehr3:
This PR adds some of the required scaffolding to start taking apart game.c more. This branch is used as the base to the main menu removal from `game.c`: #416

PR #416, opened 2026-03-30
Refactor/main menu

ricfehr3:
~~Blocked on #415~~ 

~~WIP: first screen to remove from `game.c`~~

Ready to go for a review now.


ricfehr3:
I'm going to merge this one tonight. Happy with it now. It needed the clang-format in the new subdirectories as well. Added. 

:stop_sign: Don't merge yet, I want to merge #352

ricfehr3:
@MeirGavish I'm going to merge this one pretty soon. No worries if you don't have time to review, I just didn't want to take the opportunity away if you did.

MeirGavish:
> @MeirGavish I'm going to merge this one pretty soon. No worries if you don't have time to review, I just didn't want to take the opportunity away if you did.

Thanks, I appreciate it! I probably will not be able to review the bigger refactors unfortunately...

MathisMartin31:
@ricfehr3 Will the GameVariables struct you started to fill only be related to engine stuff? I don't remember if we've had this discussion before but I think engine and run variables should live in separate structs.

include/timer.h
```diff
@@ -0,0 +1,45 @@
...
+#define TIMER_H
+
+/**
+ *  @name Timer Definitions
+ *
+ *  @brief Timer constants used throughout this project
+ *
+ *  **NOTE** these constants will
+ *  need to be placed in their own scope outside of this global context during
+ *  this refactor.
```
MeirGavish: Why put them in a single file in the first place then? I say leave them in `game.c` for now and move each value to its relevant state file once it's extracted.

source/game/main_menu.c
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+void game_main_menu_on_exit(GameVariables* vars)
+{
+    // Normally I would just cache these and hide/unhide but I didn't feel like dealing with
+    // defining a layer for it
+    card_destroy(&main_menu_ace->card);
+    card_object_destroy(&main_menu_ace);
+}
```
MeirGavish: Nice, much better place for it

source/game.c
```diff
@@ -126,8 +98,7 @@
 /* This needs to stay a power of 2 and small enough
  * for the lerping to be done before the next hand is drawn.
  */
-#define NUM_SCORE_LERP_STEPS   16
-#define TM_SCORE_LERP_INTERVAL 2
+#define NUM_SCORE_LERP_STEPS 16
```
MeirGavish: I think these 2 variables should be kept together...

include/game_variables.h
```diff
@@ -14,7 +16,8 @@
  */
 typedef struct
 {
-
+    u32 timer;
+    u32 rng_seed;
```
MeirGavish, line 20: You don't use this variable in `game.c`, making the seeding in `game_start()` not affected by the manipulation done during the main menu ruining randomization. I haven't tested this.

Furthermore, I now realize that since it's done in `game_start()`, when the game is reset, the same seed is always used since it's only changed during the main menu. 
It's a little weird because I think it would have made every reset repeat the same run, but so far it didn't seem like it does.

include/game_variables.h
```diff
@@ -14,7 +16,8 @@
  */
 typedef struct
 {
-
+    u32 timer;
```
MeirGavish: The timer should stay signed so it can properly have the `UNDEFINED` value...

MeirGavish: Also, I'm thinking maybe the timer shouldn't be global, and each state should have it's own timer like the splash screen does. That way, they can't corrupt each other by not resetting the timer.

MeirGavish (changes requested):
And I said I won't review PRs and here I am getting sucked back in. I still tried invest a little less time than usual so I may have missed some things.

Either way, this looks like a great refactor that puts things in their proper place.
I added a few comments about a few things.

MeirGavish:
I will also request a review from Copilot just in case I missed something but I will probably not have time to go over it today.

MathisMartin31:
@ricfehr3 I was playing my options menu build and had a weird glitch, so I tested yours since I recovered your changes before applying mine and sure enough, you do have the same issue with your branch: when you draw enough cards, you end up with this glitched one that gives you 34 chips for some reason.

Screenshot_20260423_004744

You always end up with this card, and it always gives you 34 chips, although I don't know if this tells us more about the glitch itself or the rng seed being locked to the same value each time (I've hit a royal flush of hearts one round 1 every time lol)

https://github.com/user-attachments/assets/34ecc5b8-f59c-4366-813b-11c8a473050d

Edit: I just checked to confirm, this does *not* happen on main

ricfehr3:
> @ricfehr3 I was playing my options menu build and had a weird glitch, so I tested yours since I recovered your changes before applying mine and sure enough, you do have the same issue with your branch: when you draw enough cards, you end up with this glitched one that gives you 34 chips for some reason.
> 
> Screenshot_20260423_004744
> 
> You always end up with this card, and it always gives you 34 chips, although I don't know if this tells us more about the glitch itself or the rng seed being locked to the same value each time (I've hit a royal flush of hearts one round 1 every time lol)
> 
> https://github.com/user-attachments/assets/34ecc5b8-f59c-4366-813b-11c8a473050d
> 
> Edit: I just checked to confirm, this does *not* happen on main
> 

Very interesting and very nice catch! Do you happen to have a save state of that glitch? Even if it's just a state with that card in your hand I think it would make it easier to debug. Thanks for alerting me.

MathisMartin31:
> Very interesting and very nice catch! Do you happen to have a save state of that glitch? Even if it's just a state with that card in your hand I think it would make it easier to debug. Thanks for alerting me.

I didn't think of making one because it always happens, but if it helps I can do one just before the bug right now

MathisMartin31:
There you go: all you have to do is play the selected hand and you will draw the glitched card right after that

[ricfehr_gbalatro.ss1.tar.gz](https://github.com/user-attachments/files/26989859/ricfehr_gbalatro.ss1.tar.gz)

If you have you tiles view open set to a width of 4 tiles so the cards are all well formed, you can see that when drawing, the glitched card is the first to appear in your hand, at a position in the array equal to the number of cards played ~(it's less obvious in my previous video but you still kind of see it)~ as seen here:

https://github.com/user-attachments/assets/56efbd54-cdb3-46cf-ae80-620bf7ff0876

MathisMartin31:
And also, it seems to happen after you have drawn many cards so maybe there's one specific card that's glitched and it takes a while to get it.

ricfehr3:
> There you go: all you have to do is play the selected hand and you will draw the glitched card right after that
> 
> [ricfehr_gbalatro.ss1.tar.gz](https://github.com/user-attachments/files/26989859/ricfehr_gbalatro.ss1.tar.gz)
> 
> If you have you tiles view open set to a width of 4 tiles so the cards are all well formed, you can see that when drawing, the glitched card is the first to appear in your hand, at a position in the array equal to the number of cards played ~(it's less obvious in my previous video but you still kind of see it)~ as seen here:
> 2026-04-23.01-31-14.mp4

Huge thank you. I'm going to take a look at this and see what's up. Thank you for the details too!

ricfehr3:
> There you go: all you have to do is play the selected hand and you will draw the glitched card right after that
> 
> [ricfehr_gbalatro.ss1.tar.gz](https://github.com/user-attachments/files/26989859/ricfehr_gbalatro.ss1.tar.gz)
> 
> If you have you tiles view open set to a width of 4 tiles so the cards are all well formed, you can see that when drawing, the glitched card is the first to appear in your hand, at a position in the array equal to the number of cards played ~(it's less obvious in my previous video but you still kind of see it)~ as seen here:
> 2026-04-23.01-31-14.mp4

@MathisMartin31 Unfortunately when loading the save state it says it's for a different version:

```sh
GBA Savestate: Savestate is for a different game
GBA Savestate: Savestate is for a different version of the game
```

Just to check, can you please confirm that you are on git hash `93ee65d`

```sh
python ./scripts/get_hash.py build/balatro-gba.gba
git hash: 93ee65d
```

include/game_variables.h
ricfehr3: 100% Agree, that makes a lot more sense.

include/game_variables.h
ricfehr3: Oops, yeah my bad on that. I'll fix the `rng_seed` variable.

I see what you're saying about the seed not being updated. I think we should make it an issue and solve it in another PR.

include/timer.h
ricfehr3: That's fair, no need for this indirection for most of these. `TM_ZERO` does need to be a common variable. I was thinking of maybe an interface to a timer, but... that's too complex. If that's the case I think it would be fair to just leave it in `game.h`. Let me clean it up and see what's good.

MathisMartin31:
@ricfehr3 I just tried and it says the hash is also 93ee65d, that's weird

If you can't make it work, just do this: discard 5 cards 4 times, then play any number of cards. The first one drawn will be the glitched one.

MathisMartin31:
I don't know why I didn't think of that sooner lol, but I gave myself 40 discards and cycled through the whole deck by discarding everything but the aces. Seems like the cursed card is the ace of spades

Screenshot_20260423_074504

Maybe you did something to it when moving the main menu out of the game.c file ?

MeirGavish:
> I don't know why I didn't think of that sooner lol, but I gave myself 40 discards and cycled through the whole deck by discarding everything but the aces. Seems like the cursed card is the ace of spades
> 
> Screenshot_20260423_074504
> Maybe you did something to it when moving the main menu out of the game.c file ?

Good catch! I tested this and can confirm this happens.

ricfehr3:
@MathisMartin31 @MeirGavish 

Okay, I found the issue. Moving the card destruction to the main_menu exit function was the issue. `game_start()` is always called first, and that creates all the cards. Then on the main menu exit, it deletes it.

To solve this, I moved the `game_start()` function to a state itself with only that as an update function. This makes it so the main_menu state has a chance to clean up. I feel that this makes the most sense. It's better to adhere to the state machine than to move resource creation and destruction out of it's pattern, imo. Otherwise, the card would need to be deleted before calling `game_start` or be moved into `game_start` itself. both of those methods are unmaintainable.

source/game/main_menu.c
```diff
@@ -0,0 +1,118 @@
...
+    main_menu_ace->sprite_object->x = main_menu_ace->sprite_object->tx;
+    main_menu_ace->sprite_object->ty = int2fx(MAIN_MENU_ACE_T_Y);
+    main_menu_ace->sprite_object->y = main_menu_ace->sprite_object->ty;
+    main_menu_ace->sprite_object->tscale = float2fx(0.8f);
+    selection_x = 0;
+}
+
+void game_main_menu_on_update(GameVariables* vars)
+{
+    change_background(BG_MAIN_MENU);
```
MathisMartin31: Do we need to do this call on every frame? I recovered your changes for my own PR, commented out this line, and nothing looks out of place 🤔

source/game/main_menu.c
```diff
@@ -0,0 +1,118 @@
...
+    main_menu_ace = card_object_new(card_new(SPADES, ACE));
+    card_object_set_sprite(main_menu_ace, 0);
+    main_menu_ace->sprite_object->sprite->obj->attr0 |= ATTR0_AFF_DBL;
+    main_menu_ace->sprite_object->tx = int2fx(MAIN_MENU_ACE_T_X);
+    main_menu_ace->sprite_object->x = main_menu_ace->sprite_object->tx;
+    main_menu_ace->sprite_object->ty = int2fx(MAIN_MENU_ACE_T_Y);
+    main_menu_ace->sprite_object->y = main_menu_ace->sprite_object->ty;
+    main_menu_ace->sprite_object->tscale = float2fx(0.8f);
+    selection_x = 0;
```
MathisMartin31, line 61: I think I will remove this line for the options menu review, that way when exiting it you will still have the button id 1 highlighted, and since it is initialized at 0 when declared and that it is equal to 0 when starting a game I don't see any drawbacks to not forcing the value on init.

source/game.c
```diff
@@ -65,8 +64,6 @@
 #define CARD_UNFOCUSED_SEL_Y 15
 #define CARD_FOCUSED_SEL_Y   20
 
-// Timer defs
-#define TM_ZERO                         0
 #define TM_RESET_STATIC_VARS            30
```
MathisMartin31, line 67: Was it intentional to move all these timer defines back here and not leave them in the timer.h file?

ricfehr3: Yes, that was a suggestion of @MeirGavish . Since they are still only used in `game.c`, we will leave them in there for now. We also plan to encapsulate timers into their own states now. This file is for pulling common timer functionality starting with `TM_ZERO`, as it is used in `game.c` and `main_menu.c`.

source/game/main_menu.c
ricfehr3: Doh! Yeah that's wrong, nice catch.

source/game/main_menu.c
ricfehr3: Nice! That is correct design. Encapsulating behavior :nerd_face:

MathisMartin31 (approved):
Nice, approving this!

source/game/common_ui.c
```diff
@@ -1,22 +1,32 @@
...
 {
+    if (id == BG_MAIN_MENU)
+    {
+        if (id != background)
+        {
+            bgCallbacks[id]();
+        }
+    }
     change_background_legacy(id);
+    id = background;
```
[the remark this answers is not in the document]
ricfehr3: Oops, the suggested fix isn't correct but the suggestion is.

ricfehr3:
This one requires #451 now.

source/game/common_ui.c
```diff
@@ -1,22 +1,32 @@
...
+BackgroundRenderCallback bgCallbacks[] = {
     [BG_NONE] = NULL,
     [BG_CARD_SELECTING] = NULL,
     [BG_CARD_PLAYING] = NULL,
     [BG_ROUND_END] = NULL,
     [BG_SHOP] = NULL,
     [BG_BLIND_SELECT] = NULL,
-    [BG_MAIN_MENU] = NULL,
+    [BG_MAIN_MENU] = game_main_menu_change_background,
 };
```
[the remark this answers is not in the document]
MeirGavish: I think it does make sense to make it `static`, doesn't it?

MeirGavish (commented):
Looks good and runs well. Just that one Copilot comment about making the table static.

MeirGavish:
> @MathisMartin31 @MeirGavish
> 
> Okay, I found the issue. Moving the card destruction to the main_menu exit function was the issue. `game_start()` is always called first, and that creates all the cards. Then on the main menu exit, it deletes it.
> 
> To solve this, I moved the `game_start()` function to a state itself with only that as an update function. This makes it so the main_menu state has a chance to clean up. I feel that this makes the most sense. It's better to adhere to the state machine than to move resource creation and destruction out of it's pattern, imo. Otherwise, the card would need to be deleted before calling `game_start` or be moved into `game_start` itself. both of those methods are unmaintainable.

Great! Seems like a good solution.

ricfehr3:
@MeirGavish @MathisMartin31 Once again, thank you for the reviews. I'm stoked to clean up `game.c` finally. I added the `static const`. I'm going to go ahead and merge.

PR #420, opened 2026-04-02
Add build archival script with timestamped snapshots (.gba, .elf, .map) 

umtipako:
- Added a bash script under /scripts/ to archive output files under a timestamped folder.
- CONTRIBUTING.md modified (**Custom Scripts** section)
- .gitignore modified to ignore /saved_builds/ folder


scripts/save_build.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+make || exit 1
```
ricfehr3:
```suggestion
set -e

make
```

I would just use `set -e` here to exit immediately on failure of any command in this script. 
https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin

CONTRIBUTING.md
```diff
@@ -95,6 +95,7 @@ In the repo we use custom scripts located in the [`scripts`](https://github.com/
 - **get_hash.py**: Get git hash from ROM.
 - **generate_font.py**: Generate a font manually.
 - **get_memory_map.sh**: Print the memory map of the pre-allocated pools.
+- **save_build.sh**: Save a timestamped copy of build outputs (.elf, .gba, .map).
```
ricfehr3, line 98: Thanks for the docs update 👍

scripts/save_build.sh
```diff
@@ -0,0 +1,9 @@
...
+timestamp=$(date +%Y%m%d_%H%M%S)
+arg=${1:-"build"}
+dir="saved_builds/${arg}_${timestamp}"
+mkdir -p "$dir"
+cp build/balatro-gba.elf build/balatro-gba.gba build/balatro-gba.map "$dir/"
```
ricfehr3: The basename `balatro-gba` for these files depends on directory name of the project. I'm sure most people will use the default `balatro-gba`, but it's a tonc makefile leftover that the ROM name depends on the directory name. @MeirGavish I'm going to make an issue for this actually.

@UmtiPako In the meantime, could you add an environment variable similar to what we have in [`get_memory_map.sh`](https://github.com/GBALATRO/balatro-gba/blob/12ec71ba2c2612481a67d315f325e393fc8dba7f/scripts/get_memory_map.sh#L5). Sort of like: `GAME_NAME="${GAME_NAME-balatro-gba}"`.

scripts/save_build.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
```
ricfehr3: Could you add a comment explaining that this can only be ran in the repo base directoy.

ricfehr3 (changes requested):
Hey, thanks for the PR. This looks like a good developer tool to add. I have suggestions for changes, nothing too crazy, but otherwise great work! It's pretty straightforward and useful 👍

scripts/save_build.sh
ricfehr3: Captured in #421

umtipako: Alrighty!

scripts/save_build.sh
```diff
@@ -1,11 +1,34 @@
...
 arg=${1:-"build"}
 dir="saved_builds/${arg}_${timestamp}"
+
 mkdir -p "$dir"
-cp build/balatro-gba.elf build/balatro-gba.gba build/balatro-gba.map "$dir/"
-echo "Build saved to $dir"
\ No newline at end of file
+
+failed=0
+  for fe in elf gba map; do
```
ricfehr3: It looks like there's an indentation issue with couple of spaces in front of the code after this line.

scripts/save_build.sh
```diff
@@ -1,11 +1,34 @@
...
+      fi
+  done
+
+  if [ $failed -eq 1 ]; then
+      echo "Some files are missing, partially saved to $dir"
+  else
+      echo "Build files saved to $dir"
+  fi
+
+
```
ricfehr3:
```suggestion
exit $failed
```

I think it would be helpful to also exit with the error value.

umtipako:
> Thanks for the quick fixes! Just a couple more things. Check the comments below. After those are addressed, I think it's good.

You're welcome! I wasn't planning to add stuff at 3am my time but wanted to add this tool quick for further development before i forget. Great project, hope to hang around more!

umtipako:
> Just tested and PR looks all good besides one last change. Can you set the file to be executable with `chmod +x`? I had to do that manually when I tested your branch.

All done. Thanks for catching it 🙏

PR #417, opened 2026-03-30
Doxygen Auto Build and Deploy

SuperbMuffin:
## Add Doxygen Documentation with GitHub Pages Auto-Deploy

This PR adds a GitHub Actions workflow that automatically builds the Doxygen documentation and deploys it to GitHub Pages on every push/merge to `main`.

### What this adds
- `.github/workflows/docs.yml` — builds docs with Doxygen and deploys to the `gh-pages` branch
- PRs will trigger a test build to catch any doc errors before merging
- Merges to `main` will build **and** deploy the live site

### One-time setup required after merging
Please enable GitHub Pages in the repo settings:

**Settings → Pages → Source → branch: `gh-pages`, folder: `/ (root)` → Save**

The docs will then be live at https://gbalatro.github.io/balatro-gba/ and will auto-update on every merge to main.

> No secrets or further configuration needed — `GITHUB_TOKEN` is provided automatically by GitHub Actions.

*this is related to #335*

PR #422, opened 2026-04-02
Update CONTRIBUTING guide to inform of hosted docs

ricfehr3:
Now that we host the docs built by Doxygen on github as of #417 we should let developers know they can find it there: https://gbalatro.github.io/balatro-gba


MeirGavish:
Great! Merging

PR #424, opened 2026-04-02
feat: implement 7 boss blind effects (The Needle, Water, Hook, Psychic, Tooth, Eye, Ox)

OutBlade:
## Summary

Implements 7 boss blind effects as a new self-contained module (), hooking into  at three call-sites. Partially addresses #354.

### Boss blinds

| Blind | Effect |
|-------|--------|
| **The Needle** | Only 1 hand allowed per round |
| **The Water** | Start with 0 discards |
| **The Hook** | After scoring, 2 held cards are discarded |
| **The Psychic** | Must play exactly 5 cards |
| **The Tooth** | Lose $1 for each card played |
| **The Eye** | Each hand type can only be played once per round |
| **The Ox** | Playing any hand drains money to $0 |

### Changes to 

- **** — calls  after cashout has reset counters; refreshes HUD.
- **** — calls  to gray-out the Play button for The Psychic (must select 5) and The Eye (no repeated hand types).
- **** — after the last played card exits the screen, applies The Eye registration, The Tooth penalty, The Ox drain, and The Hook discard.

### Tests

 — standalone host-side suite (gcc, no GBA deps) covering every public API function. A local `blind.h` stub in the test directory provides the `BlindType` enum without pulling in GBA-specific headers; the Makefile include order ensures it shadows the real header during test compilation.

### Depends on

This PR is ready to be rebased on top of #352. The `BossBlindId` enum has been removed in favour of the `BlindType` enum from #352, and all call-sites in `game.c` now use `current_blind` directly (checking `current_blind > BLIND_TYPE_BIG` rather than `== BLIND_TYPE_BOSS`).

### Known limitations / TODOs

- **The Hook** currently removes the *topmost* 2 held cards; proper randomisation + HAND_DISCARD animation is a follow-up.
- **The Manacle** (hand size −1) needs a  variable to be undone each round — excluded from this PR.

### DIFF
--- a/include/boss_blind_effects.h
+++ b/include/boss_blind_effects.h
@@ -3,17 +3,19 @@
  *
  * @brief Boss blind effect definitions and API.
  *
- * Each ante (1–7) has a distinct boss blind with a unique gameplay modifier.
- * Antes beyond 7 cycle back to the first boss blind (The Needle).
+ * The active boss blind is identified by the game''s current_blind variable
+ * (enum BlindType from blind.h). All functions in this module take a
+ * BlindType directly; callers pass current_blind instead of calling the
+ * removed boss_blind_get_for_ante() helper.
  *
  * Implemented boss blinds:
- *   Ante 1 – The Needle  : Only 1 hand allowed per round.
- *   Ante 2 – The Water   : Start the round with 0 discards.
- *   Ante 3 – The Hook    : After each hand scored, 2 held cards are discarded.
- *   Ante 4 – The Psychic : Must play exactly 5 cards.
- *   Ante 5 – The Tooth   : Lose $1 for each card played.
- *   Ante 6 – The Eye     : Each hand type can only be played once per round.
- *   Ante 7 – The Ox      : Playing any hand sets money to $0.
+ *   The Needle  : Only 1 hand allowed per round.
+ *   The Water   : Start the round with 0 discards.
+ *   The Hook    : After each hand scored, 2 held cards are discarded.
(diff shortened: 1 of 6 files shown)


MonteCrysto31:
> Boss blind assignment is deterministic (); per-run RNG rotation is a follow-up.

This should not be the case after #352 is merged, but to be confirmed.

Also, please note that that PR changes a _lot_ of things on the Blinds' end, in particular you will need to use the `BlindType` enum and the game.c variable `current_blind` instead of the `BossBlindId` you added.

OutBlade:
Thank you for the heads-up, @MonteCrysto31.

I will update the PR description to remove the RNG rotation item from the known limitations once #352 is confirmed merged, since that PR handles it directly.

Regarding the interface changes: I will replace the `BossBlindId` enum I introduced with the `BlindType` enum from #352 and switch to reading `current_blind` from `game.c` in place of any local tracking I currently do. I will rebase this branch on top of #352 once that PR lands so the migration is clean and there are no duplicate definitions.

If there are any additional API details in #352 I should be aware of before rebasing, let me know and I will account for them upfront.

MathisMartin31:
@OutBlade I took another look and tried to compile your fork but it seems there are a lot of compile errors. Did you try to build this, and if not is this still a WIP? If it's the latter, I think you should make your PR a draft until it's ready for review.

Mega-Evolved-Ziquinho:
> [Grays-out] the Play button for The Psychic (must select 5) and The Eye (no repeated hand types).

Balatro will actually let you play cards even with these effects, they just won't score

PR #430, opened 2026-04-05
Add MIT License to the project

SuperbMuffin:
### Reasoning

Adding a permissive license to make usage, modification, and distribution explicit.

MIT keeps things simple and unrestrictive.

### Other Options:
 -  Unlicense
 -  Mozilla Public License 2.0
 -  GNU General Public License

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 GBALATRO
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


MeirGavish:
So as said in https://github.com/GBALATRO/balatro-gba/pull/202, licensing is problematic since we don't have the rights to Balatro so we can't explicitly allow people to use it. I do want to say that if the project is licensed, I don't want the license to allow **selling** while I would be fine with distribution for free but I want it to require credit attribution to this repository and its contributors.

SuperbMuffin:
I didn't see #202 before making this pr so ill just close this.

PR #432, opened 2026-04-06
WIP: Refactor/blind select screen

ricfehr3:
Draft PR: Relies on #416


include/game.h
```diff
@@ -172,9 +176,20 @@ void set_game_speed(int new_game_speed);
...
 void change_background_legacy(enum BackgroundId id);
+void reset_background(void);
+
+void display_round(void);
+
+void increment_blind(enum BlindState increment_reason);
+
+void reset_top_left_panel_bottom_row(void);
+
+GameVariables* get_game_vars(void);
```
ricfehr3, line 193: This snuck in, because it is genuinely way easier to work with. I'm thinking through how I feel about the way the game variable struct is accessed (see convo from #338).

MathisMartin31: @ricfehr3 How about making `game_vars` a global var with extern? It's used at so many points that it's starting to make sense to me :thinking:

ricfehr3: I think that is probably the way to go now. It is the most pragmatic design. The typical design restrictions of that approach aren't valid here. Plus, we are going to keep that variable as concise as possible as screens are extracted out. 

I think this is a justified reason to use an extern. It is a static, deterministic hunk of data that most every system uses. Let's avoid all the redirection and just use it like a knife, carefully, haha.

MeirGavish: I think that's fine too, just put the the extern declaration in the header so it appears only in one place.

ricfehr3:
closing in favor of new PR #461

PR #433, opened 2026-04-07
[Example] add simple asset example

ricfehr3:
**Edit**: 4/7/26, It's a lot closer (but still very untested) than the first post below.

Continuing [here](https://github.com/GBALATRO/balatro-gba/pull/433#issuecomment-4203975256)

---

The purpose of this PR is to show a minimal working example of having a separate project with joker assets. It's not close to the final design. This is a simple example of how an asset system would work. This is a rough example so bear with me, but this does compile and run.

Here, the `joker_effects.c` file and the joker graphics have been moved to a separate subdirectory: `assets`. 

You can see that now the `joker.c` logic has no dependency on the joker effects or graphics. The X-Macro system for the graphics works well with this, as the relevant pngs, grit files, and sprite maps are self-contained to the assets directory. The static library `libjoker_effects.a` now contains the art and effects scoring code. So someone could take that asset dir, modify the art and effects to make a new `libjoker_effects.a`, and that could swapped out, without any changes to the actual `joker.c` logic. 

Defining the interfaces and roles of the engine and asset code is a work in progress, but I hope this illustrates how we can divide and maintain these roles.

I will be leaving this PR open for design discussion. I think it's easier to review the code in action than to write out the design, then it can be refined into a new example.

---

**Note**: To compile this you must first `cd assets` then `make` in that directory. It will make a library with the joker effects and graphics for the whole game: `libjoker_effects.a`. Then build the ROM as normal with `make` in the root of the project.

Related #431 

---

**TODO**:
- [x] I plan to remove the `extern`'s and add a well defined interface. This is WIP.
- [ ] Move "JOKER_ID" definitions to asset files, and remove extern. (look into using some other method that isn't hard coded)
- [ ] Clean up example asset directories. Making the example cleaner.


MeirGavish:
So it's more the game building by linking the assets in rather than an implementation in the assets that uses the engine as a library.

ricfehr3:
> So it's more the game building by linking the assets in rather than an implementation in the assets that uses the engine as a library.

@MeirGavish You got it. I just pushed an updated example that's closer to what I'm thinking. There's a lot going on here, but let me try to add some context to help you and others navigate (I'll be doing a better write-up once it's more concrete). 

What's going on here is a little bit of linker magic. That's why there's new copies of all the Makefile helpers from DevkitPro, janky modification to get the linker script. Basically, The joker registry has been modified to live in a specially labelled section of the cart: `jokers`. This is now in the [`joker.h`](https://github.com/GBALATRO/balatro-gba/pull/433/changes#diff-d0be7e0fdc28c2d1762804fcc3916d5854b420e3a866990854fcf523848f0851R184) file. The linker script, [`gba_cart.ld`](https://github.com/GBALATRO/balatro-gba/pull/433/changes#diff-30cb687bf1117903332ee174b64664b7e65337636d89151f7d525382ec9d7543R116) has a new section concatenates all the pointers in linked file `.joker` sections into accessible data via externs `__joker_start` and `__joker_end`. since they will all be `JokerInfo` structs, you can operate on the pointer data as a regular array, as you would with the usual joker registry. I tested this by splitting jokers between libraries, i.e. mod projects or whatever.

I also made the sprite map info part of the `JokerInfo` struct, as well as pointers to their graphics data. That removed dependencies in `joker.c`. I think we should consider merging this change regardless. However, there's lots of room to clean this up. for example: finding a wildcard of library names to certain assets so you can auto load everything on a build would be nice. Nothing is really too user friendly at this point, but also, a working example of modular jokers.

---

A clear overview of the planned workflow:

1. Copy an assets example directory and make your own project.
2. Set an environment variable to point to this project's headers.
3. Modify the game.
4. Run `make` and build a static library with an easily found name: `libjoker_<modname>.a` or something.
5. Copy the file into a directory that this game will link from. 
6. The game will link all the jokers into a big joker_registry.

---

**I have not tested this thoroughly, it's just a proof-of concept**. This design is open for debate. I want to leave it hear and explain it the best I can while I get back to refactoring with my time, though I may want to get this one in sooner rather than later. I hope it seems like a rational change.

I don't plan on removing any feature or asset that we have, rather just make them removable instead of with hard-coded dependencies.

assets_2/cool_mod.c
```diff
@@ -0,0 +1,1413 @@
...
+static u32 sinful_joker_effect(
+    Card* scored_card,
+    u8 sinful_suit,
+    enum JokerEvent joker_event,
+    JokerEffect** joker_effect
+);
+
+REGISTER_JOKER_EFFECT_FUNC(joker_effect_noop)
+
+REGISTER_JOKER(afa_, COMMON_JOKER, 2, 20, 4, joker_effect_noop)
```
ricfehr3, line 45: Symbols can't crash. Need to figure out a way to handle that better, but that's why you see this weird name, and also why in this `assets_2` directory there is only `joker_gfx20.grit/png` files, as they are a different value than the rest in `assets`

include/joker.h
```diff
@@ -177,4 +181,28 @@ bool joker_object_score(
...
+extern const JokerInfo *__jokers_end[];
+
+static inline const JokerInfo *joker_get(int id) {
+    return __jokers_start[id];
+}
+
+static inline int joker_count(void) {
+    return __jokers_end - __jokers_start;
+}
+
```
ricfehr3, line 207: This is the way the jokers are registered.

assets/joker_effects.c
```diff
@@ -90,105 +91,59 @@ REGISTER_JOKER_EFFECT_FUNC(hack_joker_effect)
...
-        return NULL;
-    }
-    return &joker_registry[joker_id];
-}
-
-size_t get_joker_registry_size(void)
-{
-    return joker_registry_size;
-}
+REGISTER_JOKER(default_, COMMON_JOKER, 2, 0, 0, default_joker_effect)
```
ricfehr3, line 94: here is an example of how the `REGISTER_JOKER` macro looks in action. Pretty similar to the current registry.

assets/Makefile
```diff
@@ -0,0 +1,30 @@
...
+SFILES := $(patsubst graphics/%.png,$(BUILD)/%.s,$(IMAGES))
+OFILES := $(patsubst graphics/%.png,$(BUILD)/%.o,$(IMAGES))
+
+all: lib/libjoker_effects.a
+
+lib/libjoker_effects.a: $(BUILD)/joker_effects.o $(OFILES)
+	mkdir -p lib && $(PREFIX)ar rcs $@ $^
+
+$(BUILD)/joker_effects.o: joker_effects.c $(SFILES) | $(BUILD)
+	$(PREFIX)gcc -c $< -o $@ -I$(DEVKITPRO)/libgba/include -I$(DEVKITPRO)/libtonc/include -I../include -I../include -I./include -I./build
```
ricfehr3, line 15: Here I manually link against the headers, but this could just point to the headers directory via an environment variable and be built "out-of-tree."

gba_cart.ld
```diff
@@ -0,0 +1,323 @@
...
+		*(.gnu.linkonce.r*)
+		SORT(CONSTRUCTORS)
+		. = ALIGN(4);   /* REQUIRED. LD is flaky without it. */
+	} >rom = 0xff
+    .jokers :
+    {
+        __jokers_start = .;
+        KEEP(*(.jokers))
+        __jokers_end = .;
+    } >rom
```
ricfehr3, line 121: This file, and the other linker/Makefile helper files: `gba.specs` and `gba_rules` are from devkitpro. It's messy, but it's all here really for this one modification, giving the linker an instruction to concatenate all `.joker` sections while linking between `__jokers_start` and `__jokers_end`.

gba.specs
```diff
@@ -0,0 +1,7 @@
+%include <sync-none.specs>
+
+*link:
++ --gc-sections --no-warn-rwx-segments
```
ricfehr3, line 4: Another mod here, removed the `-T gba_cart.ld` so it could be moved to the [Makefile](https://github.com/GBALATRO/balatro-gba/pull/433/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R59), where it worked

Makefile
```diff
@@ -55,20 +56,23 @@ CFLAGS	+=	$(INCLUDE)
...
 
 ASFLAGS	:=	-g $(ARCH)
-LDFLAGS	=	-g $(ARCH) -Wl,-Map,$(notdir $*.map),--undefined=balatro_version
+LDFLAGS	=	-g $(ARCH) -Wl,-Map,$(notdir $*.map),--undefined=balatro_version -T$(ROOT_DIR)gba_cart.ld
 
 #---------------------------------------------------------------------------------
 # any extra libraries we wish to link with the project
 #---------------------------------------------------------------------------------
-LIBS	:= -lmm -ltonc
+LIBS	:= -lmm -ltonc -Wl,--whole-archive -lcool_mod -ljoker_effects -Wl,--no-whole-archive
```
ricfehr3, line 64: Manually linking the assets here.

From the `ld` [manual](https://man7.org/linux/man-pages/man1/ld.1.html):

```
       --whole-archive
           For each archive mentioned on the command line after the
           --whole-archive option, include every object file in the
           archive in the link, rather than searching the archive for the
           required object files.  This is normally used to turn an
           archive file into a shared library, forcing every object to be
           included in the resulting shared library.  This option may be
           used more than once.

           Two notes when using this option from gcc: First, gcc doesn't
           know about this option, so you have to use -Wl,-whole-archive.
           Second, don't forget to use -Wl,-no-whole-archive after your
           list of archives, because gcc will add its own list of
           archives to your link and you may not want this flag to affect
           those as well.
```

This option is needed to do the linker trick. The compiler doesn't understand it needs `.joker` sections of data, so you need to use this option to force it to include everything in the static library.

PR #435, opened 2026-04-07
Apply git hash correctly to rom when built with docker

### LINKED ISSUE #249: Git error when building with docker locally on Windows
I finally tried building the rom locally with the Docker instructions on my Windows machine and while the build worked I got the following error
```
balatro-1  | fatal: detected dubious ownership in repository at '/balatro-gba'
balatro-1  | To add an exception for this directory, call:
balatro-1  | 
balatro-1  |    git config --global --add safe.directory /balatro-gba 
```

Which means the operation of adding the git hash to the rom fails.
I confirmed this by running
```
python ./scripts/build_bug_report.py ./build/balatro-gba.gba
```
and got the output
```
git hash: undef-dirty
```

I tried fixing it by updating the command at `docker-compose.yml`
```
    command: ["sh", "-c", "git config --global --add safe.directory /balatro-gba && make -j$(nproc)"]
```

But then received the following error
```
balatro-1  | error: could not lock config file //.gitconfig: Permission denied
```

SuperbMuffin:
this fixes #249

### DIFF
--- a/Dockerfile.gbalatro
+++ b/Dockerfile.gbalatro
@@ -2,5 +2,6 @@ FROM devkitpro/devkitarm:20251117
 
 ENV DEBIAN_FRONTEND=noninteractive
 
+RUN git config --global --add safe.directory '*'
 RUN apt-get update && apt-get install -y \
     python3-pil
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,4 +8,4 @@ services:
     volumes:
       - ./:/balatro-gba
     user: "${UID-1000}:${GID-1000}"
-    command: ["sh", "-c", "git config --global --add safe.directory /balatro-gba && make -j$(nproc)"]
+    command: ["sh", "-c", "make -j$(nproc)"]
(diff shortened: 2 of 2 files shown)


ricfehr3:
Thanks for the PR! I'll test it when I get a chance so we can merge it.

ricfehr3 (changes requested):
Hey. This doesn't seem to work on my system. It runs into an error, `fatal: not in a git directory`. This makes sense as `git config` tries to apply a config file to the current directory. When you try with `--global` you run into a permission issue, and when you try to `cd /balatro-gba` first, it still doesn't work. I'm guessing that's because everything but the `build` directory in owned by `root` in the container, as it should be. 

One approach may be to modify the Dockerfile for the user to have a global git config with the safe directory. But there's a few more ways it could be done.

SuperbMuffin:
I moved the safe.directory Git config into the Dockerfile because it was causing issues on other machines with the container environment. Let me know if you run into any issues.

Issue #421, opened 2026-04-02
ROM name depends on repo directory name


ricfehr3:
Currently, our Makefile names the ROM based on the directory of the repo (the directory the Makefile is in). This means that if someone checks out the repo and renames the directory, then instructions and scripts that specifically reference `balatro-gba.gba` are no longer correct.

The Makefile needs to be updated to always name the ROM: `balatro-gba.gba`.

MalachiBurnett:
I have fixed this issue in pull request #437

SuperbMuffin:
Wasn't this resolved by #512? We agreed to keep the flexibility of the dynamic rom naming :)

ricfehr3:
@SuperbMuffin Nice catch. You're correct, I'll close this. Thanks!

PR #437, opened 2026-04-12
Updated Makefile to hardcode the name of the output rom

MalachiBurnett:
Fixed issue #421:

Currently, our Makefile names the ROM based on the directory of the repo (the directory the Makefile is in). This means that if someone checks out the repo and renames the directory, then instructions and scripts that specifically reference balatro-gba.gba are no longer correct.

The Makefile needs to be updated to always name the ROM: balatro-gba.gba

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ LIBTONC := $(DEVKITPRO)/libtonc
 # the makefile is found
 #
 #---------------------------------------------------------------------------------
-TARGET         := $(notdir $(CURDIR))
+TARGET         := balatro-gba
 BUILD          := build
 SOURCES	       := source source/game
 INCLUDES       := include include/game


SuperbMuffin:
Hey so my two cents on this,

I strongly believe we should keep the name depending on the directory rather than hardcoding it. Our organization name is GBALATRO but our repo is called balatro-gba, and our rebrand has been inconsistent so far. Hardcoding the name to balatro-gba now would actually cement the old name rather than help, we'd be locking in the thing we're trying to move away from.

Beyond the rebrand question, I think the dynamic behavior is useful. Contributors who've already cloned into a GBAlatro/ directory would automatically get a ROM named GBAlatro.gba. The current approach already handles the rebrand organically. I think we should be updating the README and scripts not the MAKEFILE.

Finally, hardcoding introduces a new maintenance burden: every future rename or fork now requires a Makefile edit. The directory-based approach is self-maintaining. Hardcoding breaks that for no real gain.

MonteCrysto31:
Maybe the easiest solution to this is to edit the scripts depending on the ROM being named Balatro-gba so that they also take the name of the base folder into account?

I have to admit that the ROM being named after your base folder has been kind of useful when selecting the file to load in the "recent roms" tab in mgba. It's a minor thing I know but I think some people could see other QoL uses for this.

ricfehr3:
Here's my two cents. The purpose of renaming the ROM was to give our documentation and CI/developer scripts a consistent name. I think that fundamentally the workflow for using the scripts should first be addressed since it's driving the decision. I think the clear fix for that dependency is make the ROM an argument to the scripts. That would make sense for working with any number of ROM names the same way, every time. Especially useful for #420 and, selfishly, my workflow of using multiple git trees.

I don't agree or disagree with the ROM being dynamically named or not. I see pros and cons for both ways. The original issue was that the scripts could be simplified and universal if the ROM name was the same. However, I think it is a better design to just pass the ROM to the scripts. 

We can keep the ROM name dynamic, but we have to fix those scripts. I'm not concerned about the CI scripts as much, those are far more rigid in their usecase. What do y'all think?

ricfehr3:
Closing this and made an issue for my previous comment: #504

PR #439, opened 2026-04-16
Fixed blind sprite flip issue by removing affine attribute

### LINKED ISSUE #438: [BUG] Blind sprite is sometimes flipped on round end screen
**Describe the bug**
Sometimes the beaten blind sprite in the round end cash out menu is flipped horizontally.RT

**To Reproduce**
This doesn't always happen but it seems to reproduce quite reliably when you have at least 4 jokers.
When you beat a round you will see the flipped sprite.

**Expected behavior**
The sprite should maintain an unflipped orientation.

**Screenshots**
Image

**What version**
Main branch with last commit [Implement boss blind graphics](https://github.com/GBALATRO/balatro-gba/commit/db27962623e592c950f782e0f386ad5d03de86f7) (#352)

mgba

**Additional context**
This was introduced in #352.
This seems to happen when the round begins and the round end token is created. For some reason it's created with an active horizontal flip bit in `attr1`. My guesses are either an illegal write/overflow, or a leftover previous memory.
See the discussion in the PR for more information.

MeirGavish:
fixes #438
So it seems I found the issue. The actual write that changed the value was in `sprite_new()` in this line.
```
if (a0 & ATTR0_AFF)
{
   ...
    obj_aff_identity(&obj_aff_buffer[aff_index]);
}
```

When an affine sprite is created it's transform is set to the identity. Since the affine sprite memory is interleaved with the regular sprite memory, this overwrote the the flip bit, I suppose this was only sometimes done because it was affected by the number of affine joker and card sprites.
This code was reached because the blind sprite was set to be affine, and I don't see a reason why it should be and @MathisMartin31 it doesn't seem like you intended it (https://github.com/GBALATRO/balatro-gba/pull/352/changes/379063915f83caf494bdd689ed0bbc0aeb94d5ab) so the fix was just to remove the affine attribute.

### DIFF
--- a/source/blind.c
+++ b/source/blind.c
@@ -231,7 +231,7 @@ Sprite* blind_token_new(enum BlindType type, int x, int y, int layer)
     apply_blind_tiles(type, layer);
 
     Sprite* sprite = sprite_new(
-        ATTR0_SQUARE | ATTR0_4BPP | ATTR0_AFF,
+        ATTR0_SQUARE | ATTR0_4BPP,
         ATTR1_SIZE_32x32,
         get_layer_tile_index(layer),
         get_blind_pb(type),


MathisMartin31:
Damn, good job finding this! I don't know you ever figured this out but GGs! ❤️

MathisMartin31 (approved):

MathisMartin31:
But that also begs the question of maybe we should find a way to clean that memory later, in case we do need affine sprites? Definitely sounds like a bug waiting to come back and bite us.

MeirGavish:
> But that also begs the question of maybe we should find a way to clean that memory later, in case we do need affine sprites? Definitely sounds like a bug waiting to come back and bite us.

That's a good point, I'll merge this to get the bug fixed in the nightly build and open another issue to further investigate if a deeper change is needed.

Issue #440, opened 2026-04-16
Remove intermediate sprite buffer and write directly to OAM instead


MeirGavish:
Right now we have an intermediate buffer `obj_buffer` and `obj_aff_buffer` in work RAM (seemingly IWRAM) where we manipulate all the sprite data for the game and in `sprite_draw()` we copy the entire buffer of `MAX_SPRITES` (128) objects every frame.
I'm not sure what is the reason for this, it seems excessive and to result in an unnecessary performance cost.
I think removing this intermediate buffer and writing directly to OAM will improve performance and may resolve #427.
See example PR #441.

ricfehr3:
***EDIT***: Oops, I should have read the corresponding PR too. I repeated a lot of info here, but I'll leave it up and continue the convo in the PR.

---

Yo, I just went down a rabbit hole with this one. I don't think it will work, but It's pretty interesting as to why, IMO.

My first observation was the pattern of having a separate buffer that is memcpy'd into the oam memory is seen in the tonc examples, as we have it: [check here](https://github.com/search?q=repo%3Agbadev-org%2Flibtonc-examples%20oam_copy&type=code). I figured there was a reason, but I wasn't satisfied with it "just being that way". 

It was actually easy to swap out the buffer and directly write to the oam, just change the pointers, and remove the copy commands:

```c
...

//OBJ_ATTR obj_buffer[MAX_SPRITES];
//OBJ_AFFINE* obj_aff_buffer = (OBJ_AFFINE*)obj_buffer;
OBJ_ATTR* obj_buffer = oam_mem;
OBJ_AFFINE* obj_aff_buffer = (OBJ_AFFINE*)oam_mem;

...

void sprite_draw()
{
    //obj_aff_copy(obj_aff_mem, obj_aff_buffer, MAX_AFFINES);
    //oam_copy(oam_mem, obj_buffer, MAX_SPRITES);
}

...
```

The game runs (kinda...) normally with just that. I then profiled how much time it takes to do those copy functions (same method from #380). It was about 3700 cycles. I honestly don't know how much of an improvement that is, but it's a larger number than the alternative 0 cycles (but I don't think it's not much time overall). However, further investigating this is irrelevant as: **The buffer copy is required to sync the sprites to the current frame**. Playing normally shows this pretty well, look at the card sprites appearing in the top left corner:

[...]

PR #441, opened 2026-04-16
[Example] Direct write to OAM rather than using intermediate buffer

MeirGavish:
For now this is just a quick and dirty demo for #440 as an initial draft of what it would look like with the change.
It has some issues - namely when playing a hand there's a flickering sprite in the top left corner.
image
image

### DIFF
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -11,8 +11,8 @@
 #include <tonc.h>
 #include <tonc_oam.h>
 
-OBJ_ATTR obj_buffer[MAX_SPRITES];
-OBJ_AFFINE* obj_aff_buffer = (OBJ_AFFINE*)obj_buffer;
+OBJ_ATTR* obj_buffer = obj_mem;
+OBJ_AFFINE* obj_aff_buffer = obj_aff_mem;
 
 static Sprite* free_sprites[MAX_SPRITES] = {NULL};
 static bool free_affines[MAX_AFFINES] = {false};
@@ -142,8 +142,8 @@ void sprite_init()
 
 void sprite_draw()
 {
-    obj_aff_copy(obj_aff_mem, obj_aff_buffer, MAX_AFFINES);
-    oam_copy(oam_mem, obj_buffer, MAX_SPRITES);
+    // obj_aff_copy(obj_aff_mem, obj_aff_buffer, MAX_AFFINES);
+    // oam_copy(oam_mem, obj_buffer, MAX_SPRITES);
 }
 
 int sprite_get_pb(const Sprite* sprite)


MonteCrysto31:
It also happens when drawing Jokers in the shop, part of one of them appears in the same spot as the cards in your screenshots it's kind of weird. All I can think of right now is that it seems to happen only on the frame that sprite is created.
I'll have to search what happens to that memory when we write to it. Maybe @ricfehr3 knows some GBA dark magic that could help here?

MonteCrysto31:
Having the tiles memory view opened next to the game doesn't show any weird behaviour, the tiles are copied to the right spot in memory right away so it doesn't look like anything wonky is happening in the memory.

When looking at some code handling card sprites I can see we create the sprite and copy the tiles to the memory, and then apply the position:

```
// Set the sprite for the card object
card_object_set_sprite(hand[i], i);
sprite_position(
    card_object_get_sprite(hand[i]),
    fx2int(hand[i]->sprite_object->x),
    fx2int(hand[i]->sprite_object->y)
);
```

Could it be that since we're writing directly to the memory, that memory had started to get drawn to the wrong position before we set it to the right one, and got a sync issue because we're not copying the whole buffer over in one go at one point when it's not being drawn?

It would also explain why the flickering tiles are always drawn at the same position, and how they are duplicated instead of misplaced (on the first screenshot, the tiles clearly belong to the ace of diamonds, but that card is not missing tile in your hand).

Sounds far-fetched but that's all I can think of.

ricfehr3:
Whoops, I responded in the issue without looking at this PR. I repeated a lot of info you guys already had here, my bad. I'll reiterate the important parts here.

**TLDR;** I don't think it will work.

**Why?**

My first observation was the pattern of having a separate buffer that is memcpy'd into the oam memory is seen in the tonc examples, as we have it: [check here](https://github.com/search?q=repo%3Agbadev-org%2Flibtonc-examples%20oam_copy&type=code).

I initially had my hunches, but I can confirm this method is used to prevent the screen tearing we are seeing in this PR: https://gbadev.net/tonc/regobj.html?highlight=OAM#ssec-oam-buffer

MonteCrysto31:
Well, if it's in tonc's docs it can't be hurting performance *that* much to do it, is there a way of measuring the impact?

ricfehr3:
> Well, if it's in tonc's docs it can't be hurting performance _that_ much to do it, is there a way of measuring the impact?

I profiled how much time it takes to do those copy functions (same method from https://github.com/GBALATRO/balatro-gba/pull/380). It was about 3700 cycles. I honestly don't know how much of an improvement that is, but it's a larger number than the alternative 0 cycles (but I don't think it's not much time overall).

MonteCrysto31:
> > Well, if it's in tonc's docs it can't be hurting performance _that_ much to do it, is there a way of measuring the impact?
> 
> I profiled how much time it takes to do those copy functions (same method from #380). It was about 3700 cycles. I honestly don't know how much of an improvement that is, but it's a larger number than the alternative 0 cycles (but I don't think it's not much time overall).

A quick search tells me the GBA has ~281k cycles per frame. I don't know how accurate that number is but it being in this kind of ballpark means 3700 cycles is something we can clearly afford

MathisMartin31:
@MeirGavish @ricfehr3 Should we still try and investigate ways to do this, of do we scrap it since gbadev says it's kind of a lost cause?

MeirGavish:
> @MeirGavish @ricfehr3 Should we still try and investigate ways to do this, of do we scrap it since gbadev says it's kind of a lost cause?

Sorry, didn't get a chance to respond, wanted to contemplate it and understand it a bit more but generally I'm in the direction of scrapping it either way.

MathisMartin31:
> Sorry, didn't get a chance to respond, wanted to contemplate it and understand it a bit more but generally I'm in the direction of scrapping it either way.

Don't worry no rush, I was just going through our PR backlog see if we can flush anything ^^

ricfehr3:
I think we should scrap it unfortunately. It's a good idea but it seems like a lost cause.

Issue #427, opened 2026-04-03
[BUG] Card Sprite Tear Flicker on Draw


MeirGavish:
**Describe the bug**
Sometimes when drawing cards (e.g. when discarding) the card sprites tear for a single frame.

**To Reproduce**
This is a little difficult to reproduce because it doesn't always happens so I'll do my best to provide a relatively reliable way.
This is most likely a performance issue and it seems to happen mostly when you have a lot of jokers and there are many sprites on the screen.
I'm adding an mGBA save state where it seems to happen quite frequently.
[sprite_tearing_jokers_v1.1.zip](https://github.com/user-attachments/files/26458218/sprite_tearing_jokers_v1.1.zip)
When running it, select 5 cards and discard them, use mGBA's ctrl+P to pause and ctrl+N to run the emulation frame by frame. The flicker frame will be the one where a new card is drawn, it usually happens in the last card drawn.
I suspect it might be with cards that need to be sorted to the middle of the hand but I didn't confirm this.

**Expected behavior**
There shouldn't be a flickering tear in the card sprites.

**Screenshots**
<img width="480" height="320" alt="Image" src="https://github.com/user-attachments/assets/61eb6564-d233-4436-bed1-a1c2b6ad8c7c" />

<img width="480" height="320" alt="Image" src="https://github.com/user-attachments/assets/923bae31-7596-463d-8627-cf99837cf30d" />

**What version**
Main branch with latest commit 2c9264b

**What Emulator or Platform were you playing on**
mGBA

**Additional context**
I'm wondering if maybe merging #268 to improve card sorting performance will solve the issue. Edit: [Apparently not](https://github.com/GBALATRO/balatro-gba/pull/268#issuecomment-4189801788)
Maybe there are other performance optimizations that can be done to completely avoid this issue such as always preloading the sprite for the next card to be drawn.

MeirGavish:
I can confirm this still happens on the main branch as of commit [Scripts Take Rom as Arg (#512)](https://github.com/GBALATRO/balatro-gba/commit/35c7a29cc5deb28d4ed67094cbff89190dc02f47).
It's subtle but still not fixed unfortunately.

<img width="480" height="320" alt="Image" src="https://github.com/user-attachments/assets/22bf8d77-73be-44fb-8a93-15c627e84877" />

MeirGavish:
[...]

PR #444, opened 2026-04-18
Optimize Damping Coefficient

### LINKED ISSUE #373: Optimize division for the damping coefficient in sprite_object_update
For the card springiness, there's a damping coefficient 7/10 used for the velocity in `sprite_object_update()`
This happens on every frame and we can optimize it by changing the division to a power of 2 so the compiler optimizes it to a shift and doesn't call a division subroutine. It's best to not change the stray from the actual value too much so maybe 90/128 = 0.703125 would be good.
Also this should be a defined constant and not a magic number but it needs to be done carefully since these are fixed point operations, the multiplication has to be done before the division.

 _Originally posted by @MeirGavish in [#372](https://github.com/GBALATRO/balatro-gba/issues/372#issuecomment-3799543054)_

SuperbMuffin:
fix #373 

I replaced the division with a shift using the 179/256 method suggested by @MeirGavish.

### DIFF
--- a/include/sprite.h
+++ b/include/sprite.h
@@ -19,8 +19,6 @@
 #define MAX_SPRITE_OBJECTS                16
 #define SPRITE_FOCUS_RAISE_PX             10
 #define CARD_FOCUS_SFX_PITCH_OFFSET_RANGE 512
-#define SPRING_DAMP_NUM                   90
-#define SPRING_DAMP_SHIFT                 7
 
 /** @} */
 
--- a/include/sprite.h
+++ b/include/sprite.h
@@ -19,6 +19,8 @@
 #define MAX_SPRITE_OBJECTS                16
 #define SPRITE_FOCUS_RAISE_PX             10
 #define CARD_FOCUS_SFX_PITCH_OFFSET_RANGE 512
+#define SPRING_DAMP_NUM                   90
+#define SPRING_DAMP_SHIFT                 7
 
 /** @} */
 
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -223,8 +223,8 @@ void sprite_object_update(SpriteObject* sprite_object)
     }
     else
     {
-        sprite_object->vx = (sprite_object->vx * 7) / 10;
-        sprite_object->vy = (sprite_object->vy * 7) / 10;
+        sprite_object->vx = (sprite_object->vx * SPRING_DAMP_NUM) >> SPRING_DAMP_SHIFT;
+        sprite_object->vy = (sprite_object->vy * SPRING_DAMP_NUM) >> SPRING_DAMP_SHIFT;
 
         sprite_object->x += sprite_object->vx;
         sprite_object->y += sprite_object->vy;
@@ -238,7 +238,7 @@ void sprite_object_update(SpriteObject* sprite_object)
     }
     else
     {
-        sprite_object->vscale = (sprite_object->vscale * 7) / 10;
+        sprite_object->vscale = (sprite_object->vscale * SPRING_DAMP_NUM) >> SPRING_DAMP_SHIFT;
         sprite_object->scale += sprite_object->vscale;
     }
 
@@ -251,7 +251,8 @@ void sprite_object_update(SpriteObject* sprite_object)
     }
     else
     {
-        sprite_object->vrotation = (sprite_object->vrotation * 7) / 10;
+        sprite_object->vrotation =
+            (sprite_object->vrotation * SPRING_DAMP_NUM) >> SPRING_DAMP_SHIFT;
         sprite_object->rotation += sprite_object->vrotation;
     }
 
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -11,8 +11,10 @@
 #include <tonc.h>
 #include <tonc_oam.h>
 
-// Damping Constants: SPRING_DAMP_NUMERATOR/2^SPRING_DAMP_DENOM_SHIFT ≈ 0.699 damping factor
-// SPRING_DAMP_ROUNDING = 2^(SPRING_DAMP_DENOM_SHIFT - 1) rounding for fixed-point arithmetic
+// Damping Constants: SPRING_DAMP_NUMERATOR/2^SPRING_DAMP_DENOM_SHIFT ~ 0.699 damping factor
+//
+// SPRING_DAMP_ROUNDING = 2^(SPRING_DAMP_DENOM_SHIFT - 1) rounding for fixed-point arithmetic by
+// adding half the denominator to round instead of truncating
 #define SPRING_DAMP_NUMERATOR   179
 #define SPRING_DAMP_DENOM_SHIFT 8
 #define SPRING_DAMP_ROUNDING    (1 << (SPRING_DAMP_DENOM_SHIFT - 1))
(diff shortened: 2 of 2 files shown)


include/sprite.h
```diff
@@ -19,6 +19,8 @@
 #define MAX_SPRITE_OBJECTS                16
 #define SPRITE_FOCUS_RAISE_PX             10
 #define CARD_FOCUS_SFX_PITCH_OFFSET_RANGE 512
+#define SPRING_DAMP_NUM                   90
+#define SPRING_DAMP_SHIFT                 7
```
MeirGavish: Move these defines to sprite.c, since it's part of the internal implementation, they don't need to be accessible to other modules through the header.

Also change the names to make them something like `SPRING_DAMP_NUMERATOR` (so it's not confused with "number") and `SPRING_DAMP_DENOM_SHIFT` to make their purpose clearer from the name.
And finally add a comment explaining the calculation and why it's done - that a damp coefficient is used and it's value is `SPRING_DAMP_NUMERATOR / 2^SPRING_DAMP_DENOM_SHIFT ` which is about `0.7` and that it is done as shifts to avoid costly divisions.

MeirGavish (changes requested):
Great! It seems like this actually solves #427 so I'm linking it!
It looks good, the only thing is it seems like there's a small jitter to the cards after activating - they do a small additional rotation.

https://github.com/user-attachments/assets/0343e2d8-a721-4189-ac78-d72d3ec58dc5

What seems to solve this is increasing the resolution a little more and changing the numbers to **`179`** and **`8`** so the result is closer to 0.7.

MeirGavish:
@ricfehr3 Since this seems to fix #427 and the bug exists in the v1.1 release, I want to make another release as soon as this is merged so we have a (hopefully) fully bug-free release.

SuperbMuffin:
I fixed the small jitter, but it still has a weird double bounce on the cards that are not selected when 'hovering' over them. Looking into it now.

ricfehr3:
> @ricfehr3 Since this seems to fix #427 and the bug exists in the v1.1 release, I want to make another release as soon as this is merged so we have a (hopefully) fully bug-free release.

Good call. Let's do exactly what you described and bump the release to v1.2. I don't want to open the can of worms on how to version releases yet, but this justifies a hotfix release.

MeirGavish:
> > @ricfehr3 Since this seems to fix #427 and the bug exists in the v1.1 release, I want to make another release as soon as this is merged so we have a (hopefully) fully bug-free release.
> 
> Good call. Let's do exactly what you described and bump the release to v1.2. I don't want to open the can of worms on how to version releases yet, but this justifies a hotfix release.

Yeah, I probably should have named them v0.1.0 and v0.1.1...

source/sprite.c
```diff
@@ -223,8 +226,10 @@ void sprite_object_update(SpriteObject* sprite_object)
     }
     else
     {
-        sprite_object->vx = (sprite_object->vx * 7) / 10;
-        sprite_object->vy = (sprite_object->vy * 7) / 10;
+        sprite_object->vx =
+            (sprite_object->vx * SPRING_DAMP_NUMERATOR + 128) >> SPRING_DAMP_DENOM_SHIFT;
+        sprite_object->vy =
+            (sprite_object->vy * SPRING_DAMP_NUMERATOR + 128) >> SPRING_DAMP_DENOM_SHIFT;
```
MeirGavish: Why `+ 128`?

source/sprite.c
```diff
@@ -11,6 +11,9 @@
 #include <tonc.h>
 #include <tonc_oam.h>
 
+// Damping Constants
+#define SPRING_DAMP_NUMERATOR   179
+#define SPRING_DAMP_DENOM_SHIFT 8
```
MeirGavish: Add a comment explaining the calculation

MeirGavish (changes requested):

MeirGavish:
> I fixed the small jitter, but it still has a weird double bounce on the cards that are not selected when 'hovering' over them. Looking into it now.

Yeah, you're right, I missed that.

MeirGavish:
> I fixed the small jitter, but it still has a weird double bounce on the cards that are not selected when 'hovering' over them. Looking into it now.

Further increasing the resolution doesn't resolve this... I think a proper fix would be to increase the value of `epsilon` to but I'm a little hesitant...
I tried a value of 0.015 and it seems to work but maybe the value needs a bit more thought...
Edit: I think `(FIX_ONE >> 6)` ( `= 1/2^6 = 0.015625` ) would be good, that way there is no error in the fixed point representation. 
Of course add a comment with the decimal value.

```
    const FIXED epsilon = (FIX_ONE >> 6);
```

source/sprite.c
SuperbMuffin: I made it more clear in the code, it adds rounding. I noticed that that fixed a few issues without having to increase the resolution.

SuperbMuffin:
Looks all fixed, I added comments in to explain all the calculations and also used @MeirGavish's epsilon value and that fixed the double bounce.

source/sprite.c
```diff
@@ -211,7 +217,7 @@ void sprite_object_update(SpriteObject* sprite_object)
     sprite_object->vrotation += (sprite_object->trotation - sprite_object->rotation) / 8;
 
     // set velocity to 0 if it's close enough to the target
-    const FIXED epsilon = float2fx(0.01f);
+    const FIXED epsilon = (FIX_ONE >> 6);
```
MeirGavish: Add a comment explaining this value too.

source/sprite.c
```diff
@@ -223,8 +229,12 @@ void sprite_object_update(SpriteObject* sprite_object)
     }
     else
     {
-        sprite_object->vx = (sprite_object->vx * 7) / 10;
-        sprite_object->vy = (sprite_object->vy * 7) / 10;
+        // Apply damping with proper rounding: (x * 179 + 128) >> 8 rounds to nearest instead of
```
MeirGavish:
```suggestion
        // Apply damping with proper rounding: add half the denominator to round to nearest instead of
```
I understand now, good call. Just make the comment more descriptive rather than explicitly using the constants which may change.

source/sprite.c
```diff
@@ -238,7 +248,11 @@ void sprite_object_update(SpriteObject* sprite_object)
     }
     else
     {
-        sprite_object->vscale = (sprite_object->vscale * 7) / 10;
+        // Apply damping with proper rounding: (x * 179 + 128) >> 8 rounds to nearest instead of
+        // truncating
```
MeirGavish: Also combine this comment with the one in the definition of the constant, rather than repeating it every time.

MeirGavish (changes requested):
Looks good, things seem to be working well.
Just a few changes to the documentation to make it a little clearer.

Also, I gave this another test and it seems like I was a little too optimistic about solving #427, unfortunately it still exists in this PR and I just missed it... so I'm removed the attachment.

image
image

SuperbMuffin:
done!

MeirGavish (approved):
Great! Approved.

MeirGavish:
> Also, I gave this another test and it seems like I was a little too optimistic about solving #427, unfortunately it still exists in this PR and I just missed it... so I'm removed the attachment.
> 
> image image

@ricfehr3 I'm merging this, hope that's fine with you, just note that it doesn't have the fix I thought so I won't be able to make the release after all...

ricfehr3:
> > Also, I gave this another test and it seems like I was a little too optimistic about solving #427, unfortunately it still exists in this PR and I just missed it... so I'm removed the attachment.
> > image image
> 
> @ricfehr3 I'm merging this, hope that's fine with you, just note that it doesn't have the fix I thought so I won't be able to make the release after all...

Thanks for reviewing and merging! But darn, I was hoping we could get that bug out of the way.

PR #447, opened 2026-04-23
A few small updates to the bug report template

MeirGavish:
Added instructions for specifying build from source in the bug report template and added issue type bug.
That's usually how I report bugs.
I also added the instruction to just upload a zip of the rom used, so we can run the script and check the version ourselves.


MathisMartin31 (approved):

ricfehr3 (approved):
Sorry, this took me a minute. Approved and merging

PR #445, opened 2026-04-19
Rebrand to GBAlatro in main menu (art by @O-C1NNAM0N-O)

MathisMartin31:
Related to discussion #171 
Art by @O-C1NNAM0N-O

Issue #142, opened 2025-09-22
Seeded Runs


lolkid-life:
I've been loving this game so far and this is going to be a small suggestion but a way to have a seed system where you could like change the numbers and be able to control what seed is used or use a random seed.

MeirGavish:
This could be a nice feature to add, but not a simple one at all. The seed would need a form of output  - displayed at the end of the run, and input - allow the player to input a seed in the beginning of a run as well as choose whether it's a seeded run. 
The input can be tricky because it means adding some form of keyboard input though maybe the seed can be encoded as a series of controller button presses like a cheat code.
The output is more screen space at the end of the round. We also want to allocate screen space for a "New Run"/"Main Menu" button there.
Either way, I think this is a bit lower priority right now but can definitely be worked on in the future.

ricfehr3:
I randomly had a thought about this. Wanted to share.

I think one way we could do it is with a hex keypad. You could have a 4x4 selection with 0-F:

```
| 1 | 2 | 3 | F |
| 4 | 5 | 6 | E |
| 7 | 8 | 9 | D | 
| 0 | A | B | C |
```

This will let you fill in the seed as 1-8 characters, covering the 4-bytes required for the rng seed integer.

Then, that would be another game state in a new screen. The background would be another png->grit situation.

MathisMartin31:
Hey @ricfehr3 I've been cooking up some more menu mockups, since seeded runs are just around the corner with all the state refactors, the save system and the RNG system getting its own file. ^^
I also did the deck selection and resume menus, since I think they should be all together.

Seed input keypad. A bit empty on the sides but I don't know what to put there ><
<img width="480" height="320" alt="Image" src="https://github.com/user-attachments/assets/4525df05-be15-4fb5-8e5c-e73e050c98a7" />

Deck selection (shown with Yellow deck, which is the deck with the most descriptive text among the ones we wanted to implement). The little arrows will have a little press animation for a few frames :)
<img width="480" height="320" alt="Image" src="https://github.com/user-attachments/assets/c726f807-bc45-4ba9-b61d-a73de6aafd28" />

Resume game screen, with the "Seed" button disabled because we can't change the seed of the saved run.
Edit: this screen would not have the little arrows on the sides because you can't change the deck either
<img width="480" height="320" alt="Image" src="https://github.com/user-attachments/assets/95fc17a3-77e4-4103-bd70-9c9a6d99d708" />

MathisMartin31:
[...]

PR #449, opened 2026-04-25
[Feature] Implement options menu and reading/writing values from/to the SRAM

### LINKED ISSUE #198: New Debug Screen
From previous [discussion](https://github.com/cellos51/balatro-gba/pull/182#issuecomment-3471474631) It would be useful to have a debug screen that displays the git hash on the cartridge and other debug information. 

I'm thinking that by hitting a certain key combo (maybe hold L+R+A+B) at the splash screen it switches to the debug screen. 

The work involved with this would essentially be copying the splash screen, adding the appropriate `enum`s, and adding logic to handle the button combo during the splash screen.

### LINKED ISSUE #328: Use high-contrast card colors
In the original game, spades and clubs are easy to distinguish by their shapes. However, on the GBA, where pixel resolution is limited, increasing the color contrast might work better—for example, as shown in this image.

Image

https://www.deviantart.com/advancedfan2020/art/Balatro-The-Main-Deck-High-Contrast-69x93-px-1232527040
Alternatively, the colors could be mapped to the four multiplier jokers:

Image

In the current version, it takes some effort to distinguish spades from clubs when using an emulator. It’s also unclear how this would look on real GBA hardware with a non-backlit screen.

Image

### LINKED ISSUE #372: Options Menu
Add an option menu accessible through an "Options" button from the main menu. The first option to be able to set there would be the game speed, I think we should allow setting any of 1, 2, or 4, maybe then we can optimize the divisions of game speed to bitwise shifts.
We can add left/right buttons to increase or decrease the game speed, or just one button to increase and wrap back to 1.
Next we can add high contrast colors once it's implemented, it should be a toggle.

MathisMartin31:
Closes #372 
Depends on #445 and #416 

Implement a nice little options menu so that we can change the game speed, use high contrast cards, and adjust music and SFX volumes separately.

I also figured out how to read/write data from/to the SRAM (or Cart RAM), so we can now save and load persistent data! (yaaaaay :tada:). 
~The tricky part is that the maximum width of read/writes is 8 bits, so any data we want to put there needs to be split into 1-byte words and written sequentially. I don't know how expensive that is in terms of cycles, but since we're not doing it too often it's probably fine.~ @ricfehr3 figured it out, I was overthinking it lol. Also there's a githash version check to ensure data compatibility.

For now this is only used for the options values, but I reckon we could try and save run data, let's say after each shop only, so that we can quit and resume the game, without the hassle of taking game states into account. The RNG state, Jokers, deck order, money and consumables would probably be all we would need. Again, another PR and some discussions will be needed but it's definitely doable now :)

PS: also a heads up for anyone doing the same mistake as me when setting up my Everdrive GBA Mini, be sure to set the save mode of the ROM to SRAM, I don't get why since a cart using SRAM or flash32/64 should have the same memory mapped to the same address according to the memory layout documentation, but the game doesn't save when the ROM is in Flash32/64 save mode. Had a small heart attack yesterday evening when the game would not save on my modded GBA :')


ricfehr3:
Yes! This is great. Incredible work, saving is really cool!

MeirGavish:
Really cool! Great work! This is a gigantic PR and I probably won't be able to review everything but I do want to take a look at some things so I requested a review from myself and it will probably be a partial small one.

MathisMartin31:
> Really cool! Great work! This is a gigantic PR and I probably won't be able to review everything but I do want to take a look at some things so I requested a review from myself and it will probably be a partial small one.

Thanks for taking a look ^^

Don't worry too much about the size, a good part of the diff is from #416 so it will go away when it is eventually merged :)

MathisMartin31:
I've rebased the branch on main, I have a couple of issues but I had to go to work so I'll fix them and push this evening :)

MathisMartin31:
Okay the issue wasn't too bad, should be fixed now. @ricfehr3 sorry for the (still) rather long PR, good luck reviewing it ><'

include/def_state_info_table.h
```diff
@@ -1,12 +1,13 @@
...
-DEF_STATE_INFO(GAME_STATE_SPLASH_SCREEN, splash_screen_on_init, splash_screen_on_update, splash_screen_on_exit) 
-DEF_STATE_INFO(GAME_STATE_MAIN_MENU, game_main_menu_on_init, game_main_menu_on_update, game_main_menu_on_exit)
-DEF_STATE_INFO(GAME_STATE_GAME_START, noop, game_start, noop)
-DEF_STATE_INFO(GAME_STATE_PLAYING, game_round_on_init, game_playing_on_update, noop)
-DEF_STATE_INFO(GAME_STATE_ROUND_END, noop, game_round_end_on_update, game_round_end_on_exit)
-DEF_STATE_INFO(GAME_STATE_SHOP, noop, game_shop_on_update, game_shop_on_exit)
-DEF_STATE_INFO(GAME_STATE_BLIND_SELECT, game_blind_select_on_init, game_blind_select_on_update, game_blind_select_on_exit)
-DEF_STATE_INFO(GAME_STATE_LOSE, game_lose_on_init, game_lose_on_update, game_over_on_exit)
-DEF_STATE_INFO(GAME_STATE_WIN, game_win_on_init, game_win_on_update, game_over_on_exit)
+DEF_STATE_INFO(GAME_STATE_SPLASH_SCREEN, splash_screen_on_init,     splash_screen_on_update,     splash_screen_on_exit    )
```
MathisMartin31, line 3: Let me know, but I feel like the func names being aligned like this is waaaay nicer to the eyes ^^'

graphics/deck_gfx.png
MathisMartin31, line 1: Pixels shouldn't have changed here that's normal, I duplicated some colorsin the palette to allow for a nicer high contrast palette

graphics/background_main_menu_gfx.png
MathisMartin31, line 1: This still doesn't use indexed colors I couldn't get it to work, so the button and outline color indices might change again in the future if we add more buttons like a Resume one.

I did get to make it lighter by removing a bunch of duplicate colors, though.

include/game_variables.h
```diff
@@ -16,8 +29,18 @@
  */
 typedef struct
 {
-    s32 timer;
+    s32 timer; // This might already exist in libtonc but idk so i'm just making my own
     u32 rng_seed;
+    u32 rng_step; // Position in the rng sequence.
```
MathisMartin31, line 36: There's still a small issue with `rng_step` since the first boss blind is rolled during game init when tokens are created, after save data is loaded, this counter increases each time we boot the game. Not an issue right now but will have to be adressed when we get to seeded runs (#142)

scripts/get_hash.py
```diff
@@ -20,7 +20,7 @@ def grep_binary_offsets(path, pattern):
         sys.exit(1)
 
     file_path = sys.argv[1]
-    pattern   = "GBALATRO_VERSION"
+    pattern   = "GBALATRO-VERSION"
```
MathisMartin31: This is purely for aesthetic when showing the version string on screen, since the '_' char has been remapped to '.8' in the font file

source/game/main_menu.c
```diff
@@ -84,29 +96,76 @@ void game_main_menu_on_update()
...
     {
-        if (selection_x < IMPLEMENTED_BUTTONS - 1)
+        if (selection_x < MAIN_MENU_NB_BTN - 1)
         {
             selection_x++;
         }
     }
 
-    if (selection_x == PLAY_BTN_IDX)
+    // DEBUG: display git hash by holding L+R+START+SELECT
```
MathisMartin31: I added this for debug, but I thought it was neat so I left it in so let me know should this stay or should this go? (ha ha, I feel old...)

source/game/options_menu.c
```diff
@@ -0,0 +1,520 @@
...
+void game_options_menu_change_background(void)
+{
+    tte_erase_screen();
+    CBB_CLEAR(MAIN_BG_CBB);
+
+    GRIT_CPY(pal_bg_mem, background_options_menu_gfxPal);
+    GRIT_CPY(&tile_mem[MAIN_BG_CBB], background_options_menu_gfxTiles);
+    GRIT_CPY(&se_mem[MAIN_BG_SBB], background_options_menu_gfxMap);
+
+    // Recreate menu frame
```
MathisMartin31: Using only the minimal set of tiles costs a bit extra when showing the menu, but I saved a whopping 500B of image data so I think it was worth it (lol).

source/game/options_menu.c
```diff
@@ -0,0 +1,520 @@
...
+    {
+        if (selection_y < NB_OPTIONS_BUTTONS - 1)
+        {
+            selection_y++;
+        }
+    }
+
+    disable_all_outlines_except_self(selection_y);
+
+    switch (selection_y)
```
MathisMartin31: I'm realizing now that maybe I should have used the selection_grid structure for this...

source/save.c
```diff
@@ -0,0 +1,161 @@
...
+    DTAG_VOUCHER,
+    DTAG_SKIP_TAG,
+    DTAG_END,
+    DTAG_INVALID = UNDEFINED
+};
+
+extern char balatro_version[];
+
+// DelimiterTag is an enum of size sizeof(int) = 4
+#define CARDS_TAG_SIZE 4
```
MathisMartin31: Not used yet, I'm working on it in another branch since saving the game is not the point of this PR

include/game/main_menu.h
```diff
@@ -15,23 +15,17 @@ void game_main_menu_change_background(void);
 
 /**
  * @brief Main menu state initialization
- *
- * @param vars passed @ref GameVariables struct
```
ricfehr3, line 19: Nice catch!

include/def_state_info_table.h
ricfehr3: I completely agree. Looks good.

include/game.h
```diff
@@ -166,9 +167,12 @@ int get_money(void);
...
 void display_money();
 void set_retrigger(bool new_retrigger);
+int get_ante();
+int get_round();
+enum BlindType get_current_blind();
+enum BlindType get_next_boss_blind();
 
-int get_game_speed(void);
-void set_game_speed(int new_game_speed);
+u32 get_rand();
```
ricfehr3: **I STAY VOIDED**

Put `void` in the empty functions please.

include/game_variables.h
```diff
@@ -16,8 +29,18 @@
  */
 typedef struct
 {
-    s32 timer;
+    s32 timer; // This might already exist in libtonc but idk so i'm just making my own
```
ricfehr3, line 34: Thank you for reviving this. I feel like we should preserve these og comments sometimes, haha.

scripts/get_hash.py
ricfehr3: I don't mind changing it, just update the regex to handle both old and new ROMs:
```suggestion
    pattern   = "GBALATRO[_,-]VERSION"
```

source/game/main_menu.c
ricfehr3: Should I stay or should I go

I like it! I did have trouble with the `if` here though. Not sure why but removing one of those buttons from (start, select, L, R) make it work fine, otherwise all four never seems to trigger.

ricfehr3: Also, I would guess that most people who would end up using this for bug reports may want a screenshot or a picture of it. So I'm thinking it should be turned on once then left. Since it's only going to be in the main menu anyway.

source/game/main_menu.c
```diff
@@ -84,29 +96,76 @@ void game_main_menu_on_update()
...
+            if (key_hit(SELECT_CARD))
+            {
+                game_change_state(GAME_STATE_OPTIONS_MENU);
+            }
+            break;
+        }
+
+        default:
+            break;
     }
```
ricfehr3: This whole block might benefit from just being a table. A lot of stuff is repeated in the switch case. I think just a struct with: Main color Pid, Outline Pid, and a callback for the `key_hit(SELECT_CARD)`. Then the logic can look more like:

```c
typedef struct {
    u32 color_pid;
    u32 outline_pid;
    void (*on_select)();
} MainMenuEntry;

...

static MainMenuEntries[] = { Play button info, options button info }

...
memcpy16(
    &pal_bg_mem[OPTIONS_BUTTON_OUTLINE_PID],
    &pal_bg_mem[entry[selection_x].color_pid],
    1
);

memset16(&pal_bg_mem[entry[selection_x].outline_pid], BTN_HIGHLIGHT_COLOR, 1);

if (key_hit(SELECT_CARD)) entry[selection_x].on_select();
```

I've noticed I have linked back to this example a few times as I'm reviewing. So I wanted to add some general statements here about the design. 

In general this pattern is more maintainable than various switch/case if/else blocks. The data is clearly presented and is unlinked from the logic. This makes it easy to add and subtract different entries to and from the table in one spot, instead of wherever dependent switch/case/if/else blocks are. Additionally it makes the logic centralized to one written piece of code. 


What do you think?

graphics/background_main_menu_gfx.png
ricfehr3: I'm okay with that. I think you should make an issue to track it.

include/game.h
```diff
@@ -166,9 +167,12 @@ int get_money(void);
 void set_money(int new_money);
 void display_money();
 void set_retrigger(bool new_retrigger);
+int get_ante();
+int get_round();
```
ricfehr3: I think these should be moved to the global vars struct. Because it will be part of the save file.

source/game/common_ui.c
```diff
@@ -15,18 +16,18 @@ static const BackgroundRenderCallback bgCallbacks[] = {
...
     [BG_SHOP] = NULL,
     [BG_BLIND_SELECT] = NULL,
+    [BG_OPTIONS_MENU] = game_options_menu_change_background,
     [BG_MAIN_MENU] = game_main_menu_change_background,
 };
 
 void change_background(enum BackgroundId id)
 {
-    if (id == BG_MAIN_MENU)
+    if (id != background && bgCallbacks[id] != NULL)
```
ricfehr3, line 25: Way better, thank you!

source/game/main_menu.c
```diff
@@ -17,16 +17,24 @@
...
+    PLAY_BTN_IDX,
+    OPTIONS_BTN_IDX,
+    MAIN_MENU_NB_BTN
+};
 
 // Pixel sizes
 #define MAIN_MENU_ACE_T_X 88
 #define MAIN_MENU_ACE_T_Y 26
 
+extern char balatro_version[];
```
ricfehr3: I don't think this should be externed. I think it should be possible to put that variable into a new `version.h` header. `extern` is also used in save.c, that should also use the header.

source/game/main_menu.c
```diff
@@ -44,9 +52,14 @@ void game_main_menu_change_background(void)
 
     // Disable the button highlight colors
     memcpy16(&pal_bg_mem[PLAY_BUTTON_OUTLINE_PID], &pal_bg_mem[PLAY_BUTTON_MAIN_COLOR_PID], 1);
+    memcpy16(
+        &pal_bg_mem[OPTIONS_BUTTON_OUTLINE_PID],
+        &pal_bg_mem[OPTIONS_BUTTON_MAIN_COLOR_PID],
+        1
+    );
```
ricfehr3: I think you should use the other suggested table from this comment **[here](https://github.com/GBALATRO/balatro-gba/pull/449/changes#r3165242557)**, and use a for loop.

source/save.c
```diff
@@ -0,0 +1,161 @@
...
+    if (g_game_vars.music_volume > VOLUME_VALUE_MAX)
+        g_game_vars.music_volume = DEFAULT_MUSIC_VOLUME;
+
+    if (g_game_vars.sound_volume > VOLUME_VALUE_MAX)
+        g_game_vars.sound_volume = DEFAULT_SOUND_VOLUME;
+
+    // return to where we were in the random sequence so that the run stays reproducible
+    for (u32 i = 0; i < g_game_vars.rng_step; i++)
+    {
+        (void)rand();
```
ricfehr3, line 162: This is interesting, but I'm curious as to what the purpose of it is? Do we want to have games load in with the same hand to prevent save spamming?

include/save.h
```diff
@@ -0,0 +1,16 @@
+/**
+ * @file save.h
+ *
+ * @brief Utils functions to save/load data structures from/to the SRAM.
```
ricfehr3, line 4: Can you put the table from our discussion here so in doxygen you can see it rendered in markdown (hopefully, lol)

| word | Byte 0 | Byte 1 | Byte 2 | Byte 3 | name | purpose |
| - | - | - | - | - | - | - | 
| 0 | 0x47 | 0x42 | 0x41 | 0x4C | MAGIC | Identify if proceeding data is valid and not junk, spells "GBAL" |
| 1 | Dirty | **H[0]** | **H[1]** | **H[2]** | GITHASH_LOW | Dirty flag, followed by the first 3 bytes of shortened git hash **H**
| 2 | **H[3]** | **H[4]** | **H[5]** | **H[6]** | GITHASH_HIGH | Last 4 bytes of shortened git hash **H**, with a dirty flag |

include/save.h
```diff
@@ -0,0 +1,16 @@
...
+ */
+#ifndef SAVE_H
+#define SAVE_H
+
+#include "game_variables.h"
+
+#include <tonc.h>
+
+void save_game();
+void load_game();
```
ricfehr3, line 23: Please document with doxygen and add `void`

source/save.c
```diff
@@ -0,0 +1,161 @@
...
+    set_save_valid();
+    write_sram(GAME_BASE, (const u8*)&g_game_vars, sizeof(g_game_vars));
+}
+
+/**
+ ** @brief Load game data from SRAM.
+ *
+ * @sa save_game
+ */
+void load_game(void)
```
ricfehr3, line 152: These doxygen comments need to be in the header.

source/save.c
```diff
@@ -0,0 +1,161 @@
...
+#define CHECK_MAGIC     0x4C414247 // Spells GBAL, used to determine if the save data is junk
+#define CHECK_HASH_SIZE 7
+#define GIT_HASH_START  17 // starts after "GBALATRO-VERSION:" in the balatro_version var
+
+typedef struct SaveCheckInfo
+{
+    u32 magic;
+    bool dirty;
+    char githash[7];
+} SaveCheckInfo;
```
ricfehr3: nitpick: I would just call it "SaveHeader"

source/save.c
```diff
@@ -0,0 +1,161 @@
...
+    DTAG_JOKER,
+    DTAG_TAROT,
+    DTAG_PLANET,
+    DTAG_VOUCHER,
+    DTAG_SKIP_TAG,
+    DTAG_END,
+    DTAG_INVALID = UNDEFINED
+};
+
+extern char balatro_version[];
```
ricfehr3: Move to header as per [other comment](https://github.com/GBALATRO/balatro-gba/pull/449/changes#r3165312960)

source/game/main_menu.c
ricfehr3: Oh also, I think this would solve #198

source/save.c
```diff
@@ -0,0 +1,161 @@
...
+        return;
+
+    for (u32 i = 0; i < size; i++)
+    {
+        bytes[i] = sram_mem[sram_base + i];
+    }
+}
+
+// Check 7 chars of balatrà_version after the "GBALATRO_VERSION" preffix
+static inline bool check_hash(const char* preffix)
```
ricfehr3: typo: "prefix"

source/game.c
```diff
@@ -4554,9 +4580,12 @@ static void game_shop_on_exit()
     list_clear(&_shop_jokers_list);
 
     increment_blind(BLIND_STATE_DEFEATED); // TODO: Move to game_round_end()?
+
+    // test
```
ricfehr3: Leftover test code

source/game.c
```diff
@@ -1037,15 +1053,24 @@ int get_num_hands_remaining(void)
...
 }
 
-int get_game_speed(void)
+int get_ante()
 {
-    return game_speed;
+    return ante;
 }
 
-// for the future when a menu actually lets this variable be changed.
```
ricfehr3, line 1045: The future is now

source/game/options_menu.c
```diff
@@ -0,0 +1,520 @@
...
+                OPTIONS_CONTRAST_VALUE_DEST_POS
+            );
+        }
+        else
+        {
+            main_bg_se_copy_rect(
+                OPTIONS_CONTRAST_VALUE_NO_SRC_RECT,
+                OPTIONS_CONTRAST_VALUE_DEST_POS
+            );
+        }
```
ricfehr3: I see this high contrast if statement a few times above as well. I think this would similarly benefit from a table as described in the [other comment](https://github.com/GBALATRO/balatro-gba/pull/449/changes#r3165242557).

source/game/options_menu.c
ricfehr3: I believe you're right, I'm afraid.

source/game/options_menu.c
ricfehr3: hahaha, absolutely worth it.

source/game/options_menu.c
```diff
@@ -0,0 +1,520 @@
...
+    if (highlighted_btn != BACK_BTN_IDX)
+    {
+        memcpy16(
+            &pal_bg_mem[BACK_BUTTON_OUTLINE_COLOR_PAL_IDX],
+            &pal_bg_mem[BACK_BUTTON_MAIN_COLOR_PAL_IDX],
+            1
+        );
+    }
+
+    u16 color = 0x213f;
```
ricfehr3: Magic number

source/game/options_menu.c
```diff
@@ -0,0 +1,520 @@
...
+        );
+    }
+    if (highlighted_btn != BACK_BTN_IDX)
+    {
+        memcpy16(
+            &pal_bg_mem[BACK_BUTTON_OUTLINE_COLOR_PAL_IDX],
+            &pal_bg_mem[BACK_BUTTON_MAIN_COLOR_PAL_IDX],
+            1
+        );
+    }
```
ricfehr3, line 213: Similarly, a table or array would work better here as to this other [comment](https://github.com/GBALATRO/balatro-gba/pull/449/changes#r3165242557).

ricfehr3 (changes requested):
This was a fun one to test and I'm very excited for it to be added to the game! Great Work @MathisMartin31 . You always add excellent content to the game. As you said, this is a big PR. I gave it a good look this first pass, but I'll be back a few more times. I have some comments about some design considerations and then the other usual PR stuff. 

A few notes:
- That little debug menu will close #198, thanks.
- The sound and the music doesn't change. That will need to be logged in an issue.
- Additionally, the screen seems to tear when I switch back from the options menu to the main menu. I believe this is because it saves the game on exiting, which is a slow byte-per-byte thing as you experienced. We'll have to think on how to approach that. In my head I've been thinking:
  - Have a separate GAME_SAVE state that gracefully pauses the game graphics updates, saves, then unpauses
  - Have a button that does it instead (I'd rather not)
  - Consider adding fade-in/out to/from black/white with the `REG_BLDY` register

Below is the screen tearing I was experiencing:

https://github.com/user-attachments/assets/e92ee226-22c8-4307-8278-76c469247793

Thanks for the hard work, this rocks.

source/game/main_menu.c
MathisMartin31: > Should I stay or should I go
> 
> I like it! I did have trouble with the `if` here though. Not sure why but removing one of those buttons from (start, select, L, R) make it work fine, otherwise all four never seems to trigger.

That's weird, I'll try on my GBA but on mgba I never missed inputs 🤔 

> Also, I would guess that most people who would end up using this for bug reports may want a screenshot or a picture of it. So I'm thinking it should be turned on once then left. Since it's only going to be in the main menu anyway.

I agree, I'll make it a toggle.

source/card.c
```diff
@@ -19,9 +19,33 @@ const static u16 _card_sprite_lut[NUM_SUITS][NUM_RANKS] = {
...
 };
 
+// Palettes to use with cards
+// Could not fit inside the same PNG so we have to hardcode it unfortunately
+#define CARDS_PAL_SIZE 16
+// clang-format off
+const static u16 high_contrast_palette[CARDS_PAL_SIZE] = {
+    0x0000, 0x7FFF, 0x6B38, 0x023D, 0x61E0, 0x001B, 0x71B1, 0x28E7,
+    0x0017, 0x28E7, 0x7E82, 0x029F, 0x739C, 0x28E7, 0x414C, 0x0000
+};
```
MeirGavish: I think you can avoid hardcoding the colors. You can create another separate deck PNG with the high contrast colors mapped to these indices and then tell grit to only load the palette from there.

source/game/main_menu.c
MeirGavish: @MathisMartin31 @ricfehr3 
There's already a button.h/.c that does just that, I added it for the play, discard, and sort buttons.

MeirGavish: Also, changing the main menu to use selection_grid instead of `selection_x`/`selection_y` with these switch cases would be so much nicer.

source/game/options_menu.c
MeirGavish: Actually I think it would have been much better to create the image file as it should look rather than generate it in the code, grit optimizes the tiles and it's the same data in GBA memory anyway and it's much easier to follow and edit.

source/game/options_menu.c
MeirGavish: Definitely... The only issue would be support for changing the volume with left and right though I think we could just add dedicated buttons for that instead.

source/game/options_menu.c
MeirGavish: Note that doing this here depending on a static bool creates a bug in loading the save data where loaded high contrast doesn't apply unless you go to the options menu.

MeirGavish (changes requested):
Great work! Loving the high contrast cards!
There's a bug when loading that only applies the saved high contrast when going to the options menu, otherwise if you start playing, the saved setting isn't applied.
Also, I'm wary of adding the volume slider options when they still don't do anything, it can be quite confusing to players.

source/card.c
MathisMartin31: Damn I always forget I can do this without having to deal with the tiles if I just don't load them lol

source/game/options_menu.c
MathisMartin31: > Actually I think it would have been much better to create the image file as it should look rather than generate it in the code, grit optimizes the tiles and it's the same data in GBA memory anyway and it's much easier to follow and edit.

Wait wait wait, so even if I have an image with duplicated tiles, it doesn't take more space in ROM/RAM for the compiled game?

source/game/options_menu.c
MathisMartin31: Oh yeah that makes sense, so either I move it to the load/save funcs, or since we don't actually see cards before playing, I move it to the GAME_START state init function.

Edit: I moved the high contrast toggle to the game_start function, it works fine :)

source/game/options_menu.c
MathisMartin31: This one doesn't do anything, I was testing stuff and forgot to remove that ^^'

source/game/options_menu.c
MeirGavish: Yes. For example, this is the tile data (Tools->Game state views->View tiles...)  for the game round background.

image

There are no duplicates there even though the background image has a lot of duplicates for the round end panel and the top left panel for example. Grit removes them and packs them into the tile data in VRAM (`tile_mem`), the tilemap itself is the 32x32 grid in `se_mem`/`se_mat`  that maps into the tile data.

MathisMartin31: Oh that's neat, guess I'll do that instead. And here I was so proud of my (measly) optimization ><

source/save.c
MathisMartin31: Yes, that's the idea was to try and preserve the sequence of random numbers you get. I think people use this when exploring specific seeds to try and optimize their run. And it also prevent save scumming as an added bonus.

MathisMartin31:
@ricfehr3 @MeirGavish
I think I addressed everything except the flickering. Going to take a look at it now but I don't know if I'll solve it today.

~Also, sound effects now follow their volume slider. I found the function playing the sound and multiplied the volume by `(game_vars.sound_volume / MAX_VOLUME)`, but I can't figure out how to do it for the main music. Anyone got any idea?~  
Nv, found it ^^

MathisMartin31:
> We'll have to think on how to approach that. In my head I've been thinking:
>       * Have a separate GAME_SAVE state that gracefully pauses the game graphics updates, saves, then unpauses
>       * Have a button that does it instead (I'd rather not)
>       * Consider adding fade-in/out to/from black/white with the `REG_BLDY` register

@ricfehr3 Why not a button? I think I could reuse the Back button and change it to a Save button whenever there are any changes, and change it back after the game has saved.

I'm not a huge fan of the fade in/out to black, Balatro doesn't do that and I think it's a shame to hide the beautiful swirly background.

ricfehr3:
> > We'll have to think on how to approach that. In my head I've been thinking:
> > * Have a separate GAME_SAVE state that gracefully pauses the game graphics updates, saves, then unpauses
> > * Have a button that does it instead (I'd rather not)
> > * Consider adding fade-in/out to/from black/white with the `REG_BLDY` register
> 
> @ricfehr3 Why not a button? I think I could reuse the Back button and change it to a Save button whenever there are any changes, and change it back after the game has saved.
> 
> I'm not a huge fan of the fade in/out to black, Balatro doesn't do that and I think it's a shame to hide the beautiful swirly background.

I don't know why I felt so strongly when I wrote that. I think a button would be fine, it makes sense and familiar.

source/game/main_menu.c
```diff
@@ -84,29 +96,85 @@ void game_main_menu_on_update()
...
+
+            if (key_hit(SELECT_CARD))
+            {
+                game_change_state(GAME_STATE_OPTIONS_MENU);
+            }
+            break;
+        }
+
+        default:
+            break;
```
ricfehr3: I think we should use `button.[h/c]` here as @MeirGavish suggested

source/save.c
ricfehr3: That's a good idea. This seems like something we should probably test. I think this needs to be put into it's own file "gbal_random.h/c" or something. Then a test should be added to ensure sample save files have the "expected random" (lol) numbers. Then this functionality can be extracted as "set_rand_steps(u32 steps)" or similar. 

This is a lot of work on top of this PR. So let me know what you think. If you like the idea we can make an issue and I can assign it to you if you'd like.

ricfehr3 (commented):
I'll have to continue more another time but I wanted to leave my comments so far and to note that the volume on the music does start to sound rough below 20%. I think it may be unavoidable and an artifact (tho I haven't tested on real hardware). I was thinking maybe just move the sliders in blocks of 20, so there's 5 positions.

MathisMartin31:
> I'll have to continue more another time but I wanted to leave my comments so far and to note that the volume on the music does start to sound rough below 20%. I think it may be unavoidable and an artifact (tho I haven't tested on real hardware). I was thinking maybe just move the sliders in blocks of 20, so there's 5 positions.

Done ^^

MathisMartin31:
> I don't know why I felt so strongly when I wrote that. I think a button would be fine, it makes sense and familiar.

I implemented the save button so now we don't read or write to SRAM around the time of the scene transition, but I still get some flicker. This is when leaving the menu after changing nothing, we can clearly see the main menu logo being mangled for a frame before falling into place:

Edit: it's still not perfect, but I added some stuff to only change affine backgrounds if we really need to. The previous implementation had no checks at all and was always moving stuff around. This means that since the `game_main_menu_on_init` function contained a call to `affine_background_change_background`, we would move a lot of stuff around even when coming from the options menu. The flicker is still visible though, but I think it's marginally better.

image
<br>

~Also I watched the very next frame and it seems the main menu's ace card takes exactly one frame to get to its intended position:~

Edit :fixed by adding `card_object_update(main_menu_ace)` to the `game_main_menu_on_init()` function so the card is updated immediately after being created.

image

source/game/main_menu.c
MathisMartin31: Done ^^

source/game/main_menu.c
MeirGavish: You should do this in a for loop
```suggestion
    for (int i = 0; i < NUM_ELEM_IN_ARR(main_menu_buttons); i++)
    {
        button_set_highlight(&main_menu_buttons[i], false);
    }
```

source/game/main_menu.c
```diff
@@ -9,30 +9,82 @@
...
+        main_menu_return_row_size,
+        main_menu_on_selection_changed,
+        main_menu_on_key_transit,
+        {.wrap = false}
+    },
+};
+
+Button main_menu_buttons[] = {
+    {PLAY_BUTTON_OUTLINE_PID,    PLAY_BUTTON_MAIN_COLOR_PID,    play_on_pressed,    NULL},
+    //{RESUME_BUTTON_OUTLINE_PID,  RESUME_BUTTON_MAIN_COLOR_PID,  resume_on_pressed,    NULL},
```
MeirGavish: I'd say remove the commented out resume button, it's better to not have commented out code and add it only when needed.

source/game/options_menu.c
```diff
@@ -0,0 +1,618 @@
...
+    memcpy16(
+        &pal_bg_mem[SPEED_LESS_BUTTON_OUTLINE_COLOR_PAL_IDX],
+        &pal_bg_mem[MENU_BUTTON_MAIN_COLOR_PAL_IDX],
+        1
+    );
+    memcpy16(
+        &pal_bg_mem[SPEED_MORE_BUTTON_OUTLINE_COLOR_PAL_IDX],
+        &pal_bg_mem[MENU_BUTTON_MAIN_COLOR_PAL_IDX],
+        1
+    );
```
MeirGavish: Then use `button_set_highlight()` rather than changing the palette memory directly.

include/game_variables.h
```diff
@@ -8,6 +8,19 @@
 
 #include <tonc.h>
 
+#define GAME_SPEED_MIN 1
+#define GAME_SPEED_MAX 4
+
+// Volume is stored from 0 to 5 but is an increment of 20 so 0 to 100 will be displayed
+#define VOLUME_VALUE_MIN       0
+#define VOLUME_VALUE_MAX       5
+#define VOLUME_VALUE_INCREMENT 20
```
MeirGavish:
```suggestion
#define VOLUME_OPTION_MIN       0
#define VOLUME_OPTION_MAX       5
#define VOLUME_OPTION_INCREMENT 20
```

Change the name of the constants to be more specific and not confuse them with `MM_FULL_VOLUME` etc.

include/audio_utils.h
```diff
@@ -8,6 +8,12 @@
 
 #include <mm_types.h>
 
+/**
+ * @def MM_FULL_MODULE_VOLUME
+ * @brief The maximum volume for maxmod module volume (main theme)
+ */
+#define MM_FULL_MODULE_VOLUME 1024
```
MeirGavish: I guess this also warrants a rename of `MM_FULL_VOLUME` to `MM_SFX_FULL_VOLUME`

source/game/options_menu.c
```diff
@@ -0,0 +1,618 @@
...
+            "#{P:%d,%d; cx:0x%X000}%3d",
+            OPTIONS_MUSIC_VALUE_TEXT_POS.x,
+            OPTIONS_MUSIC_VALUE_TEXT_POS.y,
+            TTE_WHITE_PB,
+            (g_game_vars.music_volume * VOLUME_VALUE_INCREMENT)
+        );
+
+        music_volume_changed = false;
+        any_value_changed = true;
+    }
```
MeirGavish: Why have this series of if statements rather than put them inside the appropriate button callbacks?
It makes this function very long with lots of nested if statements...

source/game/options_menu.c
```diff
@@ -0,0 +1,618 @@
...
+        NULL,
+        {.wrap = true}
+    },
+    {
+        SOUND_VOLUME_BTN_IDX,
+        options_menu_return_row_size,
+        sound_volume_row_on_selection_changed,
+        NULL,
+        {.wrap = true}
+    },
```
MeirGavish, line 113: Are you using `.wrap` here to as a hack to allow directional input to change values? It's okay, just add a comment explaining it.

source/selection_grid.c
```diff
@@ -49,12 +49,14 @@ void selection_grid_move_selection_horz(SelectionGrid* selection_grid, int direc
...
-                &new_selection
-            );
+            selection_grid->rows[selection_grid->selection.y].on_selection_changed != NULL
+                ? selection_grid->rows[selection_grid->selection.y].on_selection_changed(
+                      selection_grid,
+                      current_row.row_idx,
+                      &selection_grid->selection,
+                      &new_selection
+                  )
+                : true;
```
MeirGavish: In general it's good to have added NULL-checks but I'm wondering why did you need them specifically in this PR? Are you setting an `on_selection_changed` to NULL anywhere?
I did see that you set `on_key_transit` to NULL but specifically `on_selection_changed` seems like shouldn't be NULL.

graphics/high_contrast_deck_pal_gfx.png
MeirGavish, line 1: Oh, it's a single tile with one pixel per color!
I guess it's okay but I actually meant you could just make a copy of `deck_gfx.png` but with the high contrast colors and then tell grit to not load any tile data, only the palette. I checked and the following flags seem to work.
```
-gB4 -g! -m! -pn 16
```

source/card.c
```diff
@@ -19,9 +20,30 @@ const static u16 _card_sprite_lut[NUM_SUITS][NUM_RANKS] = {
     {624, 640, 656, 672, 688, 704, 720, 736, 752, 768, 784, 800, 816}
 };
 
+#define CARDS_PAL_SIZE 16
+static u16 normal_palette[CARDS_PAL_SIZE];
+static u16 high_contrast_palette[CARDS_PAL_SIZE];
```
MeirGavish: I think now these intermediate buffers are redundant and you can just use `GRIT_CPY` straight from the grit-generated buffers into `pal_obj_mem`.

MeirGavish (changes requested):
Great work! Lots of improvements on the PR.

I did see a bug where you start the game with $8 instead of $4, I'm attaching a save file that reproduces it, I couldn't reproduce it manually.
[balatro-gba.sav.zip](https://github.com/user-attachments/files/27286110/balatro-gba.sav.zip)

Otherwise, I added a few comments, I looked at some things, may have missed others, not sure whether to call this review complete or partial.

MeirGavish:
Also requesting a copilot review...

MathisMartin31:
> I did see a bug where you start the game with $8 instead of $4, I'm attaching a save file that reproduces it, I couldn't reproduce it manually.

I know where this comes from, I spotted it while working on the game save in another branch. I was loading the saved data after game vars had their default value applied, so if you had ended your previous run with $8, you would start the next also with $8 since the saved value would overwrite the default $4.

source/game/options_menu.c
MathisMartin31: I wanted to separate the logic and graphics updates, and all buttons are not handled the same way, some are activated by the arrow keys and that would split this code between the `on_key_transit` and `on_selection_changed`. Since there are a few booleans to update and keep track of I thought it would be clearer to centralize them as much as possible

source/game/options_menu.c
MathisMartin31: Yes that's what I was doing, I'll document it

source/game/options_menu.c
```diff
@@ -0,0 +1,618 @@
...
+
+static bool on_boot = true;
+static bool game_speed_changed = false;
+static bool high_contrast_changed = false;
+static bool music_volume_changed = false;
+static bool sound_volume_changed = false;
+
+static bool any_value_changed = false;
+static bool back_btn_is_save_state = false;
+static bool save_pressed = false;
```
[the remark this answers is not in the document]
MathisMartin31: Yes my mistake, `back_btn_is_save_state` is filling the role I wanted to give `save_pressed` so I'll remove it

source/save.c
```diff
@@ -0,0 +1,187 @@
...
+ ** @brief Writes a magic number and ROM version info to SRAM to signal that the
+ *         save data exists and allow the game to determine if it is compatible.
+ */
+static inline void set_save_header()
+{
+    SaveHeader check = {};
+    check.magic = CHECK_MAGIC;
+    check.dirty = is_version_dirty();
+    memcpy(&(check.githash), (void*)(&balatro_version) + GIT_HASH_START, CHECK_HASH_SIZE);
+
```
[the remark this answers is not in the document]
MathisMartin31: Tested this change, it works. I'll do that

source/save.c
```diff
@@ -0,0 +1,187 @@
...
+ ** @brief Write raw binary data to SRAM
+ *
+ * @param sram_base address written to in the SRAM
+ * @param bytes pointer to the written data
+ * @param size number of bytes written
+ */
+static inline void write_sram(u32 sram_base, const u8* bytes, u32 size)
+{
+    if (sram_base + size >= SRAM_SIZE)
+        return;
```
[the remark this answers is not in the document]
MathisMartin31: Will do

include/version.h
```diff
@@ -0,0 +1,11 @@
+/**
+ * @file version.h
+ *
+ * @brief Game ROM version global var declaration.
+ */
+#ifndef VERSION_H
+#define VERSION_H
+
+extern char balatro_version[];
```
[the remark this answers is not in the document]
MathisMartin31: Makes sense and it works

include/game/options_menu.h
```diff
@@ -0,0 +1,31 @@
+/**
+ * @file options_menu.h
+ *
+ * @brief Common functions to render UI elements.
```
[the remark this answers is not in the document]
MathisMartin31: Changed

source/game/options_menu.c
```diff
@@ -0,0 +1,618 @@
...
+            save_game();
+            change_back_save_text(true);
+            back_btn_is_save_state = false;
+            save_pressed = false;
+        }
+        else
+        {
+            game_change_state(GAME_STATE_MAIN_MENU);
+        }
+    }
```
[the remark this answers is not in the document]
MathisMartin31: Copy-paste error, will remove superfluous check

source/game/main_menu.c
```diff
@@ -75,41 +144,87 @@ void game_main_menu_on_update()
...
-            play_sfx(SFX_BUTTON, MM_BASE_PITCH_RATE, BUTTON_SFX_VOLUME);
-            game_change_state(GAME_STATE_GAME_START);
+            show_version = !show_version;
+            if (show_version)
+            {
+                tte_printf("#{P:%d,%d; cx:0x%X000}%s", 0, 152, TTE_WHITE_PB, balatro_version);
+            }
+            else
+            {
+                tte_erase_screen();
```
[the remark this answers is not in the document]
MathisMartin31: There's no text on the main menu, but I guess only erasing the Rect actually used can't hurt performance. Also, magic numbers

source/save.c
```diff
@@ -0,0 +1,187 @@
...
+ ** @brief Determines if the current build is considered "dirty" aka has uncommited changes.
+ *         This works because the balatro_version string has "-dirty" added at the end if it's
+ *         dirty.
+ *
+ * @returns true if version is dirty, false otherwise.
+ */
+static inline bool is_version_dirty()
+{
+    return strlen(balatro_version) > CHECK_HASH_SIZE;
+}
```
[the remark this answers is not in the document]
MathisMartin31: Comparing against `GIT_HASH_START + CHECK_HASH_SIZE` is what I meant to do

source/game.c
```diff
@@ -4554,9 +4558,12 @@ static void game_shop_on_exit()
     list_clear(&_shop_jokers_list);
 
     increment_blind(BLIND_STATE_DEFEATED); // TODO: Move to game_round_end()?
+
+    // test
+    save_game();
```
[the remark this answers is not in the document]
MathisMartin31: I left the test comment, but honestly I think this is a good place for an autosave. What do you think @ricfehr3 ?

source/card.c
MathisMartin31: Indeed they are, removed ^^

graphics/high_contrast_deck_pal_gfx.png
MathisMartin31: This works ^^. I think keeping the single tile image is enough, I arranged the colors in the same order as the palette in Asperite, so to preview the color swapped image you only need to copy and paste it (or to play the game ^^)

source/selection_grid.c
MathisMartin31: I'll remove this check. At some point I had NULL `on_selection_changed` methods but I ended up adding them. I agree, at least this method specifically should not be NULL

include/audio_utils.h
MathisMartin31: I named them `MM_SFX_FULL_VOLUME` and `MM_MODULE_FULL_VOLUME` so they're consistent with one another

include/game_variables.h
MathisMartin31: I agree, it's better this way

source/game/options_menu.c
MathisMartin31: I had some project for these to be highlighted whenever left/right is pressed but never got around to doing it. Maybe I should take a crack at it after all

source/game/options_menu.c
MathisMartin31: Also, having the graphics update stuff in the `on_update` method allows me to easily do a first update right after loading the menu to apply the values loaded from SRAM

source/game/options_menu.c
MathisMartin31: Well it wasn't too hard, and I think it doesn't look half bad :)

source/save.c
MathisMartin31: Yes I think this should be its own issue, I added it because the idea crossed my mind, but fully fleshing it out is starting to exceed the scope of this PR which is already big enough in its own right

MathisMartin31:
Welp, looks like Azure is down lol ><

include/game_variables.h
ricfehr3: Just wanted to note here that this will need to be made into an issue once this is merged.

include/save.h
ricfehr3: @MathisMartin31 Bumping this up . I see it's in the c file, but I think it should be moved here so it's in the summary page.

source/save.c
ricfehr3: Sounds good. We will make an issue of this as well.

source/save.c
```diff
@@ -0,0 +1,187 @@
...
+    if (!check_save_header())
+        return;
+
+    read_sram(GAME_BASE, (u8*)&g_game_vars, sizeof(g_game_vars));
+
+    // Data validation in case we start from junk data
+    if (g_game_vars.game_speed < GAME_SPEED_MIN || g_game_vars.game_speed > GAME_SPEED_MAX)
+        g_game_vars.game_speed = DEFAULT_GAME_SPEED;
+
+    g_game_vars.high_contrast = ((u8)g_game_vars.high_contrast != 1) ? false : true;
```
ricfehr3: I think this is a bit too confusing to follow. This looks like it does nothing, unless I'm missing something?

source/game/main_menu.c
```diff
@@ -58,10 +110,22 @@ void game_main_menu_on_init()
     main_menu_ace->sprite_object->ty = int2fx(MAIN_MENU_ACE_T_Y);
     main_menu_ace->sprite_object->y = main_menu_ace->sprite_object->ty;
     main_menu_ace->sprite_object->tscale = float2fx(0.8f);
-    selection_x = 0;
+    card_object_update(main_menu_ace);
+
+    // Select Play button by default, but only on boot.
+    // If we return from the options menu, we want the Options button to be highlighted.
+    static bool on_boot = true;
```
ricfehr3: I think "on_boot" is too specific here. As returning from an additional menu or restarting the game would cause issues. I think it would make more sense to have a static variable that tracks the last button position and restores it here.

source/save.c
MathisMartin31: It was just an attempt to force the value stored to be strictly 0 or 1, since we can potentially read from garbage data the first time around

ricfehr3: Oh that makes sense. I think it's safe to assume that once the header is checked we don't really need to ensure the data is valid, but just assume it is. If it gets to this point and has junk data the save itself is corrupt since it passed that check. I think you can remove this and other validity checks past the header check.

source/game/options_menu.c
MeirGavish: @MathisMartin31 Then at least separate it into functions for each option, this is too much unrelated code in a single function.

source/selection_grid.c
MeirGavish: It's okay to have this NULL-check, might be better to be defensive, I was just wondering if I missed a place where it actually is NULL.

MeirGavish (changes requested):
Looks great!
Just separate the if statements in `game_options_menu_on_update()` to functions.

source/save.c
MathisMartin31: Done ^^

source/game/options_menu.c
MathisMartin31: @MeirGavish well I separated everything into its own function, and turns out I don't need most of the static booleans anymore, and I could factorize the code for both sliders into a single function so that's waaay cleaner.

Thanks for the great suggestion ^^

source/game/main_menu.c
MathisMartin31: Done, hope it looks better now ^^

MeirGavish (approved):
Great!
Approving this on my side, I haven't looked at everything on this PR, specifically not the saving.
@ricfehr3 Once all your comments are addressed and everything in the PR looks good to you, you can merge.

source/save.c
```diff
@@ -0,0 +1,175 @@
...
+
+// clang-format off
+/**
+ * @brief Structure holding save data header info to be packed and written to SRAM for validation.
+ * Defined in this discussion as follows: https://github.com/GBALATRO/balatro-gba/discussions/450
+ * word | Byte 0 | Byte 1 | Byte 2 | Byte 3 | name         | purpose
+ * -----|--------|--------|--------|--------|--------------|------------------------------------------------------------------
+ * 0    | 0x47   | 0x42   | 0x41   | 0x4C   | MAGIC        | Identify if proceeding data is valid and not junk, spells "GBAL"
+ * 1    | Dirty  | H[0]   | H[1]   | H[2]   | GITHASH_LOW  | Dirty flag, followed by the first 3 bytes of shortened git hash H
+ * 2    | H[3]   | H[4]   | H[5]   | H[6]   | GITHASH_HIGH | Last 4 bytes of shortened git hash H, with a dirty flag
```
ricfehr3, line 40:
```suggestion
 * 2    | H[3]   | H[4]   | H[5]   | H[6]   | GITHASH_HIGH | Last 4 bytes of shortened git hash H
```

source/game/main_menu.c
```diff
@@ -9,29 +9,79 @@
...
+        {.wrap = false}
+    },
+};
+
+Button main_menu_buttons[] = {
+    {PLAY_BUTTON_OUTLINE_PID,    PLAY_BUTTON_MAIN_COLOR_PID,    play_on_pressed,    NULL},
+    {OPTIONS_BUTTON_OUTLINE_PID, OPTIONS_BUTTON_MAIN_COLOR_PID, options_on_pressed, NULL},
+};
+
+SelectionGrid main_menu_selection_grid = {
```
ricfehr3:
```suggestion
static SelectionGridRow main_menu_selection_rows[] = {
    {
        0,
        main_menu_return_row_size,
        main_menu_on_selection_changed,
        main_menu_on_key_transit,
        {.wrap = false}
    },
};

static Button main_menu_buttons[] = {
    {PLAY_BUTTON_OUTLINE_PID,    PLAY_BUTTON_MAIN_COLOR_PID,    play_on_pressed,    NULL},
    {OPTIONS_BUTTON_OUTLINE_PID, OPTIONS_BUTTON_MAIN_COLOR_PID, options_on_pressed, NULL},
};

static SelectionGrid main_menu_selection_grid = {
```

Just add static.

source/save.c
```diff
@@ -0,0 +1,175 @@
...
+    DTAG_TAROT,
+    DTAG_PLANET,
+    DTAG_VOUCHER,
+    DTAG_SKIP_TAG,
+    DTAG_END,
+    DTAG_INVALID = UNDEFINED
+};
+
+// DelimiterTag is an enum of size sizeof(u8) = 1
+#define CARDS_TAG_SIZE 1
```
ricfehr3: Should we remove this?
```suggestion
```

source/save.c
```diff
@@ -0,0 +1,175 @@
...
+ * 0    | 0x47   | 0x42   | 0x41   | 0x4C   | MAGIC        | Identify if proceeding data is valid and not junk, spells "GBAL"
+ * 1    | Dirty  | H[0]   | H[1]   | H[2]   | GITHASH_LOW  | Dirty flag, followed by the first 3 bytes of shortened git hash H
+ * 2    | H[3]   | H[4]   | H[5]   | H[6]   | GITHASH_HIGH | Last 4 bytes of shortened git hash H, with a dirty flag
+ */
+// clang-format on
+typedef struct SaveHeader
+{
+    u32 magic;
+    bool dirty;
+    char githash[7];
```
ricfehr3:
```suggestion
    char githash[CHECK_HASH_SIZE];
```

source/game/options_menu.c
```diff
@@ -0,0 +1,624 @@
...
+};
+// clang-format on
+
+const Selection OPTIONS_MENU_INIT_SEL = {0, 0};
+
+SelectionGrid options_menu_selection_grid = {
+    options_menu_selection_rows,
+    NB_OPTIONS_BUTTONS,
+    OPTIONS_MENU_INIT_SEL
+};
```
ricfehr3:
```suggestion
static SelectionGridRow options_menu_selection_rows[] = {
    {
        GAME_SPEED_BTN_IDX,
        options_menu_return_row_size,
        game_speed_row_on_selection_changed,
        NULL,
        {.wrap = true} // The wrapping is used to allow button activation with Left/Right arrows by
                       // putting logic into the `game_speed_row_on_selection_changed` function
    },
    {
        HIGH_CONTRAST_BTN_IDX,
        options_menu_return_row_size,
        regular_button_row_on_selection_changed,
        options_menu_row_on_key_transit,
        {.wrap = false}
    },
    {
        MUSIC_VOLUME_BTN_IDX,
        options_menu_return_row_size,
        music_volume_row_on_selection_changed,
        NULL,
        {.wrap = true} // Same trick as Game Speed button
    },
    {
        SOUND_VOLUME_BTN_IDX,
        options_menu_return_row_size,
        sound_volume_row_on_selection_changed,
        NULL,
        {.wrap = true} // Same trick as Game Speed button
    },
    {
        BACK_BTN_IDX,
        options_menu_return_row_size,
        regular_button_row_on_selection_changed,
        options_menu_row_on_key_transit,
        {.wrap = false}
    }
};

static Button options_menu_buttons[] = {
    {SPEED_BUTTON_OUTLINE_COLOR_PAL_IDX,    MENU_BUTTON_MAIN_COLOR_PAL_IDX, NULL,                     NULL},
    {CONTRAST_BUTTON_OUTLINE_COLOR_PAL_IDX, MENU_BUTTON_MAIN_COLOR_PAL_IDX, high_contrast_on_pressed, NULL},
    {MUSIC_BUTTON_OUTLINE_COLOR_PAL_IDX,    MENU_BUTTON_MAIN_COLOR_PAL_IDX, NULL,                     NULL},
    {SOUND_BUTTON_OUTLINE_COLOR_PAL_IDX,    MENU_BUTTON_MAIN_COLOR_PAL_IDX, NULL,                     NULL},
    {BACK_BUTTON_OUTLINE_COLOR_PAL_IDX,     BACK_BUTTON_MAIN_COLOR_PAL_IDX, back_on_pressed,          NULL}
};

static Button game_speed_buttons[] = {
    {SPEED_DOWN_BUTTON_OUTLINE_COLOR_PAL_IDX, MENU_BUTTON_MAIN_COLOR_PAL_IDX, game_speed_down_on_pressed, NULL},
    {SPEED_UP_BUTTON_OUTLINE_COLOR_PAL_IDX,   MENU_BUTTON_MAIN_COLOR_PAL_IDX, game_speed_up_on_pressed,   NULL},
};
// clang-format on

static const Selection OPTIONS_MENU_INIT_SEL = {0, 0};

static SelectionGrid options_menu_selection_grid = {
    options_menu_selection_rows,
    NB_OPTIONS_BUTTONS,
    OPTIONS_MENU_INIT_SEL
};
```

Static here

source/save.c
```diff
@@ -0,0 +1,175 @@
+#include "save.h"
```
ricfehr3, line 4: The documentation in this file is good, but doxygen isn't setup for that. Here is a patch of what I did to make it work. Feel free to apply it to this PR.

```sh
git apply fix-doxygen-for-save.patch
```

[fix-doxygen-for-save.patch](https://github.com/user-attachments/files/27308279/fix-doxygen-for-save.patch)

source/game.c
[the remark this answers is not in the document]
ricfehr3: Copilot does make a fair point here that the SRAM should really be used as little as possible. However, most people will not be using this on a physical cartridge. So I think this should stay/

ricfehr3 (changes requested):
This is looking really good. Thanks for keeping on top of this. It's almost ready to go. I left suggestions and a patch to fix an issue with doxygen. When those are addressed, I'm good to merge. Thanks!

source/save.c
MathisMartin31: Yes we can, I'll probably need them for the game save but I don't know if I'll keep them like this so I'll just get rid of them

source/save.c
ricfehr3: Oops, in that patch I reversed the `clang-format on/off`, heads up.

MathisMartin31: Yeah I just saw that ha ha

MathisMartin31:
@ricfehr3 should be good now ^^

ricfehr3:
> @ricfehr3 should be good now ^^

Nice! I'm just about ready here. One more quick thing. ~~Can you add 'B' to also return from the options menu?~~ That way we don't need to save. I'll make another issue to clean that up (maybe a greyed out save button and a "B to go back" icon), but we don't need that here. ~~But the 'B' to go back would be epic.~~

Actually. let's just leave the button for the issue. Don't want to bolt on something last minute. This is good! Merging :D

ricfehr3 (approved):

PR #451, opened 2026-04-26
Refactor/make game vars extern

ricfehr3:
After the discussion [here](https://github.com/GBALATRO/balatro-gba/pull/432#discussion_r3042810642). It's been decided that an extern was fitting for the game variables struct. This PR sets that up. I would like to merge this one asap so no one has any conflicts.

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -121,7 +121,7 @@ typedef struct ContainedHandTypes
 } ContainedHandTypes;
 // clang-format on
 
-typedef void (*GameStateCallback)(GameVariables* gs);
+typedef void (*GameStateCallback)();
 
 typedef struct
 {
--- a/include/game.h
+++ b/include/game.h
@@ -4,9 +4,9 @@
 #include "game/common_ui.h"
 #include "game_variables.h"
 
-extern GameVariables g_game_vars;
-
 #include <tonc.h>
+
+extern GameVariables g_game_vars;
 
 #define MAX_HAND_SIZE        16
 #define MAX_DECK_SIZE        52
--- a/include/game.h
+++ b/include/game.h
@@ -6,8 +6,6 @@
 
 #include <tonc.h>
 
-extern GameVariables g_game_vars;
-
 #define MAX_HAND_SIZE        16
 #define MAX_DECK_SIZE        52
 #define MAX_JOKERS_HELD_SIZE 5 // This doesn't account for negatives right now.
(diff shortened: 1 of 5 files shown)


MathisMartin31:
Since nobody uses it as an extern for now, shouldn't we wait to not create conflicts and do this with the game save PR?

ricfehr3:
> Since nobody uses it as an extern for now, shouldn't we wait to not create conflicts and do this with the game save PR?

I begin using it in the `main_menu` refactor #416 . Future work on the refactor is blocked on this because a lot of future work is dependent on how this variable is accessed.

include/game.h
```diff
@@ -6,6 +6,8 @@
 
 #include <tonc.h>
 
+extern GameVariables g_game_vars;
```
MathisMartin31: As @MeirGavish suggested, maybe move this to game_variables.h? This way we don't repeat ourselves and all files including the header have access to it.

That's what I did in my branch and it works fine

ricfehr3: That sounds good. I was thinking, since everyone will be including `game.h` usually it would make sense there. But in the `game_variables.h` header works as well and makes more sense.

Still, leaving the actual data in `game.c` makes sense to me for now.

MathisMartin31 (approved):

include/game.h
MeirGavish: I didn't mean a necessarily dedicated header, I was just making sure we don't have `extern GameVariables g_game_vars;` in every .c file, which I don't expect would have happened but I'm just a little traumatized from my previous workplace.
It's good either way.

MeirGavish (approved):
Looks good, approving.
Just make sure to run it before merging because I haven't.

include/game.h
MathisMartin31: Yeah I get that, here we have a mix of C and C++ code, with a least a dozen global vars, and `extern "C"` all over the place it's a nightmare T_T

(Yes, raising a C++ exception makes the program crash with sig-abort, what do you mean it's bad code for a billion-dollars company? 🤡)

ricfehr3:
Thanks for the reviews. I tested it and it looks good to go, I'll merge so I can merge the main menu refactor.

include/game.h
ricfehr3: Lol, I have some painful memories myself. It's telling that we are all scared when adding an `extern` :joy:

PR #454, opened 2026-05-01
sprite optimizations + static array fix

SuperbMuffin:
Moved sprite_object_update to IWRAM for faster execution since it runs every frame for every sprite. Added an early-exit to skip the update entirely when a sprite is stationary and already at its target. Also made the joker spacing lookup table static const so it stops getting copied onto the stack every single call.

I tested like 10 times to try to recreate #427 on this branch and I'm optimistic that it *might* fix it.


MathisMartin31:
Looks nice ^^
I don't remember in which conversation that was, but I remember somebody talking about counting cycles and whatnot to measure performance. I think this could be a good candidate to test that if it's ready and see how much we are actually optimizing here.

ricfehr3:
I'll take a look later and test later but looks nice! @MathisMartin31 I think you are talking about #380 .

MathisMartin31:
> I'll take a look later and test later but looks nice! @MathisMartin31 I think you are talking about #380 .

That's the one, thanks ^^

source/game/main_menu.c
```diff
@@ -63,9 +63,14 @@ void game_main_menu_on_init()
...
-    card_object_update(main_menu_ace);
     main_menu_ace->sprite_object->trotation = lu_sin((g_game_vars.timer << 8) / 2) / 3;
-    main_menu_ace->sprite_object->rotation = main_menu_ace->sprite_object->trotation;
+    card_object_update(main_menu_ace);
+    obj_aff_rotscale(
+        main_menu_ace->sprite_object->sprite->aff,
+        main_menu_ace->sprite_object->scale,
+        main_menu_ace->sprite_object->scale,
+        main_menu_ace->sprite_object->trotation
+    );
```
MeirGavish: Why make these changes?
`obj_aff_rotscale()` is already applied in `sprite_object_update()`

source/sprite.c
```diff
@@ -207,8 +207,16 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
-void sprite_object_update(SpriteObject* sprite_object)
+IWRAM_CODE void sprite_object_update(SpriteObject* sprite_object)
 {
+    if (sprite_object->vx == 0 && sprite_object->vy == 0 && sprite_object->vscale == 0 &&
+        sprite_object->vrotation == 0 && sprite_object->x == sprite_object->tx &&
+        sprite_object->y == sprite_object->ty && sprite_object->scale == sprite_object->tscale &&
+        sprite_object->rotation == sprite_object->trotation)
+    {
+        return;
+    }
```
MeirGavish: This is a lot of conditions to check, I think the entire condition needs to be extracted to an `is_sprite_object_static()` function...
Also adding this check feels a little dangerous and might be what's causing the bug but it might just require a deeper fix...

MeirGavish (changes requested):
Nice, it does seem like it solves it but there are a few issues wit this.
Namely, if you select a card and quickly move away, specifically to the buttons in the bottom, the sprite stops moving and stays stuck in an intermediate position.
image
image

source/game/main_menu.c
SuperbMuffin: It was leftover from something else I was trying, oops...

source/sprite.c
SuperbMuffin: I extracted it to a inline function, but I don't know which bug your referencing.

MeirGavish: > Namely, if you select a card and quickly move away, specifically to the buttons in the bottom, the sprite stops moving and stays stuck in an intermediate position. image image

This one

SuperbMuffin:
I fixed it.

source/sprite.c
```diff
@@ -207,8 +207,24 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
-void sprite_object_update(SpriteObject* sprite_object)
+static inline bool is_sprite_object_static(const SpriteObject* sprite_object)
 {
+    return sprite_object->vx == 0 && sprite_object->vy == 0 && sprite_object->vscale == 0 &&
+           sprite_object->vrotation == 0 && sprite_object->x == sprite_object->tx &&
+           sprite_object->y == sprite_object->ty && sprite_object->scale == sprite_object->tscale &&
+           sprite_object->rotation == sprite_object->trotation &&
+           (sprite_object->sprite == NULL ||
+            (fx2int(sprite_object->x) == sprite_object->sprite->pos.x &&
+             fx2int(sprite_object->y) == sprite_object->sprite->pos.y));
```
MeirGavish: Why this check? The `Sprite`'s position is always set to match its `SpriteObject`'s in the update function but it doesn't necessarily mean it's static, it could still be in motion.

MeirGavish (changes requested):
What I now realize about these fixes is that the code used to set a value and target value of a `SpriteObject` to the same value to move it in ways that are not defined in `sprite_object_update()` and this PR removes this ability and I'm a little hesitant about that. On the one hand it centralizes the sprite movement to `sprite_object_update()` which is a good thing, on the other it makes the types of movements less flexible.

source/sprite.c
SuperbMuffin: It checks if the Screen is actually showing the updated position.

SuperbMuffin:
It's all fixed and I updated some documentation.

source/sprite.c
```diff
@@ -208,8 +208,41 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
     sprite_object->vrotation = 0;
 }
 
-void sprite_object_update(SpriteObject* sprite_object)
+static inline bool is_sprite_object_static(const SpriteObject* sprite_object)
 {
+    // Check all velocities are zero
+    if (sprite_object->vx != 0 || sprite_object->vy != 0 || sprite_object->vscale != 0 ||
+        sprite_object->vrotation != 0)
+        return false;
```
ricfehr3: Make a new inline function `sprite_object_has_velocity`

source/sprite.c
```diff
@@ -208,8 +208,41 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
+
+    // Check all properties are at targets
+    if (sprite_object->x != sprite_object->tx || sprite_object->y != sprite_object->ty ||
+        sprite_object->scale != sprite_object->tscale ||
+        sprite_object->rotation != sprite_object->trotation)
+        return false;
```
ricfehr3: similarly, make a new function `sprite_object_at_target`

source/sprite.c
```diff
@@ -208,8 +208,41 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
+    // position has been updated to match the sprite object's internal state
+    // before declaring it static. Without this, sprite_object_update might exit
+    // early, skipping the sprite_position call and potentially causing a freeze
+    // bug where the hardware sprite is not updated to its final converged position.
+    if (sprite_object->sprite != NULL)
+    {
+        if (fx2int(sprite_object->x) != sprite_object->sprite->pos.x ||
+            fx2int(sprite_object->y) != sprite_object->sprite->pos.y)
+            return false;
+    }
```
ricfehr3:
```suggestion
```

You shouldn't need this. See PR comment.

source/sprite.c
```diff
@@ -271,13 +313,16 @@ void sprite_object_update(SpriteObject* sprite_object)
...
+    {
+        obj_aff_rotscale(
+            sprite_object->sprite->aff,
+            sprite_object->scale,
+            sprite_object->scale,
+            -sprite_object->vx + sprite_object->rotation
+        );
+        sprite_position(sprite_object->sprite, fx2int(sprite_object->x), fx2int(sprite_object->y));
+    }
 }
```
ricfehr3, line 300: All these changes here should go.

ricfehr3 (changes requested):
Hey, nice work. I got around to testing this and I have some interesting news. From what I can tell there is no need to have all the extra logic added for all sprite_object checks. It's a bit cumbersome as well, so I'd like to avoid it. What I had instead was do a minimal patch to `main` that added `IWRAM_CODE` to the call, the `static` array change, and that early `return` on a stationary sprite.

I didn't encounter any oddness with the cards staying in one spot. Here is my patch (note it needs your update to `main_menu.c` to rotate the ace still):

```
diff --git a/include/sprite.h b/include/sprite.h
index ca10438..7340d2b 100644
--- a/include/sprite.h
+++ b/include/sprite.h
@@ -236,7 +236,7 @@ void sprite_object_reset_transform(SpriteObject* sprite_object);
  *
  * @param sprite_object pointer to SpriteObject to update. Cannot be **NULL**.
  */
-void sprite_object_update(SpriteObject* sprite_object);
+IWRAM_CODE void sprite_object_update(SpriteObject* sprite_object);
 
 /**
  * @brief Shake SpriteObject on screen and play a sound
diff --git a/source/game.c b/source/game.c
index f7bc07f..74fcad6 100644
--- a/source/game.c
+++ b/source/game.c
@@ -833,7 +833,7 @@ static inline void discarded_jokers_update_loop(void)
 
 static inline void held_jokers_update_loop(void)
 {
-    const int spacing_lut[MAX_JOKERS_HELD_SIZE][MAX_JOKERS_HELD_SIZE] = {
+    static const int spacing_lut[MAX_JOKERS_HELD_SIZE][MAX_JOKERS_HELD_SIZE] = {
         {0,  0,   0,   0,   0  },
         {13, -13, 0,   0,   0  },
         {26, 0,   -26, 0,   0  },
diff --git a/source/sprite.c b/source/sprite.c
index b747a9d..924aff6 100644
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -208,8 +208,29 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
     sprite_object->vrotation = 0;
 }
 
-void sprite_object_update(SpriteObject* sprite_object)
+static inline bool is_sprite_object_static(const SpriteObject* sprite_object)
 {
+    // Check all velocities are zero
+    if (sprite_object->vx != 0 || sprite_object->vy != 0 || sprite_object->vscale != 0 ||
+        sprite_object->vrotation != 0)
+        return false;
+
+    // Check all properties are at targets
+    if (sprite_object->x != sprite_object->tx || sprite_object->y != sprite_object->ty ||
+        sprite_object->scale != sprite_object->tscale ||
+        sprite_object->rotation != sprite_object->trotation)
+        return false;
+
+    return true;
+}
+
+IWRAM_CODE void sprite_object_update(SpriteObject* sprite_object)
+{
+    if (is_sprite_object_static(sprite_object))
+    {
+        return;
+    }
+
     sprite_object->vx += ((sprite_object->tx - sprite_object->x) * g_game_vars.game_speed) / 8;
     sprite_object->vy += ((sprite_object->ty - sprite_object->y) * g_game_vars.game_speed) / 8;
``` 

Now, into something more interesting. I tested the timings by using the profiler and measuring the frame time without vsync. It looks like with the changes it actually saves about `~5000` ticks, which is about ~0.3ms, or `1/55` of a frame. That is not too significant of an improvement, but at the very least it is measurable. These measurements were also done on stationary cards to have a controlled experiment. Moving the cards will probably shoot it back up as the biggest improvement is returning early when nothing is moving.

What is also good news is that we currently run way under the vsync frame waiting time. By a longshot. That is about 280000 ticks, and we clock in 50000 ticks on `main` rn.

Here are the results:
```
# upstream rn
main-default.txt
avg = 48459.46
min = 41272
max = 55461

# fastest with all three changes
main-with-iwram-static-array-ret-early.txt
avg = 43166.18
min = 35509
max = 50358

# Only move update function to IWRAM and static array
main-with-iwram-static-array.txt
avg = 44839.12
min = 37137
max = 51614

# Only with IWRAM update
main-with-iwram.txt
avg = 45679.48
min = 38335
max = 52440

# Your branch
optimizations.txt
avg = 42757.92
min = 35371
max = 49589
```

As you can tell, it looks like my patch gets matching values to your branch. This was done by using the `print_frame_time` function in this gdb_script:

```
set logging file gdb-output.txt
set logging overwrite off

define print_ram
    set $addr = $arg0
    set $len = $arg1
    x/$lenxb $addr
end

define print_frame_tick
    enable 1
    continue
    set $i = 0
    while $i < 50
        next
        set logging enabled on
        print frame_time
        set logging enabled off
        set $j = 0
        while $j < 6
            continue
            set $j = $j + 1
        end
        set $i = $i + 1
    end
end

define print_list
    p $arg0 
end

set prompt [gbagdb]>

set style enabled on

set history save on
set print pretty on

file build/balatro-gba.elf

target remote localhost:2345

b profile_timer_breakpoint

disable 1

continue
```
And the output is processed through a script:
```sh
#!/bin/bash

awk '
{
  sum += $3
  n++

  if (n == 1 || $3 < min) min = $3
  if (n == 1 || $3 > max) max = $3
}

END {
  printf "avg = %.2f\n", sum / n
  print  "min =", min
  print  "max =", max
}
' $1
```

My comments detail what needs to be removed, but once its prepared and tested, let's save that 1/55 of a frame. 😎

source/sprite.c
SuperbMuffin: > Nice, it does seem like it solves it but there are a few issues wit this. Namely, if you select a card and quickly move away, specifically to the buttons in the bottom, the sprite stops moving and stays stuck in an intermediate position. image image

A similar bug appears when removed.

SuperbMuffin:
I have a strange 'wiggly wobbly' effect happening occasionally during gameplay I think I will add a snap to target if it's close to the goal or something similar. Actually there's probably something like that already I have to look into the code more.

MeirGavish:
@ricfehr3 Great job actually profiling it! I should start profiling performance myself instead of just eyeballing things, I managed to get your script to generally run but I need to give deeper look to #380 so I can start profiling.
Afterwards I can come back and take another look at this PR.

One thing I want to note about the numbers though, is that the 280000 cycle number is for an entire frame including both VBlank and VDraw. 
But our code doesn't run in VDraw, only in VBlank. 
In `main()` you can see our code runs by waiting for VBlank with `VBlankIntrWait()` and only then running everything, including `update()` that runs all the logic, and `draw()` being last. We don't want `draw()` to run during VDraw (probably explains why performance issues manifest as sprite flickers) so everything has to happen within VBlank, leaving us only with the 83000 or so cycles of VBlank to do everything.

It also seems like all Tonc demos work like this and never try to do anything during VDraw. I tried playing around with the order of calls in `main()` to maybe allow things to be done during VDraw but it created different sprite flicker issues.

ricfehr3:
> @ricfehr3 Great job actually profiling it! I should start profiling performance myself instead of just eyeballing things, I managed to get your script to generally run but I need to give deeper look to #380 so I can start profiling. Afterwards I can come back and take another look at this PR.
> 
> One thing I want to note about the numbers though, is that the 280000 cycle number is for an entire frame including both VBlank and VDraw. But our code doesn't run in VDraw, only in VBlank. In `main()` you can see our code runs by waiting for VBlank with `VBlankIntrWait()` and only then running everything, including `update()` that runs all the logic, and `draw()` being last. We don't want `draw()` to run during VDraw (probably explains why performance issues manifest as sprite flickers) so everything has to happen within VBlank, leaving us only with the 83000 or so cycles of VBlank to do everything.
> 
> It also seems like all Tonc demos work like this and never try to do anything during VDraw. I tried playing around with the order of calls in `main()` to maybe allow things to be done during VDraw but it created different sprite flicker issues.

No problem! Glad it's finding some use. 

For the frame timing, the **2800000**comes from the cpu clock over the framerate **16,777,216 / 59.7275**. This makes sense as the timer is on 1x prescaler and is incremented every clock.

The data I gathered was after `VBlankIntrWait()`, so an unoptimized average of 50000 is still under 83000. I'm unsure if there are spikes we are missing...? Maybe we can use the rewind feature to get a good screen tear and dig into it with the profiler?

MeirGavish:
> > @ricfehr3 Great job actually profiling it! I should start profiling performance myself instead of just eyeballing things, I managed to get your script to generally run but I need to give deeper look to #380 so I can start profiling. Afterwards I can come back and take another look at this PR.
> > One thing I want to note about the numbers though, is that the 280000 cycle number is for an entire frame including both VBlank and VDraw. But our code doesn't run in VDraw, only in VBlank. In `main()` you can see our code runs by waiting for VBlank with `VBlankIntrWait()` and only then running everything, including `update()` that runs all the logic, and `draw()` being last. We don't want `draw()` to run during VDraw (probably explains why performance issues manifest as sprite flickers) so everything has to happen within VBlank, leaving us only with the 83000 or so cycles of VBlank to do everything.
> > It also seems like all Tonc demos work like this and never try to do anything during VDraw. I tried playing around with the order of calls in `main()` to maybe allow things to be done during VDraw but it created different sprite flicker issues.
> 
> No problem! Glad it's finding some use.
> 
> For the frame timing, the **2800000**comes from the cpu clock over the framerate **16,777,216 / 59.7275**. This makes sense as the timer is on 1x prescaler and is incremented every clock.
> 
> The data I gathered was after `VBlankIntrWait()`, so an unoptimized average of 50000 is still under 83000. I'm unsure if there are spikes we are missing...? Maybe we can use the rewind feature to get a good screen tear and dig into it with the profiler?

I honestly wanted to try and get a measurement of the improvements in #444 because I think I may have missed something there...

Edit: Also the times are from here https://gbadev.net/tonc/video.html#sec-blanks

MeirGavish:
I haven't forgotten about this PR, I went on a performance profiling adventure, I want to do a few more tests, I'll update, and take another look here.

source/sprite.c
```diff
@@ -208,8 +208,47 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
+    return s->x == s->tx && s->y == s->ty && s->scale == s->tscale && s->rotation == s->trotation;
+}
+
+static inline bool is_sprite_object_static(const SpriteObject* sprite_object)
+{
+    if (sprite_object_has_velocity(sprite_object))
+        return false;
+
+    if (!sprite_object_at_target(sprite_object))
+        return false;
```
ricfehr3:
```suggestion
    if (sprite_object_has_velocity(sprite_object || !sprite_object_at_target(sprite_object))
        return false;
```

These can just be one if-statement

source/sprite.c
ricfehr3: The comment associated with this check doesn't make sense to me. All sprites are processed and updated between frames, so there is no instance where the hardware should be out of sync with the x,y position of a sprite object. 

To prove this I added a breakpoint to this check. It did in fact break, showing that there are instances where the `sprite_object`'s position is out of sync with it's `sprite`. However, this is an unrelated bug. It would only break intermittently and not often. When it did break the location of `sprite_object` to `sprite` was off by a massive amount (40323 for `sprite_object` and 157 for `sprite`. Here, the value for `sprite_object` seems invalid). Having a check for that is incorrect as it would be masking that bug.

This did make me realize how fragile having two layers of shadow registers is for the sprites position. I'll make an issue for this, but otherwise please remove the check.

source/sprite.c
```diff
@@ -219,10 +258,17 @@ void sprite_object_update(SpriteObject* sprite_object)
...
+    const FIXED epsilon_pos = (FIX_ONE >> 3); // = 1/2^3 = 0.125
+
+    // set velocity to 0 if it's close enough to the target AND we are already close to the target
+    // this prevents snapping on the first frame of animation when velocity is 0
+    if (sprite_object->vx < epsilon_vel && sprite_object->vx > -epsilon_vel &&
+        sprite_object->vy < epsilon_vel && sprite_object->vy > -epsilon_vel &&
+        (sprite_object->tx - sprite_object->x) < epsilon_pos &&
+        (sprite_object->tx - sprite_object->x) > -epsilon_pos &&
+        (sprite_object->ty - sprite_object->y) < epsilon_pos &&
+        (sprite_object->ty - sprite_object->y) > -epsilon_pos)
```
ricfehr3:
```suggestion
    if (sprite_object->vx < epsilon_vel && sprite_object->vx > -epsilon_vel &&
        sprite_object->vy < epsilon_vel && sprite_object->vy > -epsilon_vel)
```

The checks for the target position are redundant since `sprite_object_at_target` is checked first.

source/sprite.c
```diff
@@ -255,14 +301,14 @@ void sprite_object_update(SpriteObject* sprite_object)
         sprite_object->scale += sprite_object->vscale;
     }
 
-    // Set rotation to 0 if it's close enough to the target
-    if (sprite_object->vrotation < epsilon && sprite_object->vrotation > -epsilon)
+    // For rotation, prioritize snapping to target if close enough, then zero velocity.
+    if ((sprite_object->trotation - sprite_object->rotation) < epsilon_pos &&
+        (sprite_object->trotation - sprite_object->rotation) > -epsilon_pos)
```
ricfehr3: Similarly, the checks for the target position are redundant since `sprite_object_at_target` is checked first.

ricfehr3 (changes requested):
Hey, thanks for working with me on this one. Thanks for changes! I have some comments and changes requested below. Interestingly, this PR did help me find an interesting bug. @MeirGavish , maybe the `sprite_object` corrupted position has something to do with the tearing?

source/sprite.c
SuperbMuffin: I realized how stupid that comment was, I fixed the comment to clarify what it really does.

source/sprite.c
SuperbMuffin: The early return in is_sprite_object_static already handles the case where rotation == trotation, so by the time we reach this block we know they're not equal. The epsilon check here is for snapping when close but not exactly equal, so it's not redundant.

ricfehr3: Ah, gotcha. I understand now, you are correct. I do think that this could be simplified.

Since above, `sprite_object->vrotation += (sprite_object->trotation - sprite_object->rotation) / 8;`, `vrotation` is already calculated. So here, you can essentially use the old check just fine:

```c
 if (sprite_object->vrotation < epsilon && sprite_object->vrotation > -epsilon)
```

As it is an equivalent check.

But you can make it more streamlined by using `abs()` to get the absolute value:

```c
if (abs(sprite_object->vrotation) < epsilon)
```

I think keeping the old check is also preferred because you can keep just one epsilon value as we had previously for simplicity.

source/sprite.c
```diff
@@ -219,10 +251,13 @@ void sprite_object_update(SpriteObject* sprite_object)
...
-    const FIXED epsilon = (FIX_ONE >> 6); // = 1/2^6 = 0.015625
-    if (sprite_object->vx < epsilon && sprite_object->vx > -epsilon &&
-        sprite_object->vy < epsilon && sprite_object->vy > -epsilon)
+    const FIXED epsilon_vel = (FIX_ONE >> 6); // = 1/2^6 = 0.015625
+    const FIXED epsilon_pos = (FIX_ONE >> 3); // = 1/2^3 = 0.125
+
+    // set velocity to 0 if it's close enough to the target AND we are already close to the target
+    // this prevents snapping on the first frame of animation when velocity is 0
+    if (sprite_object->vx < epsilon_vel && sprite_object->vx > -epsilon_vel &&
+        sprite_object->vy < epsilon_vel && sprite_object->vy > -epsilon_vel)
```
ricfehr3:
```suggestion
    if (abs(sprite_object->vx) < epsilon && abs(sprite_object->vy) < epsilon)
```

Similar to below, use `abs()` to simplify

source/sprite.c
```diff
@@ -208,8 +208,40 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
+
+    return true;
+}
+
+IWRAM_CODE void sprite_object_update(SpriteObject* sprite_object)
+{
+    if (is_sprite_object_static(sprite_object))
+    {
+        return;
+    }
```
ricfehr3:
```suggestion
    if (is_sprite_object_static(sprite_object))
        return;
```

nitpick: Remove braces around one-line `return`

source/game.c
```diff
@@ -3033,10 +3354,12 @@ static inline void cards_in_hand_update_loop(void)
                     {
                         hand_y -= int2fx(CARD_FOCUSED_SEL_Y);
                     }
-
                     if (i != selected_card_idx && hand[i]->sprite_object->y > hand_y)
                     {
                         hand[i]->sprite_object->y = hand_y;
+                        hand[i]->sprite_object->ty =
+                            hand_y; // Set target y to match y. Ensures target is updated
+                                    // even when vy becomes 0, preventing immediate snap back.
```
ricfehr3:
```suggestion
                        // Set target y to match y. Ensures target is updated even when vy becomes 0, preventing immediate snap back.
                        hand[i]->sprite_object->ty = hand_y;
```

Put the comment above to let comment look prettier with clang-format

ricfehr3 (changes requested):
Looking good. Checking again this time around it looks like some of the conditionals for the card movement could be simplified. I think we should go back to using the older checks since they were simpler with only one `epsilon`, but they are proven to already have good values. See the comment for details. Thanks!

source/sprite.c
```diff
@@ -208,8 +208,38 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
+    if (sprite_object_has_velocity(sprite_object) || !sprite_object_at_target(sprite_object))
+        return false;
+
+    // Ensure sprite_position() has been called with the final position.
+    if (sprite_object->sprite != NULL)
+    {
+        if (fx2int(sprite_object->x) != sprite_object->sprite->pos.x ||
+            fx2int(sprite_object->y) != sprite_object->sprite->pos.y)
+            return false;
+    }
```
MeirGavish: So this check is more specific to `sprite_object_update()`, isn't it? I think you should separate this to another function, and call it from `sprite_object_update()` rather than here, static should just check if it's static and be generic.

source/sprite.c
```diff
@@ -219,10 +249,13 @@ void sprite_object_update(SpriteObject* sprite_object)
...
 
-    // set velocity to 0 if it's close enough to the target
-    const FIXED epsilon = (FIX_ONE >> 6); // = 1/2^6 = 0.015625
-    if (sprite_object->vx < epsilon && sprite_object->vx > -epsilon &&
-        sprite_object->vy < epsilon && sprite_object->vy > -epsilon)
+    const FIXED epsilon_vel = (FIX_ONE >> 6); // = 1/2^6 = 0.015625
+    const FIXED epsilon_pos = (FIX_ONE >> 3); // = 1/2^3 = 0.125
+
+    // set velocity to 0 if it's close enough to the target AND we are already close to the target
+    // this prevents snapping on the first frame of animation when velocity is 0
```
MeirGavish: Is this comment still relevant?

MeirGavish (changes requested):
Just a few comments about some things I saw in the code, not a very thorough review.

SuperbMuffin:
I addressed all the comments!

MeirGavish:
> Hey, thanks for working with me on this one. Thanks for changes! I have some comments and changes requested below. Interestingly, this PR did help me find an interesting bug. @MeirGavish , maybe the `sprite_object` corrupted position has something to do with the tearing?

This is a good point, I do suspect that the tearing in #427 is caused by a deeper issue rather than just performance and may require a deeper investigation. And I agree that having a position for both `sprite` and `sprite_object` that needs to be kept in sync is pretty bad...

source/sprite.c
```diff
@@ -208,8 +208,40 @@ void sprite_object_reset_transform(SpriteObject* sprite_object)
...
+{
+    return s->x == s->tx && s->y == s->ty && s->scale == s->tscale && s->rotation == s->trotation;
+}
+
+static inline bool is_sprite_object_static(const SpriteObject* sprite_object)
+{
+    return !sprite_object_has_velocity(sprite_object) && sprite_object_at_target(sprite_object);
+}
+
+static inline bool sprite_object_needs_position_sync(const SpriteObject* sprite_object)
```
ricfehr3: It looks like this is a band-aid around the sprite to sprite_object position issue we have. Since it doesn't actually solve the problem and (as far as I can tell from testing) doesn't effect the experience of the game, I think we should remove it and the check below. Unless I'm missing something?

ricfehr3 (changes requested):
One more comment.

SuperbMuffin:
Okay, I finally figured out why I needed that *crude* band-aid. Whenever you touched SpriteObject's coordinates you needed to run sprite_position. 3 times in game.c this didn't happen. Adding a sprite_position call to all those spots fixed the actual problem.

ricfehr3 (approved):
Awesome! Good work finding where the issue was. I think we should still have an issue to rework how we handle the positions between `sprite` and `sprite_object`. I think your fixes now are good until we can take on that bigger task. I'm approving for now, I'll leave it up for a bit so anyone else can check if they'd like. If there's no more requests for changes, I'll merge. Thanks for working with us on this PR @SuperbMuffin , it was a long one 😄

ricfehr3:
@SuperbMuffin I'm ready to merge this one, and sorry I did just add a conflict merging #475 . Once you can resolve it I'll merge, if you don't mind me resolving it, I should be able to do it in github easily. Just let me know. Thanks!

SuperbMuffin:
Awesome, sorry for taking so long to just resolve a merge conflict lol. I resolved it now so it should be good to go!

ricfehr3 (approved):
Great! Thanks for resolving that conflict @SuperbMuffin . Also, thanks for all the changes and keeping up with this long PR. Glad to get it merged :)

PR #461, opened 2026-05-03
Refactor/blind select screen

ricfehr3:
New branch to take over #432 

This PR is a refactor for the blind select screen. It pulls the logic out of `game.c` and tries to sanely apply fixes. Though, some will need to be pushed to other PRS as they are too large of changes. Such as replacing the precious block moving code with helper functions. I don't want to get too lost in modifying logic just yet, right now I'm just looking to pull the logic out of `game.c` without breaking anything or too huge of fixes.

### DIFF
--- a/include/game/blind_select.h
+++ b/include/game/blind_select.h
@@ -1,0 +1,31 @@
+/**
+ * @file blind_select.h
+ *
+ * @brief Blind select state functions
+ */
+#ifndef GAME_BLIND_SELECT_H
+#define GAME_BLIND_SELECT_H
+
+#include "game_variables.h"
+
+/**
+ * @brief Change to the blind select background
+ */
+void game_blind_select_change_background(void);
+
+/**
+ * @brief Blind select state initialization
+ */
+void game_blind_select_on_init(void);
+
+/**
+ * @brief Blind select state update
+ */
+void game_blind_select_on_update(void);
+
+/**
+ * @brief Blind select cleanup
+ */
+void game_blind_select_on_exit(void);
+
+#endif // GAME_BLIND_SELECT_H
--- a/include/game/blind_select.h
+++ b/include/game/blind_select.h
@@ -6,6 +6,8 @@
 #ifndef GAME_BLIND_SELECT_H
 #define GAME_BLIND_SELECT_H
 
+#include "blind.h"
+
 /**
  * @brief Change to the blind select background
  */
@@ -25,5 +27,7 @@ void game_blind_select_on_update(void);
  * @brief Blind select cleanup
  */
 void game_blind_select_on_exit(void);
+
+void increment_blind(enum BlindState increment_reason);
 
 #endif // GAME_BLIND_SELECT_H
--- a/include/game/blind_select.h
+++ b/include/game/blind_select.h
@@ -6,8 +6,6 @@
 #ifndef GAME_BLIND_SELECT_H
 #define GAME_BLIND_SELECT_H
 
-#include "game_variables.h"
-
 /**
  * @brief Change to the blind select background
  */
--- a/include/game_variables.h
+++ b/include/game_variables.h
@@ -8,6 +8,8 @@
 
 #include <tonc.h>
 
+#include "blind.h"
+
 #define GAME_SPEED_MIN 1
 #define GAME_SPEED_MAX 4
 
@@ -40,6 +42,10 @@ typedef struct
     int round;
     int ante;
     int money;
+
+    // Blind variables
+
+    enum BlindType current_blind;
 
     // Options variables
 
--- a/include/game_variables.h
+++ b/include/game_variables.h
@@ -46,6 +46,7 @@ typedef struct
     // Blind variables
 
     enum BlindType current_blind;
+    enum BlindType next_boss_blind;
 
     // Options variables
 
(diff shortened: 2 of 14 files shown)


include/game/blind_select.h
```diff
@@ -0,0 +1,33 @@
...
+ * @brief Blind select state update
+ */
+void game_blind_select_on_update(void);
+
+/**
+ * @brief Blind select cleanup
+ */
+void game_blind_select_on_exit(void);
+
+void increment_blind(enum BlindState increment_reason);
```
ricfehr3, line 31: This is a bit of a band aid. It's only called at the end of the `game.c` shop. I don't think it should be called there, but breaking that logic right now is not something I want to introduce just yet.

source/game/blind_select.c
```diff
@@ -0,0 +1,640 @@
...
+                    &se_mem[MAIN_BG_SBB][x_to + 32 * y_to],
+                    &se_mem[MAIN_BG_SBB][x_from + 32 * y_from],
+                    3
+                );
+                break;
+            }
+            default:
+                break;
+        }
+    }
```
ricfehr3, line 636: This is one chunk that I think is outside of this PR that we should address later.

include/game_variables.h
```diff
@@ -41,6 +43,13 @@ typedef struct
     int ante;
     int money;
 
+    // Blind variables
+
+    enum BlindType current_blind;
+    enum BlindType next_boss_blind;
+    enum BlindState blinds_states[NUM_BLINDS_PER_ANTE];
+    bool boss_rolled_this_ante;
```
MathisMartin31: When adding new stuff to this struct, be sure to check its new size and if it fits in its allocated place in the save memory. I think we should a comment to this effect

source/game/common_ui.c
```diff
@@ -15,13 +16,18 @@ static const BackgroundRenderCallback bgCallbacks[] = {
...
     [BG_OPTIONS_MENU] = game_options_menu_change_background,
     [BG_MAIN_MENU] = game_main_menu_change_background,
 };
 
-void change_background(enum BackgroundId id)
+void change_background(enum BackgroundId id, bool force_redraw)
 {
+    if (force_redraw)
+    {
+        background = BG_NONE;
```
MathisMartin31, line 28: ~It seems that `reset_background()` already sets background to `BG_NONE`, so I don't know which one but a line here is redundant.~
Hold on I think it's because there are two `background` variables, one here and one in game.c? Maybe we should put a comment somewhere that we'll need to remove the latter during the refactor

source/game.c
```diff
@@ -423,14 +375,11 @@ static const Rect GAME_WIN_MSG_TEXT_RECT    = {112,      72,     UNDEFINED, UNDE
...
 static const BG_POINT CARD_DRAW_POS         = {208,     110};
-static const BG_POINT CUR_BLIND_TOKEN_POS   = {8,       18};
 static const BG_POINT CARD_DISCARD_PNT      = {240,     70};
 static const BG_POINT HAND_START_POS        = {120,     90};
 static const BG_POINT HAND_PLAY_POS         = {120,     70};
 // clang-format on
 
-typedef void (*SubStateActionFn)(void);
-
 static enum BackgroundId background = BG_NONE;
```
MathisMartin31: Maybe add a comment saying this will need to be removed in favor of the one in common_ui.c ?

source/game.c
```diff
@@ -1442,7 +1316,8 @@ static void bg_copy_current_item_to_top_left_panel(void)
 
 // Resets bottom row bg tiles of the top left panel (shop/blind) after
 // it is dismissed to match the rest of the game panel background.
-static inline void reset_top_left_panel_bottom_row()
+// TODO: this is shared in multiple files now, probably shouldn't be
```
MathisMartin31: It seems to be only used in blind_select.c, if I didn't miss something? Maybe this could be moved over there or blind.c

source/game/common_ui.c
ricfehr3: We should, it's a temporary situation. I'll leave a not of it with the legacy change background call as well.

include/game_variables.h
ricfehr3: Sounds good. Are we limited too hard in any way on that? I thought we had 32KB, which should be plenty. Still, that doesn't mean we should be careless, so I'll note it.

source/game.c
ricfehr3: It's used in `game.c` as well. I'm going to move it to `graphic_utils` as it looks like it should belong there.

include/game_variables.h
MathisMartin31: Oh no I meant that I currently set the Header address to 0x00, the GameVariables addres to 0x10 and the variable list encoding address to 0x60 so we have like, 80 bytes for the GameVariables ^^'

But I realize it's a bit silly, I will remove this and write everything sequentially since we have that Header, if it checks out we know we can read from memory with no issues.

ricfehr3:
I think this one is good now. I will leave it up to get some looks in, but will be merging tomorrow and moving onto #468 .

MathisMartin31:
I'll correct a small issue with my game over screen PR, I noticed that with your changes, when skipping a blind, there is no button highlighted

Issue #452, opened 2026-04-27
[BUG] Seed RNG on every run


MeirGavish:
Since the RNG is seeded in `game_start()` which is only called once when starting the game, the RNG is only seeded once globally and not per run.
This isn't an issue currently so it's not marked with a bug label but it needs to be seeded per run for or as part of #142, so I'll make this issue blocking.
The issue with just seeding the RNG again on reset is that the seed is only changed in the main menu so doing that will actually use the same seed every time and will probably make the reset just repeat the same run.

_Originally posted by @MeirGavish in https://github.com/GBALATRO/balatro-gba/pull/416#discussion_r3110092331_

MathisMartin31:
@MeirGavish From my testing this is no longer the case as per #449

MeirGavish:
> [@MeirGavish](https://github.com/MeirGavish) From my testing this is no longer the case as per [#449](https://github.com/GBALATRO/balatro-gba/pull/449)

This is still an issue. RNG is only seeded once when `game_start()` is called when starting from the main menu. I'm editing the issue so the actual problem I realized is clearer.

MathisMartin31:
I think this can be done along with #452, once the whole RNG system is its own file we would have access to functions to seed it in places other than the main menu

MeirGavish:
> I think this can be done along with [#452](https://github.com/GBALATRO/balatro-gba/issues/452), once the whole RNG system is its own file we would have access to functions to seed it in places other than the main menu

Along with #142? Yes, they can and probably should be done together.

MathisMartin31:
Oh I completely forgot about this third issue, or course being able to choose the seed is important ^^'

I think the easiest way to implement that would be to have the "Play" button in the main menu lead the player to a menu when they could choose their run settings, like resume the previous run, or start a new one with a deck and seed (optional)

PR #468, opened 2026-05-03
Refactor/game over screen

### LINKED ISSUE #463: Create file for game over screen `game_over.h`

MathisMartin31:
Closes #463 

I just basically moved everything I could to the game_over.c file.

Only caveat though, and I don't know what to do with that, is the `game_over_on_exit()` function. There is a loooot of cleaning of structures that are exclusive to game.c, like jokers lists etc.

I don't want to move those to the GameVariables struct since we don't want to dump these lists to the save file, we can just encode some minimal stuff (see #450), but manipulating them in game_over.c would mean calling a lot of getters and I don't think it's very clean. @ricfehr3 what do you think about this?

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -12,6 +12,7 @@
 #include "card.h"
 #include "game/blind_select.h"
 #include "game/common_ui.h"
+#include "game/game_over.h"
 #include "game/main_menu.h"
 #include "game/options_menu.h"
 #include "game_variables.h"
@@ -44,9 +45,6 @@
 
 // SE sizes
 #define ROUND_END_BLACK_PANEL_INIT_BOTTOM_SE 12
-
-// TODO: Properly define and use
-#define GAME_OVER_ANIM_FRAMES 15
 
 #define SHOP_LIGHTS_1_CLR 0xFFFF
 #define SHOP_LIGHTS_2_CLR 0x32BE
@@ -184,11 +182,6 @@ static void game_round_end_on_update(void);
 static void game_round_end_on_exit(void);
 static void game_shop_on_update(void);
 static void game_shop_on_exit(void);
-static void game_lose_on_init(void);
-static void game_lose_on_update(void);
-static void game_over_on_exit(void);
-static void game_win_on_init(void);
-static void game_win_on_update(void);
 static void game_shop_intro(void);
 static void game_shop_process_user_input(void);
 static void game_shop_outro(void);
@@ -211,14 +204,12 @@ static void display_discards(int value);
 static void set_hand(void);
 static int deck_get_size(void);
 static int deck_get_max_size(void);
-static void game_over_init(void);
 static bool check_and_score_joker_for_event(
     ListItr* starting_joker_itr,
     CardObject* card_object,
     enum JokerEvent joker_event
 );
 static int calculate_interest_reward(void);
-static void game_over_anim_frame(void);
 
 static void game_playing_discard_on_pressed(void);
 static void game_playing_execute_discard(void);
@@ -317,11 +308,6 @@ static const Rect BIG_BLIND_TITLE_SRC_RECT  = {0,       26,     8,      26 };
 static const Rect BOSS_BLIND_TITLE_SRC_RECT = {0,       27,     8,      27 };
 static const Rect CASHOUT_DEST_RECT =         {10,      8,      23,     10 };
 static const BG_POINT CASHOUT_SRC_3X3_RECT_POS =   {5,  29};
-static const BG_POINT GAME_OVER_SRC_RECT_3X3_POS = {25, 29};
-static const Rect GAME_OVER_DIALOG_DEST_RECT= {11,      21,      23,     28};
-static const Rect GAME_OVER_ANIM_RECT       = {11,      8,       23,     28};
-static const BG_POINT NEW_RUN_BTN_DEST_POS  = {15,      26};
-static const Rect NEW_RUN_BTN_SRC_RECT      = {0,       30,      4,      31};
 static const BG_POINT ROUND_END_REWARDS_ELLIPSIS_POS = {10, 13};
 
 // Flaming score animation frames
@@ -363,9 +349,6 @@ static const Rect ROUND_END_BLIND_REQ_RECT  = {104,     96,     136,       UNDEF
 static const Rect ROUND_END_BLIND_REWARD_RECT = { 168,  96,     UNDEFINED, UNDEFINED };
 static const Rect CASHOUT_TEXT_RECT         = {88,      72,     UNDEFINED, UNDEFINED };
 static const Rect SHOP_REROLL_RECT          = {88,      96,     UNDEFINED, UNDEFINED };
-static const Rect GAME_LOSE_MSG_TEXT_RECT   = {104,     72,     UNDEFINED, UNDEFINED};
-// 1 character to the right of GAME_LOSE
-static const Rect GAME_WIN_MSG_TEXT_RECT    = {112,      72,     UNDEFINED, UNDEFINED};
 
 static const BG_POINT HELD_JOKERS_POS       = {108,     10};
 static const BG_POINT JOKER_DISCARD_TARGET  = {240,     30};
@@ -695,6 +678,54 @@ void game_init()
     // Initialize/reset unbeaten Boss/Showdown Blinds so they are all available
     init_unbeaten_blinds_list(false);
     init_unbeaten_blinds_list(true);
+}
+
+// This function isn't set in stone. This is just a placeholder
+// allowing the player to restart the game. Thought it would be nice to have
+// util we decide what we want to do after a game over.
+void game_clear()
+{
+    while (list_get_len(&_owned_jokers_list) > 0)
+    {
+        JokerObject* joker_object = list_get_at_idx(&_owned_jokers_list, 0);
+        remove_owned_joker(0);
+        joker_object_destroy(&joker_object);
+    }
+
+    tte_erase_screen();
+
+    // For some reason that I haven't figured out yet,
+    // if I don't destroy the blind tokens they won't
+    // show up on the next run.
+    sprite_destroy(&playing_blind_token);
+    sprite_destroy(&round_end_blind_token);
+
+    list_clear(&_owned_jokers_list);
+    list_clear(&_discarded_jokers_list);
+    list_clear(&_expired_jokers_list);
+    list_clear(&_shop_jokers_list);
+
+    game_init();
+
+    display_round();
+    display_score(score);
+    display_chips();
+    display_mult();
+    display_hands(hands);
+    display_discards(discards);
+    display_money();
+    // Ante
+    tte_printf(
+        "#{P:%d,%d; cx:0x%X000}%d#{cx:0x%X000}/%d",
+        ANTE_TEXT_RECT.left,
+        ANTE_TEXT_RECT.top,
+        TTE_YELLOW_PB,
+        g_game_vars.ante,
+        TTE_WHITE_PB,
+        MAX_ANTE
+    );
+
+    affine_background_load_palette(affine_background_gfxPal);
 }
 
 static inline void discarded_jokers_update_loop(void)
@@ -1458,6 +1489,11 @@ void change_background_legacy(enum BackgroundId id)
     background_legacy = id;
 }
 
+void reset_background(void)
+{
+    background_legacy = BG_NONE;
+}
+
 static void display_temp_score(u32 value)
 {
     char temp_score_str_buff[UINT_MAX_DIGITS + 1];
@@ -1680,28 +1716,6 @@ static void game_round_on_init(void)
      * otherwise or for the buttons.
      */
     game_playing_selection_grid.selection = GAME_PLAYING_INIT_SEL;
-}
-
-static void game_over_init(void)
-{
-    // Clears the round end menu
-    main_bg_se_clear_rect(POP_MENU_ANIM_RECT);
-    main_bg_se_copy_expand_3x3_rect(GAME_OVER_DIALOG_DEST_RECT, GAME_OVER_SRC_RECT_3X3_POS);
-    main_bg_se_copy_rect(NEW_RUN_BTN_SRC_RECT, NEW_RUN_BTN_DEST_POS);
-}
-
-static void game_lose_on_init(void)
-{
-    game_over_init();
-    // Using the text color to match the "Game Over" text
-    affine_background_set_color(TEXT_CLR_RED);
-}
-
-static void game_win_on_init(void)
-{
-    game_over_init();
-    // Using the text color to match the "You Win" text
-    affine_background_set_color(TEXT_CLR_BLUE);
 }
 
 // General functions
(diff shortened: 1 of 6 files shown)


MeirGavish:
I think it does make sense to move these to GameVariables, the whole purpose of the struct is to expose globals to the different state files. In general when implementing a save you do want to save the list of jokers for example... I think that needs to be figured out when saving is implemented and prioritize the main purpose of the GaemVariables struct.

MeirGavish:
@ricfehr3 I probably won't review this, do you want me to request a review from Copilot?

MathisMartin31:
> I think it does make sense to move these to GameVariables, the whole purpose of the struct is to expose globals to the different state files. In general when implementing a save you do want to save the list of jokers for example... I think that needs to be figured out when saving is implemented and prioritize the main purpose of the GaemVariables struct.

Problem is that these lists hold pointers to data that will not exist when the struct is loaded back, and the GameVariables are dumped as is when saving right now, so it will only take up space for nothing. For now, for the shop screen, I used getters and it works fine but this will have to be a conversation at some point.

MeirGavish:
> > I think it does make sense to move these to GameVariables, the whole purpose of the struct is to expose globals to the different state files. In general when implementing a save you do want to save the list of jokers for example... I think that needs to be figured out when saving is implemented and prioritize the main purpose of the GaemVariables struct.
> 
> Problem is that these lists hold pointers to data that will not exist when the struct is loaded back, and the GameVariables are dumped as is when saving right now, so it will only take up space for nothing. For now, for the shop screen, I used getters and it works fine but this will have to be a conversation at some point.

I don't think you should dump the GameVariables struct as is, the struct should serve a single purpose, when it doubles as a save struct and everything is dumped by default it creates these issues and bugs. It also means that right now you are already dumping data that is not loaded and shouldn't be there (yet at least). I think you should create a separate `SaveData` struct for this purpose and should live only when loading or saving and be dumped/loaded and copied between the `GameVariables` struct in a single function. Probably should also make separate structs and functions for the settings and the future game save data.

ricfehr3:
> @ricfehr3 I probably won't review this, do you want me to request a review from Copilot?

Yes, could you please request a copilot review. I guess I still can't.

ricfehr3:
> > > I think it does make sense to move these to GameVariables, the whole purpose of the struct is to expose globals to the different state files. In general when implementing a save you do want to save the list of jokers for example... I think that needs to be figured out when saving is implemented and prioritize the main purpose of the GaemVariables struct.
> > 
> > 
> > Problem is that these lists hold pointers to data that will not exist when the struct is loaded back, and the GameVariables are dumped as is when saving right now, so it will only take up space for nothing. For now, for the shop screen, I used getters and it works fine but this will have to be a conversation at some point.
> 
> I don't think you should dump the GameVariables struct as is, the struct should serve a single purpose, when it doubles as a save struct and everything is dumped by default it creates these issues and bugs. It also means that right now you are already dumping data that is not loaded and shouldn't be there (yet at least). I think you should create a separate `SaveData` struct for this purpose and should live only when loading or saving and be dumped/loaded and copied between the `GameVariables` struct in a single function. Probably should also make separate structs and functions for the settings and the future game save data.

Hmm.. maybe we should make an issue for it now? Seems like a big can of worms.

source/game/game_over.c
```diff
@@ -0,0 +1,111 @@
...
+#include "game.h"
+#include "game/main_menu.h"
+#include "graphic_utils.h"
+#include "layout.h"
+#include "soundbank.h"
+#include "util.h"
+
+#include <tonc.h>
+
+#define GAME_OVER_ANIM_FRAMES 15
```
ricfehr3: This should be a `static const u32` since to keep it scoped to this file

source/game/game_over.c
```diff
@@ -0,0 +1,111 @@
...
+    if (key_hit(SELECT_CARD))
+    {
+        play_sfx(SFX_BUTTON, MM_BASE_PITCH_RATE, BUTTON_SFX_VOLUME);
+        game_change_state(GAME_STATE_BLIND_SELECT);
+    }
+}
+
+void game_lose_on_update(void)
+{
+    if (g_game_vars.timer < GAME_OVER_ANIM_FRAMES)
```
ricfehr3: I think you should keep this as it is now, but I think we should have a timer that's used only in this file. Since it's just a u32. 

You can either leave it as `g_game_vars.timer` or make a `static u32 timer;` variable that you manually increment in the update loop here. It's a simple enough variable that I feel comfortable leaving it in now and removing later. 

I just don't think we should use the global timer unless it's needed. Since it could be modified outside of this file. It's better practice to just keep it local.

ricfehr3 (changes requested):
Nice work, thank you again for helping out with the refactor. This is great. I left a couple of comments. The interesting one being about the timer. I'll let you decide what to do with it. 

Once you address that and the copilot review, I think this one is good.

MeirGavish:
> > @ricfehr3 I probably won't review this, do you want me to request a review from Copilot?
> 
> Yes, could you please request a copilot review. I guess I still can't. 

@ricfehr3 You need a paid Copilot subscription for that feature...
For now I have one but I might cancel it, especially because Microsoft/Github are changing the pricing scheme.

source/game.c
```diff
@@ -809,6 +792,57 @@ void game_init()
...
+    sprite_destroy(&playing_blind_token);
+    sprite_destroy(&round_end_blind_token);
+    sprite_destroy(&blind_select_tokens[SMALL_BLIND]);
+    sprite_destroy(&blind_select_tokens[BIG_BLIND]);
+    sprite_destroy(&blind_select_tokens[BOSS_BLIND]);
+
+    list_clear(&_owned_jokers_list);
+    list_clear(&_discarded_jokers_list);
+    list_clear(&_expired_jokers_list);
+    list_clear(&_shop_jokers_list);
```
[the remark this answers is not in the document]
MeirGavish: Owned jokers are handled. Discarded and expired jokers should be empty because they are short lived only during the animation. Shop jokers should be empty since we are not in the shop. The only concern is if `game_clear()` is called from states other than the game over and that should be documented in a comment. But other changes to this are outside of the scope of this refactor.

source/game.c
```diff
@@ -809,6 +792,57 @@ void game_init()
     blind_tokens_init();
 }
 
+// This function isn't set in stone. This is just a placeholder
+// allowing the player to restart the game. Thought it would be nice to have
+// util we decide what we want to do after a game over.
+void game_clear()
```
MeirGavish: Also, I say rename it to `game_reset`, it's a clearer name

MeirGavish (changes requested):

source/game/game_over.c
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @brief Polls user input to detect "Retry" button press.
+ */
+static inline void game_over_process_user_input()
+{
+    if (key_hit(SELECT_CARD))
+    {
+        play_sfx(SFX_BUTTON, MM_BASE_PITCH_RATE, BUTTON_SFX_VOLUME);
+        game_change_state(GAME_STATE_BLIND_SELECT);
```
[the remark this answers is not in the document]
MeirGavish: No, the reset happens in `game_over_on_exit()` and `game_change_state()` is the correct thing to do here. The issue with reseeding the RNG is documented in #452.

MeirGavish:
> > > > I think it does make sense to move these to GameVariables, the whole purpose of the struct is to expose globals to the different state files. In general when implementing a save you do want to save the list of jokers for example... I think that needs to be figured out when saving is implemented and prioritize the main purpose of the GaemVariables struct.
> > > 
> > > 
> > > Problem is that these lists hold pointers to data that will not exist when the struct is loaded back, and the GameVariables are dumped as is when saving right now, so it will only take up space for nothing. For now, for the shop screen, I used getters and it works fine but this will have to be a conversation at some point.
> > 
> > 
> > I don't think you should dump the GameVariables struct as is, the struct should serve a single purpose, when it doubles as a save struct and everything is dumped by default it creates these issues and bugs. It also means that right now you are already dumping data that is not loaded and shouldn't be there (yet at least). I think you should create a separate `SaveData` struct for this purpose and should live only when loading or saving and be dumped/loaded and copied between the `GameVariables` struct in a single function. Probably should also make separate structs and functions for the settings and the future game save data.
> 
> Hmm.. maybe we should make an issue for it now? Seems like a big can of worms.

Yeah, if it's a lot of work to create an intermediate struct for saving, it can go in another issue.

source/game/game_over.c
MathisMartin31: What's the pros of it being a static const variable instead of a define evaluated at compile?

source/game/game_over.c
MathisMartin31: Yeah I'll make a timer for this file it makes sense

MeirGavish:
> > > > > I think it does make sense to move these to GameVariables, the whole purpose of the struct is to expose globals to the different state files. In general when implementing a save you do want to save the list of jokers for example... I think that needs to be figured out when saving is implemented and prioritize the main purpose of the GaemVariables struct.
> > > > 
> > > > 
> > > > Problem is that these lists hold pointers to data that will not exist when the struct is loaded back, and the GameVariables are dumped as is when saving right now, so it will only take up space for nothing. For now, for the shop screen, I used getters and it works fine but this will have to be a conversation at some point.
> > > 
> > > 
> > > I don't think you should dump the GameVariables struct as is, the struct should serve a single purpose, when it doubles as a save struct and everything is dumped by default it creates these issues and bugs. It also means that right now you are already dumping data that is not loaded and shouldn't be there (yet at least). I think you should create a separate `SaveData` struct for this purpose and should live only when loading or saving and be dumped/loaded and copied between the `GameVariables` struct in a single function. Probably should also make separate structs and functions for the settings and the future game save data.
> > 
> > 
> > Hmm.. maybe we should make an issue for it now? Seems like a big can of worms.
> 
> Yeah, if it's a lot of work to create an intermediate struct for saving, it can go in another issue.

Though, it does need to be done for the refactor so saving considerations don't block exposing variables through the struct.

MathisMartin31:
> > > > > > I think it does make sense to move these to GameVariables, the whole purpose of the struct is to expose globals to the different state files. In general when implementing a save you do want to save the list of jokers for example... I think that needs to be figured out when saving is implemented and prioritize the main purpose of the GaemVariables struct.
> > > > > 
> > > > > 
> > > > > Problem is that these lists hold pointers to data that will not exist when the struct is loaded back, and the GameVariables are dumped as is when saving right now, so it will only take up space for nothing. For now, for the shop screen, I used getters and it works fine but this will have to be a conversation at some point.
> > > > 
> > > > 
> > > > I don't think you should dump the GameVariables struct as is, the struct should serve a single purpose, when it doubles as a save struct and everything is dumped by default it creates these issues and bugs. It also means that right now you are already dumping data that is not loaded and shouldn't be there (yet at least). I think you should create a separate `SaveData` struct for this purpose and should live only when loading or saving and be dumped/loaded and copied between the `GameVariables` struct in a single function. Probably should also make separate structs and functions for the settings and the future game save data.
> > > 
> > > 
> > > Hmm.. maybe we should make an issue for it now? Seems like a big can of worms.
> > 
> > 
> > Yeah, if it's a lot of work to create an intermediate struct for saving, it can go in another issue.
> 
> Though, it does need to be done for the refactor so saving considerations don't block exposing variables through the struct.

I can do it now if it's easier that way. I don't think it's gonna be too much trouble. It might actually help address some stuff like saving Jokers

source/game.c
MathisMartin31: I agree, done ^^

ricfehr3:
Thanks for updating the Save struct @MathisMartin31 ! I wanted to ask, could you cherry-pick the commit for it into a new branch and open a new PR. I'd like to merge this and put some more thought into the save file itself. If that's good for you.

MathisMartin31:
> Thanks for updating the Save struct @MathisMartin31 ! I wanted to ask, could you cherry-pick the commit for it into a new branch and open a new PR. I'd like to merge this and put some more thought into the save file itself. If that's good for you.

Yes it's fine, I'll do it when I have a moment ^^

source/game/blind_select.c
```diff
@@ -222,6 +222,7 @@ static void game_blind_select_handle_input()
                     }
 
                     game_blind_select_print_blinds_reqs_and_rewards();
+                    highlight_select_button();
```
ricfehr3, line 225: Thanks for catching this!

source/game/game_over.c
ricfehr3: It ensures the variable is both scoped to this file only, and has a type. It's just more specific really. In how it's used in this file, the compiler should treat it no differently, but it is cleaner imo.

source/game/game_over.c
```diff
@@ -0,0 +1,116 @@
...
+    else if (timer == GAME_OVER_ANIM_FRAMES)
+    {
+        tte_printf(
+            "#{P:%d,%d; cx:0x%X000}GAME OVER",
+            GAME_LOSE_MSG_TEXT_RECT.left,
+            GAME_LOSE_MSG_TEXT_RECT.top,
+            TTE_RED_PB
+        );
+    }
+    game_over_process_user_input();
```
ricfehr3: Looking at this now, it's clear that the `win` and `lose` functions are logically the same, but just have different data (what is printed with tte_printf in the update, and the of the box (red/blue). 

I think it would be better to pull that logic out into one spot. Either we can use the table of data as we have done in the past:

```c
enum EndCondition
{
   WIN, LOSS, NUM_ENDING
};

struct GameOverData 
{
    Color text_color;
    char* text;
};

GameOverData datas[NUM_ENDING] = { ... }

...

static void __update(enum EndCondition cond)
{
  tte_printf(
    "#{P:%d,%d; cx:0x%X000}%s",
    datas[cond].text,
    GAME_LOSE_MSG_TEXT_RECT.left,
    GAME_LOSE_MSG_TEXT_RECT.top,
    datas[cond].text_color
  );  
}

void game_loss_on_update()
{
    __update(LOSS);
}
```
Or if you wanted to do an if-statement that's totally chill too here. We know there's only going to be two states at the end, so it's fine if it's simple.

source/game/game_over.c
```diff
@@ -0,0 +1,116 @@
...
+    // Using the text color to match the "Game Over" text
+    affine_background_set_color(TEXT_CLR_RED);
+}
+
+void game_win_on_init(void)
+{
+    game_over_init();
+    // Using the text color to match the "You Win" text
+    affine_background_set_color(TEXT_CLR_BLUE);
+}
```
ricfehr3: Similar to the comment below: https://github.com/GBALATRO/balatro-gba/pull/468/changes#r3198433879 . The update function should be modified, however these can stay if you'd like. They are both so small, that the duplication is fine for me as more code is required to move it to another function, haha. But you can apply the same pattern if you'd like for consistency. Either way is good.

ricfehr3 (changes requested):
Nice! Thank you again for taking on a refactoring task. This one is looking really good. I left a few comments. Mostly that one pattern with the update functions are my concern. After that has been addressed, I think it's time to merge.

source/game.c
```diff
@@ -697,6 +680,55 @@ void game_init()
     init_unbeaten_blinds_list(true);
 }
 
+// This function isn't set in stone. This is just a placeholder
+// allowing the player to restart the game. Thought it would be nice to have
+// util we decide what we want to do after a game over.
+// Warning: do NOT call outside of the "GAME OVER" state, some data will not be reset here.
```
MeirGavish: Maybe it's time to remove the "Not set in stone" comment and properly give this function a doxygen documentation, especially since it's exposed in the header so that's where the documentation should be.
Also, I would change the warning a bit to make it clearer that's just the current state of the code and not a hard rule to never be broken.

"Warning: This function is currently only meant to be called from the "GAME_OVER" state and shouldn't be called from other states, otherwise some data such as shop jokers may not be properly reset."

MeirGavish (changes requested):
Great! Just a change to the documentation and I can approve this.

source/game.c
MathisMartin31: Totally makes sense ^^

source/game/game_over.c
MathisMartin31: Yep, I'll factorize some code this makes a lot of sense

MathisMartin31:
@MeirGavish I documented the `game_reset()` function ^^

@ricfehr3 I tried using tables as you advised, but I kept getting weird errors from affine_backrgound.c about some some "statement will never be executed" or stuff not being determined at compile time? I ended up using switches but I think it looks fine.

ricfehr3 (approved):
Awesome! Looks good. As for the table vs the switch statement, this is good for me. When it's just always two states it's probably not worth all the complexity anyways. I'm going to merge this since you addressed @MeirGavish's [comment](https://github.com/GBALATRO/balatro-gba/pull/468#discussion_r3201132445) as well.

Thanks again for helping with the refactors. This is really good to get out of the way finally!

MeirGavish (commented):
I did have one thing to say though... I guess we can open an issue about this.

source/game/game_over.c
```diff
@@ -0,0 +1,131 @@
...
+static const Rect     GAME_LOSE_MSG_TEXT_RECT    = {104,  72, UNDEFINED, UNDEFINED};
+// 1 character to the right of GAME_LOSE
+static const Rect     GAME_WIN_MSG_TEXT_RECT     = {112,  72, UNDEFINED, UNDEFINED};
+static const BG_POINT NEW_RUN_BTN_DEST_POS       = { 15,  26};
+static const Rect     NEW_RUN_BTN_SRC_RECT       = {  0,  30,   4,  31};
+// clang-format on
+
+static const u32 GAME_OVER_ANIM_FRAMES = 15;
+
+static enum EndCondition condition = END_CONDITION_NONE;
```
MeirGavish, line 36: I think this variable adds another unnecessary layer that could go out of sync and it would be better to remove it and use the game state instead.
Edit: I see the game state is not exposed yet, I guess this can be an issue for after it is.

MathisMartin31: Yes I think this should be done with the state machine refactor

ricfehr3:
> I did have one thing to say though... I guess we can open an issue about this.

Oops! My bad Meir. I'll do "will merge soon" warning next time.

MeirGavish:
> > I did have one thing to say though... I guess we can open an issue about this.
> 
> Oops! My bad Meir. I'll do "will merge soon" warning next time.

It's okay, no big deal 😄

PR #469, opened 2026-05-03
Bump clang-format version from 18 (via apt) to 22.1.3 (via pip)

SuperbMuffin:
Version 18 is pretty outdated and this bumps it.

### DIFF
--- a/source/game.c
+++ b/source/game.c
@@ -3764,8 +3764,7 @@ static inline void game_round_end_print_hand_reward(int hand_y_offset)
     // Increment the hand reward text until the hand reward variable is depleted
     else if (
         g_game_vars.timer > TM_HAND_REWARD_INCR_WAIT &&
-        g_game_vars.timer % FRAMES(TM_REWARD_INCREMENT_INTERVAL) == 0
-    )
+        g_game_vars.timer % FRAMES(TM_REWARD_INCREMENT_INTERVAL) == 0)
     {
         hand_reward--;
         tte_printf(
@@ -3802,8 +3801,7 @@ static inline void game_round_end_print_interest_reward(int interest_y_offset)
     // Increment the interest reward text until the interest reward variable is depleted
     else if (
         g_game_vars.timer > interest_start_time + TM_REWARD_DISPLAY_INTERVAL &&
-        g_game_vars.timer % FRAMES(TM_REWARD_INCREMENT_INTERVAL) == 0
-    )
+        g_game_vars.timer % FRAMES(TM_REWARD_INCREMENT_INTERVAL) == 0)
     {
         interest_to_count--;
         tte_printf(
@@ -3855,8 +3853,7 @@ static void game_round_end_display_rewards()
     }
     else if (
         interest_start_time != UNDEFINED && g_game_vars.timer >= interest_start_time &&
-        interest_to_count > 0
-    )
+        interest_to_count > 0)
     {
         game_round_end_print_interest_reward(interest_y_offset);
     }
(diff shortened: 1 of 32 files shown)


ricfehr3 (changes requested):
Nice! Honestly this one has tripped me up, but I wanted to keep it consistent with "the _ubuntu_ standard" I guess, lol. But it is old and the changes here show where the compatibility is broken. 

Two things one this:

1. I think we should find a way to modify the `.clang-format` file to keep the long if statements as we had them.

2. Can you add documentation in `CONTRIBUTING.md` so others know what version of `clang-format` to use and how to download it with `pip`? 

thank you!

SuperbMuffin:
I updated the CONTRIBUTING.md section, I also noticed that source/game and include/game directories weren't included so I just updated that while I was at it.

I tried a couple methods to fix 1. but it's kinda been a struggle.

SuperbMuffin:
I think we might just have to bite the bullet on these formatting changes, if anyone else knows how let me know.

source/game.c
```diff
@@ -1625,7 +1625,8 @@ static void game_round_on_init(void)
         CUR_BLIND_TOKEN_POS.x,
         CUR_BLIND_TOKEN_POS.y,
         PLAYING_BLIND_TOKEN_LAYER
-    ); // Create the blind token sprite at the top left corner
+    ); // Create the blind token sprite at the top left
```
ricfehr3, line 1628: Let's put this comment above so it can be formatted nicely.

ricfehr3 (commented):
Hey, sorry it took me a minute to get to this. Thanks for the PR! I agree that we should update the clang format version as the default apt version is old. However after looking at it, going to an option that only uses pip brings potential problems. Most linux distros will not let you install pip packages globally. You have to force it to install and it's bad practice as it breaks system dependencies. Otherwise, you'd have to use a virtual environment. I don't think the headaches of virtual environments are worth it for something not really "required" for the project, like clang-format. 

It is possible to install the newest via apt. But you have to set the repos up to fetch from the latest llvm repos: https://apt.llvm.org/ . This is also too much to do, imo.

It's unfortunate, but I don't see any way of simplifying the pip approach or the llvm apt repo approach to get the latest version. It looks like `clang-format-20` is available in default `apt` repos. I think it's fair to update to that as a middle ground. But keep in mind; IDEs and other tools expect a `clang-format` binary and not `clang-format-20` for example. From a quick test it looks like ubuntu installs `clang-format-18` by default and symlinks it in `/usr/bin`. So, if possible, calling `clang-format` should correctly link to `clang-format-20`. I would rather not modify our written instructions and CI calls to use `clang-format-20` as that may need an upgrade to `clang-format-21` one day, for example:

```sh
[balatro-gba] ls -alh $(which clang-format)                                                                                                                               
lrwxrwxrwx 1 root root 15 Mar 30  2024 /usr/bin/clang-format -> clang-format-18
```

An alias may also work, but either way, it's an extra step to make everything work. I really want to avoid any of that, as clang-format is a necessary evil that already pisses everyone off. To have different versions installed on your machine that may make different results that don't work in the CI but work on your machine is a nightmare.

I'm thinking it would be better to keep it on version 18 for now, instead of fighting the uphill battle with getting a newer version. I think we do need some instructions on clarifying what version of clang-format we use, instructions on how to install it, and instructions on checking the version. 

For example, on ubuntu machines, you just install `clang-format`, but on arch you can install `clang18` and get the compatible version. 

Sorry to be so particular on this. It's a tough change because it's a big workflow inhibitor and doesn't have much gain for the effort. For now, I'm going to close this. Feel free to make another PR if you have a way to have a simple approach that is almost as straightforward as the regular apt install. I'll be making a ticket for clarifying the use and version of clang-format as well.

Issue #465, opened 2026-05-03
Finish Refactor of `Game.c`


MeirGavish:

MathisMartin31:
I think I'll take a look at this one, since it doesn't look like the `game_playing_...` functions use the state/substate machine

Edit: Welllll, maybe I should go for the shop screen first in the end, I need the encapsulation of the joker selection I did there.

MeirGavish:
This is probably the biggest, most complex one

MathisMartin31:
> This is probably the biggest, most complex one

Indeed, just looking at it makes my head spin lol

ricfehr3:
Taking this one on now.

MathisMartin31:
> Taking this one on now.

Best of luck comrade 🫡

ricfehr3:
After starting and restarting a couple times now. I think I should convert this work into subtasks. I don't even have the full list yet, but I know some subtasks I can already start. I don't want to bring a huge PR for this, haha. 

For now, I'm going to focus on things to do in `game.c` before moving it to a separate file. This way I can shrink some code without worrying about variable scope yet.

ricfehr3:
I renamed the issue after deciding to restructure the remaining refactor, see [comment in discussion](https://github.com/GBALATRO/balatro-gba/discussions/518#discussioncomment-17121169)

MeirGavish:
I just want to record that extracting round code from game.c/.h into round.h/.c still needs to be done so I guess make it another sub-issue for this?

ricfehr3:
> I just want to record that extracting round code from game.c/.h into round.h/.c still needs to be done so I guess make it another sub-issue for this?

Sorry I didn't get back sooner. But yeah, the subissue looks good to me. Thanks for adding it!

PR #470, opened 2026-05-05
Refactor/state machine

### LINKED ISSUE #471: Create file for state machine `state_machine.h`

ricfehr3:
A WIP PR of refactoring the state machine. 

The state machine is strongly tied to the game states only. However, it's structure can be used elsewhere and decoupling from the existing state machine is a better design and easier to maintain going forward. 

:stop_sign: Based on #461 , need to merge that first.
Related state machine discussion: [here](https://github.com/GBALATRO/balatro-gba/issues/156)

---

**EDIT** 5/20/26 (MM/DD/YY):

This is no longer WIP, copilot review is finished. Will merge 5/21/26 if no further reviews. Thanks!


include/state_machine.h
```diff
@@ -0,0 +1,45 @@
...
+// efficient. Well, this is the answer.
+// Thanks!
+// https://github.com/cellos51/balatro-gba/issues/137#issuecomment-3322485129
+void noop(void);
+
+// clang-format off
+#define STATE_INFO_UPDATE_FN_ONLY(fn) {.on_init = noop, .on_update = fn, .on_exit = noop}
+// clang-format on
+
+#endif // STATE_MACHINE_H
```
ricfehr3, line 107: Document this.

source/state_machine.c
```diff
@@ -0,0 +1,61 @@
...
+    if (new_state >= 0 && new_state < state_machine->num_infos)
+    {
+        state_machine->state_infos[new_state].on_init();
+    }
+
+    // This has an inadvertant "feature". If you are in the calling function of a state, calling
+    // this function on it's own state machine. Aka, you are changing states mid update. Then
+    // Because it is only updating the active pointer of code to-be ran, the update function
+    // finishes before continuing. This is incredibly helpful as you can initiate a state change in
+    // the update, but just don't think the code is breaking on that change state call.
```
ricfehr3: I can clear this up. Doesn't make a lot of sense rereading it.

source/list.c
```diff
@@ -306,3 +306,20 @@ void list_itr_remove_current_node(ListItr* itr)
     s_list_remove_node(itr->list, itr->current_node);
     itr->current_node = tmp_prev;
 }
+
+bool list_remove_at(List* list, void* data)
+{
```
ricfehr3: Add a test for this

source/game/blind_select.c
```diff
@@ -39,15 +42,22 @@ enum BlindSelectState
...
+    STATE_INFO_UPDATE_FN_ONLY(game_blind_select_start_anim_seq),
+    STATE_INFO_UPDATE_FN_ONLY(game_blind_select_handle_input),
+    STATE_INFO_UPDATE_FN_ONLY(game_blind_select_selected_anim_seq),
+    STATE_INFO_UPDATE_FN_ONLY(game_blind_select_display_blind_panel),
+    STATE_INFO_UPDATE_FN_ONLY(game_blind_select_exit),
+};
+
+static StateMachine blind_select_sm = {
+    .state_infos = &state_info[0],
+    .num_infos = BLIND_SELECT_MAX,
```
ricfehr3, line 61: I'm considering making this a macro to declare everything. Any thoughts from anyone on that? I think it would look nicer.

source/game.c
```diff
@@ -225,6 +217,18 @@ static const BG_POINT HAND_PLAY_POS         = {120,     70};
 // variable and handling in common.c once the related refactor is finished
 static enum BackgroundId background_legacy = BG_NONE;
 
+static StateInfo state_info[] = {
+#define DEF_STATE_INFO(stateEnum, init_fn, update_fn, exit_fn) \
```
ricfehr3, line 208: I think it may be time to remove the x macro wrapper around this. With the new way of declaring state machines it makes sense to be consistent with other parts of the code now.

I'm going to remove it.

MathisMartin31:
@ricfehr3 I'll let you address the Copilot review first as it's quite extensive, and I'll take a look afterwards 😃

source/game/blind_select.c
ricfehr3: nah, not yet at least

include/list.h
```diff
@@ -68,6 +68,11 @@ typedef struct List
     int len;
 } List;
 
+/**
+ * @brief Default list declaration for empty lists
+ */
+#define LIST_DEFAULT {.head = NULL, .tail = NULL, .len = 0}
```
ricfehr3: @MathisMartin31 I took inspiration from your default const values for resetting variables that you have been doing. I think this would be a lot more straightforward over calling `list_create()`. I think we should use these everywhere instead of `list_create()` when initializing lists.

What do you think? @MeirGavish as well.

ricfehr3:
> @ricfehr3 I'll let you address the Copilot review first as it's quite extensive, and I'll take a look afterwards 😃

Done! Thanks for taking a look  :)

include/list.h
MathisMartin31: I think both solutions work, but I would assume this is more efficient since it's evaluated at compile time and saves us both the call to `list_create()` and an extra allocation within that function. Maybe the compiler is optimizing it, and we're not calling it too often but at least we're sure.

include/list.h
```diff
@@ -222,6 +229,16 @@ void* list_get_at_idx(List* list, unsigned int idx);
...
 
+/**
+ * Remove a List's node with the matching pointer
+ *
+ * @param list pointer to a @ref List
+ * @param data pointer to data in node in list
+ *
+ * @return `true` if successfully removed, `false` otherwise
+ */
+bool list_remove_at(List* list, void* data);
```
MathisMartin31: Nice! Definitely a nice addition to the API.

How would that work with a List iterator? Like, while iterating on a list, I get a node, check some things and decide I need to remove it. Would the iterator then be invalid?

source/game.c
MathisMartin31: Yes I think it's best to stick to the new way of handling the states

MathisMartin31 (approved):
Nice work, very clean! Just two comments but it's more questions than anything so I'll approve this PR, so merge to your heart's content ^^

include/list.h
ricfehr3: Those are good points. For me the only "upgrade" is style. `list_create` sounds like more is happening in my opinion, but `LIST_DEFAULT` is more accurate to what's actually going on. I think either this should be used, or `list_create` should be renamed to `list_default`. Either way, that can be a ticket. Don't want to inflate the PR.

include/list.h
ricfehr3: It would break the list iterator. But there is a special function to remove nodes for list iterators: `list_itr_remove_current_node`. I think I'll add a bit of docs to say when to use what.

source/game.c
ricfehr3: I'm actually reconsidering this. I think the X-macro for this is still good after trying to remove it. I think it should be expanded on to be used to compress the enum/array/state machine pattern. Once again though, let's not inflate this. We can revisit these after the feature release.

include/list.h
MeirGavish: I think there is value in `list_create`'s documentation that explains that a list needs to be cleared. Maybe such documentation needs to be added to `list_clear` to explain that it **has** to be called to free up data. I guess it's okay to remove the function as long as the documentation is preserved. Either way I think better names could be `list_init()`/`LIST_INIT`.

MeirGavish: Also, I think the default values macro can be added without removing the function.

include/list.h
```diff
@@ -222,6 +229,18 @@ void* list_get_at_idx(List* list, unsigned int idx);
...
+ * Remove a List's node with the matching pointer
+ *
+ * @param list pointer to a @ref List
+ * @param data pointer to data in node in list
+ *
+ * @return `true` if successfully removed, `false` otherwise
+ *
+ * @note When working with @ref ListItr, use @ref list_remove_at()
+ */
+bool list_remove_at(List* list, void* data);
```
MeirGavish: @ricfehr3 
Maybe just rename this to `list_remove_data()` to better differentiate from `list_remove_at_idx()`?

include/list.h
ricfehr3: I agree Meir. I'll update the docs for `list_clear` and rename `create` to `init`.

include/list.h
ricfehr3: I agree, that sounds much better.

ricfehr3:
Going to merge this one so I can work on #465 .

PR #472, opened 2026-05-06
Updating the save mechanism by creating an intermediary SaveData struct

MathisMartin31:
From our discussion in PR #468.

Put it up as a draft for now so @ricfehr3 can take a look at the SaveData struct

### DIFF
--- a/include/save.h
+++ b/include/save.h
@@ -11,15 +11,27 @@
 #include <tonc.h>
 
 /**
- * @brief Save game data to SRAM.
+ * @brief Save current run data to SRAM.
  */
 void save_game();
 
 /**
- * @brief Load game data from SRAM.
+ * @brief Load previous run data from SRAM.
  *
  * @sa save_game
  */
 void load_game();
+
+/**
+ * @brief Save options values to SRAM.
+ */
+void save_options();
+
+/**
+ * @brief Load options values from SRAM.
+ *
+ * @sa load_options
+ */
+void load_options();
 
 #endif // SAVE_H
--- a/include/save.h
+++ b/include/save.h
@@ -5,6 +5,7 @@
  *
  * Here is an overwiew of the contents of a valid save file.
  *
+ * ```
  * ┌─────────────┐ <-- SaveHeader
  * │   Header    │
  * ├─────────────┤ <-- SaveOptions
@@ -13,6 +14,7 @@
  * │ Engine vars │
  * │   Jokers    │
  * └─────────────┘
+ * ```
  *
  * SaveHeader indicates save validity by its presence.
  * SaveOptions contains options values set in the corresponding menu and apply to the game itself.
--- a/include/save.h
+++ b/include/save.h
@@ -2,6 +2,25 @@
  * @file save.h
  *
  * @brief Utils functions to save/load data structures from/to the SRAM.
+ *
+ * Here is an overwiew of the contents of a valid save file.
+ *
+ * ┌─────────────┐ <-- SaveHeader
+ * │   Header    │
+ * ├─────────────┤ <-- SaveOptions
+ * │   Options   │
+ * ├─────────────┤ <-- SaveGame
+ * │ Engine vars │
+ * │   Jokers    │
+ * └─────────────┘
+ *
+ * SaveHeader indicates save validity by its presence.
+ * SaveOptions contains options values set in the corresponding menu and apply to the game itself.
+ * SaveGame contains values tied to a run with round, ante, money, owned Jokers etc...
+ *
+ * @sa SaveHeader, SaveOptions, SaveGame
+ *
+ * I strongly recommend using `xxd -l 512 -e -g 4 <gbalatro.sav>` to view the contents of the save
  */
 #ifndef SAVE_H
 #define SAVE_H
--- a/include/save.h
+++ b/include/save.h
@@ -30,7 +30,7 @@ void save_options();
 /**
  * @brief Load options values from SRAM.
  *
- * @sa load_options
+ * @sa save_options
  */
 void load_options();
 
--- a/include/save.h
+++ b/include/save.h
@@ -13,25 +13,25 @@
 /**
  * @brief Save current run data to SRAM.
  */
-void save_game();
+void save_game(void);
 
 /**
  * @brief Load previous run data from SRAM.
  *
  * @sa save_game
  */
-void load_game();
+void load_game(void);
 
 /**
  * @brief Save options values to SRAM.
  */
-void save_options();
+void save_options(void);
 
 /**
  * @brief Load options values from SRAM.
  *
  * @sa save_options
  */
-void load_options();
+void load_options(void);
 
 #endif // SAVE_H
--- a/source/game.c
+++ b/source/game.c
@@ -678,8 +678,6 @@ void game_init()
 
     jokers_available_to_shop_init();
 
-    load_game();
-
     hands = max_hands;
     discards = max_discards;
     g_game_vars.timer = TM_ZERO;
--- a/source/save.c
+++ b/source/save.c
@@ -141,34 +141,6 @@ static const SaveOptions SaveOptions_default = {
     .padding = {UNDEFINED, UNDEFINED, UNDEFINED},
 };
 
-GBAL_UNUSED
-static inline u32 get_header_size()
-{
-    static u32 SaveHeader_size = 0;
-    static bool SaveHeader_size_known = false;
-    if (!SaveHeader_size_known)
-    {
-        SaveHeader header;
-        SaveHeader_size = sizeof(header);
-        SaveHeader_size_known = true;
-    }
-    return SaveHeader_size;
-}
-
-GBAL_UNUSED
-static inline u32 get_options_size()
-{
-    static u32 SaveOptions_size = 0;
-    static bool SaveOptions_size_known = false;
-    if (!SaveOptions_size_known)
-    {
-        SaveOptions options;
-        SaveOptions_size = sizeof(options);
-        SaveOptions_size_known = true;
-    }
-    return SaveOptions_size;
-}
-
 /**
  * @brief Write raw binary data to SRAM
  *
--- a/source/save.c
+++ b/source/save.c
@@ -26,7 +26,7 @@
 #define CHECK_HASH_SIZE 7
 #define GIT_HASH_START  17 // starts after "GBALATRO-VERSION:" in the gbalatro_version var
 
-#define SAVE_LABEL_SIZE 4
+#define SAVE_LABEL_SIZE 16
 
 // clang-format off
 /**
@@ -63,7 +63,7 @@ typedef struct SaveHeader
  */
 typedef struct SaveData
 {
-    u32 tag_internal[SAVE_LABEL_SIZE];
+    char tag_internal[SAVE_LABEL_SIZE];
 
     s32 timer;
     u32 rng_seed;
@@ -73,7 +73,7 @@ typedef struct SaveData
     int money;
     u32 padding0[2];
 
-    u32 tag_options[SAVE_LABEL_SIZE];
+    char tag_options[SAVE_LABEL_SIZE];
 
     u8 game_speed;
     bool high_contrast;
@@ -81,23 +81,17 @@ typedef struct SaveData
     u8 sound_volume;
     u32 padding1[3];
 
-    u32 tag_jokers[SAVE_LABEL_SIZE];
+    char tag_jokers[SAVE_LABEL_SIZE];
 
     u32 jokers_data[2 * MAX_JOKERS_HELD_SIZE];
 
-    u32 tag_end;
+    char tag_end[4];
 } SaveData;
 
 // clang-format off
-static const SaveData SaveData_default = {
-    // Spells "INTERNAL DATA"
-    .tag_internal = {
-        0x544E492D,
-        0x414E5245,
-        0x4144204C,
-        0x2D204154
-    },
-
+static const SaveData SaveData_default =
+{
+    .tag_internal = "-INTERNAL DATA -",
     .timer = 0,
     .rng_seed = 0,
     .rng_step = 0,
@@ -106,29 +100,17 @@ static const SaveData SaveData_default = {
     .money = 0,
     .padding0 = {UNDEFINED, UNDEFINED},
 
-    // Spells "GAME OPTIONS"
-    .tag_options = {
-        0x4147202D, 
-        0x4F20454D, 
-        0x4F495450, 
-        0x2D20534E
-    },
+    .tag_options = "- OPTIONS DATA -",
     .game_speed = 0,
     .high_contrast = false,
     .music_volume = 0,
     .sound_volume = 0,
     .padding1 = {UNDEFINED, UNDEFINED, UNDEFINED},
 
-    .tag_jokers = {
-        0x574F202D,
-        0x2044454E,
-        0x454B4F4A,
-        0x2D205352,
-    },
-
+    .tag_jokers = "- OWNED JOKERS -",
     .jokers_data = {},
 
-    .tag_end = 0x444E455F
+    .tag_end = "_END"
 };
 // clang-format on
 
--- a/source/save.c
+++ b/source/save.c
@@ -21,11 +21,12 @@
 
 #define CHECK_BASE 0x0000
 #define GAME_BASE  0x0010
-#define LISTS_BASE 0x0060
 
 #define CHECK_MAGIC     0x4C414247 // Spells GBAL, used to determine if the save data is junk
 #define CHECK_HASH_SIZE 7
 #define GIT_HASH_START  17 // starts after "GBALATRO-VERSION:" in the gbalatro_version var
+
+#define SAVE_LABEL_SIZE 4
 
 // clang-format off
 /**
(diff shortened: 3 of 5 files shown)


MathisMartin31:
Also just a bit of context, I tried to make everything aligned and added some labels so the save looks like this in a hex viewer

image

source/save.c
```diff
@@ -47,6 +48,90 @@ typedef struct SaveHeader
...
+
+// clang-format off
+static const SaveData SaveData_default = {
+    // Spells "INTERNAL DATA"
+    .tag_internal = {
+        0x544E492D,
+        0x414E5245,
+        0x4144204C,
+        0x2D204154
+    },
```
MeirGavish: Why not make the tags a char/u8/s8 (not sure which would be best) so you can actually spell out the string instead of giving hex values?

MeirGavish (commented):
Great!
I would say there is no need to keep the game variables in the save data before progress save is implemented and they are actually used.
Either way I would separate it to 2 different structs of settings and game variables so they can be loaded separately.

source/save.c
MathisMartin31: Omg I'm so dumb it works looool. I con't do it for the Header's magic word because it would become a pain to test against it to validate the data.

MathisMartin31: Ok so several things here:
- If we save/load several structs instead of 1, we would also need that many headers because we can save options values without actually going in game long enough to reach a checkpoint and save the game. With only 1 header, we would believe the game data is valid event though we didn't actually write anything at that address.
- I'm encountering a bug I hadn't before: past the first write, the saved data is mangled so that the first byte of each 4-bytes word is repeated 4 times to fill the entire word. This wasn't happening before...
image

MathisMartin31: Okay I found when the bug happens:

- This code works fine
```
write_sram(OPTIONS_ADDRESS, (const u8*)&options, sizeof(options));
```

- This doesn't work
```
u32 address = SAVE_ADDRESS + get_header_size();
write_sram(address, (const u8*)&options, sizeof(options));
```

I don't know *why* it happens but giving addresses as defines works, but a variable doesn't... *by all accounts, this makes no sense!*

Sooooo, I'm reverting to defines. I don't think this will be too problematic since by separating options and run data into 2 structs, and writing the options before the run data, the only struct that may grow in the future is the last one, so the defines shouldn't move

MathisMartin31: > If we save/load several structs instead of 1, we would also need that many headers because we can save options values without actually going in game long enough to reach a checkpoint and save the game. With only 1 header, we would believe the game data is valid event though we didn't actually write anything at that address.

Also @MeirGavish, we can't really load game data before saving them along options values since that would defeat the whole purpose of separating the structs to only save/load what we need

Edit: Ooooor... we can store 2 additional booleans in the SaveHeader to indicate if options or game data are valid, we just have to check the header before writing it

Edit Edit: a bitfield also works ^^

source/save.c
```diff
@@ -114,53 +208,138 @@ static inline bool is_version_dirty()
...
+
+    // Lists
+
+    List* jokers_list = get_jokers_list();
+    u32 nb_jokers = list_get_len(jokers_list);
+
+    int i = 0;
+    for (; i < nb_jokers; i++)
+    {
+        JokerObject* joker_object = list_get_at_idx(jokers_list, i);
```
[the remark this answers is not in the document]
MathisMartin31: Weirdly enough I don't get any warnings for this? In any case it's still good practice to compare apples to apples so I'll do it.
As for clamping I don't think this is necessary, since we make sure elsewhere that the Jokers' list cannot have more than `MAX_JOKERS_HELD_SIZE` elements in it.

source/save.c
```diff
@@ -38,15 +46,101 @@
...
+    u32 rng_seed;
+    u32 rng_step;
+    int round;
+    int ante;
+    int money;
+    u32 padding[2];
+
+    char tag_jokers[SAVE_LABEL_SIZE];
+    u32 jokers_data[2 * MAX_JOKERS_HELD_SIZE];
+
```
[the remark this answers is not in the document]
MathisMartin31: I think this is on the verge of what's in the scope of this PR or not. This PR aims to update the way we save things, imo the "what" we save/restore and how we do it should be for another PR once all the states have been properly refactored so we can add a "resume" menu to load the previous save.

source/save.c
```diff
@@ -114,53 +208,138 @@ static inline bool is_version_dirty()
...
+    memcpy(&(header.githash), gbalatro_version + GIT_HASH_START, CHECK_HASH_SIZE);
+
+    write_sram(HEADER_ADDRESS, (const u8*)&header, sizeof(header));
+}
+
+void save_options(void)
+{
+    set_save_header(SAVE_SECTION_FLAG_OPTIONS);
+
+    SaveOptions options = SaveOptions_default;
```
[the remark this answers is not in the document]
MathisMartin31: That's an actually good point :)

source/game.c
```diff
@@ -678,8 +678,6 @@ void game_init()
 
     jokers_available_to_shop_init();
 
-    load_game();
-
     hands = max_hands;
     discards = max_discards;
     g_game_vars.timer = TM_ZERO;
```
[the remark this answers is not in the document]
MathisMartin31: The "resume" mechanism is not a feature *yet*. As of now only options values are saved/loaded so I maintained this fact by replacing the call to `load_game` by one to `load_options` and moved it to main.c where it is only called once on game boot, instead of each time we start a run

source/save.c
```diff
@@ -114,53 +208,138 @@ static inline bool is_version_dirty()
...
 }
 
 void save_game(void)
 {
-    set_save_header();
-    write_sram(GAME_BASE, (const u8*)&g_game_vars, sizeof(g_game_vars));
+    set_save_header(SAVE_SECTION_FLAG_GAME);
+
+    SaveGame game = SaveGame_default;
+
```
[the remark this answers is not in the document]
MathisMartin31: Same comment as for the `save_options` function, this makes sense.

source/save.c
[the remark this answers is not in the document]
MeirGavish: `list_get_len()` returns int so that should be the type of `nb_jokers`, I guess you can do the suggested cast for `list_get_at_idx()`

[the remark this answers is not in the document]
MeirGavish: > Weirdly enough I don't get any warnings for this? In any case it's still good practice to compare apples to apples so I'll do it.
> As for clamping I don't think this is necessary, since we make sure elsewhere that the Jokers' list cannot have more than `MAX_JOKERS_HELD_SIZE` elements in it.

Yeah, I think you can forego the clamping, it feels a little off here...

source/save.c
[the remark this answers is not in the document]
MeirGavish: Agreed. Resolving.

source/save.c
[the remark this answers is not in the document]
MathisMartin31: > `list_get_len()` returns int so that should be the type of `nb_jokers`, I guess you can do the suggested cast for `list_get_at_idx()`

@ricfehr3 Question though (I believe you were the one implementing the List struct): why is `list.len` an int and not a u32? From what I can see it is either increased, set to zero, or decreased, but in that last case we ensure the list is not empty because there are elements in it, so `list.len > 0` which means that `list.len-- >= 0`

MathisMartin31:
@MeirGavish @ricfehr3 Your comments should be addressed now ^^

source/save.c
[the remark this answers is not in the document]
ricfehr3: > > `list_get_len()` returns int so that should be the type of `nb_jokers`, I guess you can do the suggested cast for `list_get_at_idx()`
> 
> @ricfehr3 Question though (I believe you were the one implementing the List struct): why is `list.len` an int and not a u32? From what I can see it is either increased, set to zero, or decreased, but in that last case we ensure the list is not empty because there are elements in it, so `list.len > 0` which means that `list.len-- >= 0`

¯\_(ツ)_/¯ I don't remember if there was a reason, honestly. I just checked and there is no time it's `UNDEFINED`. We could change it if we do encounter any issues, but I think it's fine for now.

include/save.h
```diff
@@ -11,15 +11,27 @@
...
+ * @brief Save options values to SRAM.
+ */
+void save_options();
+
+/**
+ * @brief Load options values from SRAM.
+ *
+ * @sa save_options
+ */
+void load_options();
```
ricfehr3:
```suggestion
void save_options(void);

/**
 * @brief Load options values from SRAM.
 *
 * @sa save_options
 */
void load_options(void);
```

For consistency.

source/save.c
```diff
@@ -19,14 +20,21 @@
 //   - Memory is filled with 1s by default
 //     (at least in mgba, not sure about real HW)
 
-#define CHECK_BASE 0x0000
-#define GAME_BASE  0x0010
-#define LISTS_BASE 0x0060
+// For some reason writing to SRAM only works if the address is given as a define
+#define HEADER_ADDRESS  0x0
+#define OPTIONS_ADDRESS 0x10
+#define GAME_ADDRESS    0x30
```
ricfehr3:
```suggestion
#define HEADER_ADDRESS  0x0
#define OPTIONS_ADDRESS 0x10
#define GAME_ADDRESS    0x30
```

These are fine to keep as `#define`. I was able to get it to work with `static const u32` instead, and with local variables. Either way, we don't have to change it, but maybe just remove the comment.

source/save.c
```diff
@@ -114,53 +208,136 @@ static inline bool is_version_dirty()
...
+    {
+        JokerObject* joker_object = list_get_at_idx(jokers_list, (u32)i);
+        game.jokers_data[2 * i] = (u32)joker_object->joker->id;
+        game.jokers_data[2 * i + 1] = joker_object->joker->persistent_state;
+    }
+    for (; i < MAX_JOKERS_HELD_SIZE; i++)
+    {
+        game.jokers_data[2 * i] = UNDEFINED;
+        game.jokers_data[2 * i + 1] = UNDEFINED;
+    }
```
ricfehr3: I think it would make sense here to use a different structure to hold this info. 

Maybe an array of structs that have an `{ int id, int state }`? `struct JokerObjectSaveData` or similar?

This way we wouldn't need to iterate through an array with offsets like this.

source/save.c
```diff
@@ -38,15 +46,101 @@
  * 0    | 0x47   | 0x42   | 0x41   | 0x4C   | MAGIC        | Identify if proceeding data is valid and not junk, spells "GBAL"
  * 1    | Dirty  | H[0]   | H[1]   | H[2]   | GITHASH_LOW  | Dirty flag, followed by the first 3 bytes of shortened git hash H
  * 2    | H[3]   | H[4]   | H[5]   | H[6]   | GITHASH_HIGH | Last 4 bytes of shortened git hash H, with a dirty flag
+ * 3    | SEC[0] | SEC[1] | SEC[2] | SEC[3] | VALID_SCTNS  | Identifies whether each section of the save data is valid or not
```
ricfehr3: I would be nice here if the entire save file was written out like this too. We can leave that as a new issue if you would like though.

source/save.c
ricfehr3:
```suggestion
#define HEADER_ADDRESS  0x0
#define OPTIONS_ADDRESS 0x10
#define GAME_ADDRESS    0x30
```

I was able to use `static const u32` here and local variables. I think you can remove the comment. You can leave them as `#define` though if you'd prefer.

ricfehr3 (changes requested):
Hey, gave this another look. Looking good! I left a couple of comments that should be addressed, but otherwise it's just about ready for approval. Hopefully this will unblock further work for you too!

source/save.c
MathisMartin31: You mean doing that for all Save structs (Game and Options)?

ricfehr3: Yes, or just the general structure in the memory of the following layout past the header. Maybe it shouldn't be here because it's only the header... But we should illustrate the layout of the save file somewhere.

MathisMartin31: Maybe in the description of save.h, after the brief?

Or a big comment at the beginning of save.c

ricfehr3: I think the beginning of `save.h` should have a general overview. Something that just shows an overview of how the data is structured and how to view it with `xxd`

```c
               
 ┌───────────┐ 
 │  Header   │ 
 └───────────┘ 
 ┌───────────┐ 
 │   Data    │ 
 └───────────┘ 
 ┌───────────┐ 
 │   Data    │ 
 └───────────┘ 
               
docs. docs. docs.

How to view it instructions
```

include/save.h
```diff
@@ -2,6 +2,25 @@
...
+ * │   Jokers    │
+ * └─────────────┘
+ *
+ * SaveHeader indicates save validity by its presence.
+ * SaveOptions contains options values set in the corresponding menu and apply to the game itself.
+ * SaveGame contains values tied to a run with round, ante, money, owned Jokers etc...
+ *
+ * @sa SaveHeader, SaveOptions, SaveGame
+ *
+ * I strongly recommend using `xxd -l 512 -e -g 4 <gbalatro.sav>` to view the contents of the save
```
ricfehr3, line 25:
```suggestion
 *
 * Here is an overwiew of the contents of a valid save file.
 *
 * ```
 * ┌─────────────┐ <-- SaveHeader
 * │   Header    │
 * ├─────────────┤ <-- SaveOptions
 * │   Options   │
 * ├─────────────┤ <-- SaveGame
 * │ Engine vars │
 * │   Jokers    │
 * └─────────────┘
 * ```
 *
 * SaveHeader indicates save validity by its presence.
 * SaveOptions contains options values set in the corresponding menu and apply to the game itself.
 * SaveGame contains values tied to a run with round, ante, money, owned Jokers etc...
 *
 * @sa SaveHeader, SaveOptions, SaveGame
 *
 * I strongly recommend using `xxd -l 512 -e -g 4 <gbalatro.sav>` to view the contents of the save
```

It needs this or it doesn't render properly.

source/save.c
```diff
@@ -19,34 +20,167 @@
...
+ * 13   | 'N'    | 'E'    | 'D'    | ' '    | *            |
+ * 14   | 'J'    | 'O'    | 'K'    | 'E'    | *            |
+ * 15   | 'R'    | 'S'    | ' '    | '-'    | *            |
+ * 16   | ID[0]  | ID[1]  | ID[2]  | ID[3]  | JOKER DATA 0 | Minimal necessary data to reconstruct a JokerObject
+ * 17   | STT[0] | STT[1] | STT[2] | STT[3] | *            | Contains the Joker's `id` and `persistent_state`
+ * ...  | ...    | ...    | ...    | ...    | ...          |
+ * ...  | ...    | ...    | ...    | ...    | ...          |
+ * ??   | '_'    | 'E'    | 'N'    | 'D'    | END_TAG      | Spells "_END", marks the end of the savefile
+ */
+// clang-format on
```
ricfehr3, line 125: Image

Image

The table needs an entry in the last spot.
```suggestion
// clang-format off
/**
 * @brief SaveGame will contain the data about the current run to be saved to SRAM.
 *         GameVariables was used for this purpose at first, but some data needed to be shared but
 *         not saved, so it couldn't be dumped "as is" anymore and this struct had to be created.
 *
 * word | Byte 0 | Byte 1 | Byte 2 | Byte 3 | name         | purpose
 * -----|--------|--------|--------|--------|--------------|------------------------------------------------------------------
 * 0    | '-'    | 'I'    | 'N'    | 'T'    | TAG          | Spells "-INTERNAL DATA -"
 * 1    | 'E'    | 'R'    | 'N'    | 'A'    | *            | -
 * 2    | 'L'    | ' '    | 'D'    | 'A'    | *            | -
 * 3    | 'T'    | 'A'    | ' '    | '-'    | *            | -
 * 4    | T[0]   | T[1]   | T[2]   | T[3]   | GLOB TIMER   | The global timer used for animations thoughout the game
 * 5    | SED[0] | SED[1] | SED[2] | SED[3] | RNG SEED     | The seed used for RNG, either randomly shuffled or chosen by the player at game start
 * 6    | STP[0] | STP[1] | STP[2] | STP[3] | RNG STEP     | The current position in the RNG sequence for the given seed, since the start of the run
 * 7    | RND[0] | RND[1] | RND[2] | RND[3] | ROUND        | What Round we are about to start
 * 8    | ANT[0] | ANT[1] | ANT[2] | ANT[3] | ANTE         | What Ante we are on
 * 9    | MNY[0] | MNY[1] | MNY[2] | MNY[3] | MONEY        | How much money we currently have left
 * 10   | UNDEF  | UNDEF  | UNDEF  | UNDEF  | PADDING      | Some padding
 * 11   | UNDEF  | UNDEF  | UNDEF  | UNDEF  | *            | -
 * 12   | '-'    | ' '    | 'O'    | 'W'    | TAG          | Spells "- OWNED JOKERS -"
 * 13   | 'N'    | 'E'    | 'D'    | ' '    | *            | -
 * 14   | 'J'    | 'O'    | 'K'    | 'E'    | *            | -
 * 15   | 'R'    | 'S'    | ' '    | '-'    | *            | -
 * 16   | ID[0]  | ID[1]  | ID[2]  | ID[3]  | JOKER DATA 0 | Minimal necessary data to reconstruct a JokerObject
 * 17   | STT[0] | STT[1] | STT[2] | STT[3] | *            | Contains the Joker's `id` and `persistent_state`
 * ...  | ...    | ...    | ...    | ...    | ...          | -
 * ...  | ...    | ...    | ...    | ...    | ...          | -
 * ??   | '_'    | 'E'    | 'N'    | 'D'    | END_TAG      | Spells "_END", marks the end of the savefile
 */
// clang-format on
```

source/save.c
```diff
@@ -19,34 +20,167 @@
...
+ * 0    | '-'    | ' '    | 'O'    | 'P'    | TAG          | Pretty tag to clearly visualize the Options section in a hex viewer
+ * 1    | 'T'    | 'I'    | 'O'    | 'N'    | *            | Spells "- OPTIONS DATA -"
+ * 2    | 'S'    | ' '    | 'D'    | 'A'    | *            |
+ * 3    | 'T'    | 'A'    | ' '    | '-'    | *            |
+ * 4    | SPEED  | CNTRST | MUSIC  | SOUND  | OPTN_VALUES  | All 4 option values, packed in a single word.
+ * 5    | UNDEF  | UNDEF  | UNDEF  | UNDEF  | PADDING      | Padding, so that the next section starts at the beginning of the
+ * 6    | UNDEF  | UNDEF  | UNDEF  | UNDEF  | *            | next 4-word row in a hex viewer
+ * 7    | UNDEF  | UNDEF  | UNDEF  | UNDEF  | *           |
+ */
+// clang-format on
```
ricfehr3, line 73:
```suggestion
// clang-format off
/**
 * @brief SaveOptions will only contain options data set in the Options Menu
 *
 * word | Byte 0 | Byte 1 | Byte 2 | Byte 3 | name         | purpose
 * -----|--------|--------|--------|--------|--------------|------------------------------------------------------------------
 * 0    | '-'    | ' '    | 'O'    | 'P'    | TAG          | Pretty tag to clearly visualize the Options section in a hex viewer
 * 1    | 'T'    | 'I'    | 'O'    | 'N'    | *            | Spells "- OPTIONS DATA -"
 * 2    | 'S'    | ' '    | 'D'    | 'A'    | *            | -
 * 3    | 'T'    | 'A'    | ' '    | '-'    | *            | -
 * 4    | SPEED  | CNTRST | MUSIC  | SOUND  | OPTN_VALUES  | All 4 option values, packed in a single word.
 * 5    | UNDEF  | UNDEF  | UNDEF  | UNDEF  | PADDING      | Padding, so that the next section starts at the beginning of the
 * 6    | UNDEF  | UNDEF  | UNDEF  | UNDEF  | *            | next 4-word row in a hex viewer
 * 7    | UNDEF  | UNDEF  | UNDEF  | UNDEF  | *           | -
 */
// clang-format on
```

Same thing seen here: https://github.com/GBALATRO/balatro-gba/pull/472/changes#r3214494443

ricfehr3 (approved):
This looks great! I'll leave it up for a day and merge tomorrow. I did leave some simple changes in the form of suggestions to clean up some doxygen rendering errors, but otherwise this is ready to go for me. Nice work again!

MathisMartin31:
Yeah, I really should generate the docs ore often this isn't the first time they're broken ><'
Should be all fixed now :)

MeirGavish (approved):
Looks great!
Did a quick perusal of the code and ran the PR to see nothing seems broken.
Approving.

ricfehr3:
Excellent! Merging. Thank you @MathisMartin31

PR #476, opened 2026-05-07
Refactor/RNG system

### LINKED ISSUE #452: [BUG] Seed RNG on every run

### LINKED ISSUE #456: Move RNG System into it's own file
To avoid save-spamming and allow deterministic random numbers, random numbers have been centralized to `get_rand()` as of #449 . This should be extracted into it's own file along with other random number manipulation. This allows the system to be testable too in our unit tests.

MathisMartin31:
Closes #452 and #456 

To seed the RNG on each run, I extracted the method used by the main menu into its own helper function in random.h and called it in the update function for the game over state. This way, The seed is shuffled when we start the game AND when restarting a run, be it after a win or a loss.


source/random.c
```diff
@@ -0,0 +1,21 @@
...
+#include <stdlib.h>
+
+void shuffle_rng_seed(void)
+{
+    g_game_vars.rng_seed++;
+    // If the keys have changed, make it more pseudo-random
+    if (key_curr_state() != key_prev_state())
+    {
+        g_game_vars.rng_seed *= 2;
+    }
```
ricfehr3: This isn't too random. Every time the seed is shuffled it just increments it and the modifier x2 is easily in control of the player since it's only measuring if there is an input. Maybe two sources of psuedo-random numbers and a hash function? Seems like a good opportunity for something fun.

MathisMartin31: The best way to have good rng would be to get some hardware values like time and CPU stuff. Do we have access to stuff like the epoch and CPU time?

ricfehr3: The only source of timing we have via the cpu is the hardware timers themselves. We could keep a 1x prescaler timer going (just like what the profiler stuff does under the hood #380). That is actually a very precise timer, counting each cpu clock. So the time spent in the menu could be a source of randomness. 

I was thinking that maybe we could also XOR the controls register with itself every frame. So the inputs to the random function generation can be related to all previous inputs and the timer could give randomness to how long it took to put the inputs in.

MathisMartin31: I successfully captured the clock cycle count from game start using the same method as the profiler demo you linked, what do you suggest I can do with that?

MathisMartin31: @ricfehr3 I've been reading stuff a bit, why not take that clock count and pass it through some 32-bits hash function to try and distribute it as much as we can, and use the result as the new seed?

The [FNV-1a](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) looks simple enough to implement

MeirGavish: Sounds like a good idea, we really need to fix our RNG. Just make sure the profiler still works if you use the timer.
From the `profile_start()` documentation
> Routine uses timers 3 and 3; if you're already using these somewhere, chaos is going to ensue.

MathisMartin31: @MeirGavish Oh, I didn't think of that, right now it's always running in the background so I can get the total cycles elapsed ><. Maybe I can start it at the beginning of a menu and stop it when we recover the value and that's it.
When is the profiler actually running?

MathisMartin31: Okay so what I did is that I have a function to start profiling, one to stop it in case we end up not using it, and a function that stops the profiling but uses that value to generate a seed through the hashing function.

This way, we can start the profiling when booting the game and in `game_over_init`, and when starting a run we either generate a seed (always done for now), or discard the counter and stop the profiling if we choose to start with a custom seed (not implemented yet, in the works for #482).

With this the profiler is running only during the Main, Options and Seed menus, or Game Over screen

MathisMartin31: Well now that I think about it the hashing function changes nothing, since it just maps the value I give it to another in a deterministic manner :facepalm: 

I just deleted it and the randomness feels better smh

ricfehr3: Ah, that's good to know for the hash function. I happened to be trying to break "Gamble with Your Friends" when you made this PR and was blindly suggesting things I had just learned from that games decompiled code. So, I'm right there with ya, haha.

As for the profiler thing. There are 4 timers on the GBA 0,1,2,3. We definitely use one for audio (0): https://maxmod.org/ref/tut/streaming.html , and we use two more for the profiler (2 and 3).

I think we should use timer 1 with 1x prescale. It would make sense not to use the `profiler` functions here as well, since we aren't profiling, just taking advantage of the fact that it's a thin wrapper around a timer.

source/game.c
```diff
@@ -4243,8 +4230,8 @@ static void game_shop_on_exit(void)
 
 void game_start(void)
 {
-    // set_seed(9); // 9 is a full house
-    set_seed(g_game_vars.rng_seed);
+    // rng_set_seed(9); // 9 is a full house
```
ricfehr3: You can delete this comment now. 9 is no longer a full house :(

source/random.c
```diff
@@ -0,0 +1,39 @@
+#include "random.h"
+
+#include "game_variables.h"
+
+#include <stdlib.h>
+#include <tonc.h>
+
+void rng_start_sampling(void)
+{
+    profile_start();
```
ricfehr3: https://github.com/GBALATRO/balatro-gba/pull/476#discussion_r3238540524

ricfehr3 (changes requested):
This looks great @MathisMartin31 ! I am requesting changes to the use of the `profiler` functions to be replaced with our own (more accurately named) implementation. See the comment for details. Otherwise, looks good. 😎 Nice work!

source/random.c
MathisMartin31: Ooookay I found some doc on this and was able to use timer 1 ^^
https://gbadev.net/tonc/timers.html

MathisMartin31:
@ricfehr3 Just a reminder to take another look at this when you get some time, no rush don't worry just saying it's here :3

include/random.h
```diff
@@ -0,0 +1,48 @@
+/**
+ * @file random.h
+ *
+ * @brief Common functions to handle RNG manipulation.
```
ricfehr3: I think this file should have an explanation in the docs to ***why*** we added this interface into getting random numbers. Detailing the operation of the `game_step` plus the benefits of the system: 

- Accommodate base-36 for the seed screen input
- Prevent save spamming (shop spawns, hands dealt, boss blinds)
- Give a deterministic RNG system for people play with via shared saves.

source/random.c
```diff
@@ -0,0 +1,39 @@
...
+void rng_shuffle_seed(void)
+{
+    rng_set_seed(timer_acc);
+}
+
+u32 rng_get_u32(void)
+{
+    g_game_vars.rng_step++;
+    return rand();
+}
```
ricfehr3, line 39: The game won't load with the same rng because `load_game()` in `save.c` doesn't seed with `srand()`. I think you should:

1. Make a new struct `RngInfo` that will replace `rng_step` and `rng_seed` in the  `GameVariables` struct:
```c
typedef struct 
{
    u32 seed;
    u32 step;
    ... anything else
} RngInfo;

// Then: g_game_vars.rng_step is now g_game_vars.rng_info.rng_step
```
2. make another function, `restore_rng(RngInfo info)`. 

With that, the RNG can be restored from the data in the `RngInfo` struct, so the function calls don't get modified even if the implementation changes. Also, the struct can be saved to the savefile easily if it changes.

ricfehr3 (changes requested):
Hell yeah this is looking good. I left some comments for suggestions. 

Also, we need tests for this to ensure that the random number generation from saves is deterministic. This can be left to another issue. You can do it here if you'd like, but I feel it's a big task, especially when we are getting closer to a release. I'd like to do the tests correctly, even if that means deferring it while we work on other features.

include/random.h
MathisMartin31: Yeah I agree, although the actual implementation of the base-36 stuff will come with #482

source/random.c
MathisMartin31: > The game won't load with the same rng because load_game() in save.c doesn't seed with srand().

Whoops, I'll take a look at it

>I think you should:
>
> 1. Make a new struct RngInfo that will replace rng_step and rng_seed in the GameVariables struct:
>```
>typedef struct 
>{
>    u32 seed;
>    u32 step;
>    ... anything else
>} RngInfo;
>
>// Then: g_game_vars.rng_step is now g_game_vars.rng_info.rng_step
>```
>2. make another function, restore_rng(RngInfo info).
>
>With that, the RNG can be restored from the data in the RngInfo struct, so the function calls don't get modified >even if the implementation changes. Also, the struct can be saved to the savefile easily if it changes.

That makes sense to me

MathisMartin31:
> Also, we need tests for this to ensure that the random number generation from saves is deterministic. This can be left to another issue. You can do it here if you'd like, but I feel it's a big task, especially when we are getting closer to a release. I'd like to do the tests correctly, even if that means deferring it while we work on other features.

I don't really know how we could achieve that, apart from testing seeds by hand and resetting a few times see if they give the same results 🤔.

I think this can be its own issue so we can further discuss this point in particular

ricfehr3:
> > Also, we need tests for this to ensure that the random number generation from saves is deterministic. This can be left to another issue. You can do it here if you'd like, but I feel it's a big task, especially when we are getting closer to a release. I'd like to do the tests correctly, even if that means deferring it while we work on other features.
> 
> I don't really know how we could achieve that, apart from testing seeds by hand and resetting a few times see if they give the same results 🤔.
> 
> I think this can be its own issue so we can further discuss this point in particular

Yeah, there is a way to test it, but it's sort of a project in itself. I think we can make it an issue and safely use the current implementation since it seems to work well.

include/random.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @brief Get the next "randomly" generated number in the sequence from the current seed.
+ *
+ * @return u32
+ */
+u32 rng_get_u32(void);
+
+/**
+ * @brief Restore RNG info struct in the GameVariables. Sets the `seed` and seeks the
```
ricfehr3, line 67: Please note this as untested right now.

ricfehr3 (approved):
Sweet! Thanks for getting that little bit of infrastructure to handle the game save/load for the random number generators. Besides the one comment about noting the restoration function is untested (and I need to make a ticket as well to track it), I think it's good! Great work as always :) . I'll leave this up for a day and give it one last go through and give others a chance to review. Thanks again!

include/random.h
MathisMartin31: There you go ^^

source/random.c
```diff
@@ -0,0 +1,50 @@
+#include "random.h"
+
+#include "game_variables.h"
+
+#include <stdlib.h>
+#include <tonc.h>
+
+// Accumulate timer 1 into a bigger variable so we can generate more diverse seeds
+u32 timer_acc = 0;
```
ricfehr3:
```suggestion
static u32 timer_acc = 0;
```

include/random.h
```diff
@@ -0,0 +1,68 @@
...
+ * 2. Prevent the players from engaging in save-scumming since the pseudo-RNG is deterministic
+ * 3. Accomodate for base-36 seeds, so that we can freely chose one with alpha-numeric characters in
+ * the seed screen input
+ */
+#ifndef RANDOM_H
+#define RANDOM_H
+
+#include <tonc.h>
+
+#define MAX_SEED 0x81BF0FFF // Hex value of "ZZZZZZ" in base 36
```
ricfehr3:
```suggestion
/**
 * @brief Hex value of "ZZZZZZ" in base 36
 */
#define MAX_SEED 0x81BF0FFF
```

So the comment can show up in doxygen

include/random.h
```diff
@@ -0,0 +1,68 @@
...
+
+typedef struct RngInfo
+{
+    u32 seed;
+    u32 step; // Position in the rng sequence.
+} RngInfo;
```
ricfehr3:
```suggestion
/**
 * @brief Information to track and restore RNG state
 */
typedef struct
{
    /** Initial seed */
    u32 seed;
    /** Position in the rng sequence. */
    u32 step; 
} RngInfo;
```

Similar change for doxygen. I think this struct is worth documenting since the `step` variable needs a little explaining.

ricfehr3 (changes requested):
I'm just about ready to merge this but I spotted a couple small changes. When they are good, I'll merge. Thanks again, @MathisMartin31 !

MathisMartin31:
@ricfehr3 It's ready to go ^^

ricfehr3:
Thanks for getting those last changes. Thank you again for your work on another section of the refactor. The final gauntlet of `round.h/c` is close! Great work :100:

PR #475, opened 2026-05-07
Refactor/shop

MathisMartin31:
Closes #467 

~I'll put it up as a draft for now because I need to clean things up and document functions, but at least it's there~
All good now ^^

PR #479, opened 2026-05-10
Refactor/round end

### LINKED ISSUE #466: Create file for game round end screen `round_end.h`

ricfehr3:


ricfehr3:
@MeirGavish Would you mind requesting a review from copilot for this one?

source/game.c
```diff
@@ -1567,21 +1476,26 @@ void display_round(void)
...
-static void display_hands(int value)
+void display_hands(int value)
 {
     // tte_erase_rect_wrapper(HANDS_TEXT_RECT);
-    tte_printf("#{P:%d,%d; cx:0xD000}%d", HANDS_TEXT_RECT.left, HANDS_TEXT_RECT.top, hands); // Hand
+    tte_printf(
+        "#{P:%d,%d; cx:0xD000}%d",
+        HANDS_TEXT_RECT.left,
+        HANDS_TEXT_RECT.top,
+        g_game_vars.hands
```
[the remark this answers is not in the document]
ricfehr3: I think I'll just remove the argument completely.

include/game_variables.h
```diff
@@ -49,6 +52,12 @@ typedef struct
     enum BlindType next_boss_blind;
     enum BlindState blinds_states[NUM_BLINDS_PER_ANTE];
 
+    int hands;
+    int discards;
+    int score;
```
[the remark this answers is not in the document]
MeirGavish: Yeah, this is true, it was changed to u32 specifically so the extra bit can be used to store larger scores and there's code that relies on that...

source/game.c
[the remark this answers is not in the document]
MeirGavish: Yeah, I think it makes sense. I think the only display functions that really need the parameter are `score` and `temp_score` so there's a bit of inconsistency there but I guess it's fine. Maybe in the future we can make a refactor to separate the functions to `display_*(value)` and a wrapper `display_curr_*(void)` if we need to.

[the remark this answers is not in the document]
ricfehr3: I agree with that. It looks like we weren't using the argument before so it's probably safe to drop. We can revisit it later and add it, but I like keeping it simple for now.

include/game_variables.h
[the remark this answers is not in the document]
ricfehr3: Oops, yes this should be a u32.

include/game/options_menu.h
```diff
@@ -6,8 +6,6 @@
 #ifndef GAME_OPTIONS_MENU_H
 #define GAME_OPTIONS_MENU_H
 
-#include "game_variables.h"
```
MathisMartin31, line 9: Good catch ^^

source/game/round_end.c
```diff
@@ -0,0 +1,473 @@
...
+        {
+            reset_top_left_panel_bottom_row();
+        }
+        else if (g_game_vars.timer == 2)
+        {
+            int y = 5;
+            memset16(&se_mem[MAIN_BG_SBB][32 * (y - 1)], 0x0001, 1);
+            memset16(&se_mem[MAIN_BG_SBB][1 + 32 * (y - 1)], 0x0002, 7);
+            memset16(&se_mem[MAIN_BG_SBB][8 + 32 * (y - 1)], 0x0401, 1);
+        }
```
MathisMartin31, line 233: So I gather that you didn't understand what is going on here either ^^'
Who even wrote this in the first place?

source/game/round_end.c
```diff
@@ -0,0 +1,473 @@
...
+void game_round_end_on_exit(void)
+{
+    // Cleanup blind tokens from this round to avoid accumulating
+    // allocated blind sprites each round
+    blind_reward = 0;
+    hand_reward = 0;
+    interest_reward = 0;
+    sprite_destroy(&g_game_vars.playing_blind_token);
+    sprite_destroy(&g_game_vars.round_end_blind_token);
+    // TODO: Reuse sprites for blind selection?
```
MathisMartin31, line 472: I think we can remove this comment, the way the next boss blind is rolled right now depends on the fact that this token is not the same as the one in the blind select screen

source/game/round_end.c
ricfehr3: I remember seeing this and thinking, "I will fix it", then well... I sort of kept going :joy: . Let me give it a good try here. 

This is from an early refactor done by @drewjohnson2 . It was a huge cleanup step but we left the old TODO stale. Previously it was picked from wherever it used to live in `game.c`. Definitely time to clean it up. I'll do that.

MathisMartin31: I mean, it's really dirty but it works, so if you can't make make sense of it don't push it and delay your PR for this ^^'

ricfehr3: I think I'll leave it to an issue, so we can keep the refactor pace going.

ricfehr3:
Looking to merge this tomorrow if there's no more reviews. Then I will move onto merging #475 and finish #470

MeirGavish:
> Looking to merge this tomorrow if there's no more reviews. Then I will move onto merging #475 and finish #470

I will probably not get around to fully review this but it looks good in general, it seems like the refactor is going well.

ricfehr3:
> > Looking to merge this tomorrow if there's no more reviews. Then I will move onto merging #475 and finish #470
> 
> I will probably not get around to fully review this but it looks good in general, it seems like the refactor is going well.

No worries. I'll merge this and log that issue that @MathisMartin31 brought up. We can always revisit these files later. It's going to be a lot easier after the refactor.

PR #486, opened 2026-05-16
Update getdata() to use get_flattened_data()

SuperbMuffin:
getdata() is being deprecated in pillow, and it's throwing a warning, this would change it to get_flattened_data(). I also removed the list() wrapping around that cause it's not really needed; we aren't using any of the list features.


SuperbMuffin:
Crap, I forgot the CI ran on a different version of PIL oops

PR #491, opened 2026-05-20
[Bug] Fix applying boss blind color twice

MathisMartin31:
So I encountered a weird bug that has flown under the radar for a long time despite being quite obvious ^^'

By calling `change_background(BG_CARD_SELECTING, false)` both at the end of the Blind Select screen, and at the beginning of the Playing state, we effectively applied the boss blind color twice to the affine background.

Notice how the background first changes color, then suddenly gets darker:

https://github.com/user-attachments/assets/6c3bc024-1008-43e7-8146-6bfb404129a7

I fixed it by applying the boss blind color to the background outside of the `change_background_legacy` function, at the end of the blind select screen if next blind is a boss/showdown. The background is significantly lighter as a result, with better colors:

https://github.com/user-attachments/assets/ba262584-db29-433f-8ae7-a2a6755c480a

### DIFF
--- a/include/blind.h
+++ b/include/blind.h
@@ -103,8 +103,6 @@ typedef struct
 
 void blind_init();
 
-void blind_set_boss_graphics(const unsigned int* tiles, const u16* palette);
-
 u32 blind_get_requirement(enum BlindType type, int ante);
 int blind_get_reward(enum BlindType type);
 u16 blind_get_color(enum BlindType type, enum BlindColorIndex index);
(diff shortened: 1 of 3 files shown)


ricfehr3 (approved):
Nice catch! Going to merge this. The change is simple and _somewhat_ crucial. Thanks!

MeirGavish:
Interesting, I would have expected applying the same color a second time would not do anything, maybe the function to change the affine background color needs to be examined too?

ricfehr3:
> Interesting, I would have expected applying the same color a second time would not do anything, maybe the function to change the affine background color needs to be examined too?

Totally fair to note. I'll open an issue to track it.

PR #492, opened 2026-05-21
Add logging

### LINKED ISSUE #167: Support for Error Logging
It could be very useful to generate logs when the game is running so whenever the code encounters errors they can be logged and more easily debugged.
However, I don't know how this could be done on the GBA.
I still think this is worth looking into, even if it's only on emulator runs and not on hardware.

ricfehr3:
Closes #167 

I was talking silly previously -> https://github.com/GBALATRO/balatro-gba/issues/167#issuecomment-3392751528

Error logging is great. I really should have explored this earlier. @MathisMartin31 @MeirGavish  Let's prioritize this if y'all like it. I think it's going to be handy immediately.

Usage:

```c
MGBA_WARN("This is a test")
```

```sh
MGBA_LOGGING=1 make
```

Then, you can enable logging via the `-l` or `--log-level` option:

```sh
mgba -l 7 game.rom
```

### DIFF
--- a/include/mgba_logger.h
+++ b/include/mgba_logger.h
@@ -29,11 +29,11 @@
 
 typedef enum
 {
-    MGBA_LOG_FATAL = 0,
-    MGBA_LOG_ERROR = 1,
-    MGBA_LOG_WARN = 2,
-    MGBA_LOG_INFO = 3,
-    MGBA_LOG_DEBUG = 4,
+    MGBA_LOG_FATAL,
+    MGBA_LOG_ERROR,
+    MGBA_LOG_WARN,
+    MGBA_LOG_INFO,
+    MGBA_LOG_DEBUG,
 } MgbaLogLevel;
 
 /**
--- a/include/mgba_logger.h
+++ b/include/mgba_logger.h
@@ -1,0 +1,70 @@
+/**
+ * @file mgba_logger.h
+ *
+ * @brief Interface to interact with the mgba logger.
+ *
+ * Use with mgba in the command line.
+ *
+ * You can pass which logs you'd like to print with a flag passed to mgba
+ * with `-l` or `--log-level`.
+ *
+ * | 7 | 6 | 5 |   4   |   3  |   2  |   1   |   0   |
+ * |---|---|---|-------|------|------|-------|-------|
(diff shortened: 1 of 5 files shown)


MeirGavish:
Yes! I should have explored this earlier myself

ricfehr3:
I surrounded the implementations with `#ifdef` so no debug logging code is compiled at all. 
I'm happy to merge this whenever.

include/mgba_logger.h
```diff
@@ -0,0 +1,71 @@
...
+#include <stdbool.h>
+
+typedef enum
+{
+    MGBA_LOG_FATAL = 0,
+    MGBA_LOG_ERROR = 1,
+    MGBA_LOG_WARN = 2,
+    MGBA_LOG_INFO = 3,
+    MGBA_LOG_DEBUG = 4,
+} MgbaLogLevel;
```
MathisMartin31, line 36: Small nitpick here, but I think we should stick to the same style as the rest of the code and not put `=0`, `=1`, etc, when enum values start at 0 and are consecutive.

source/game.c
```diff
@@ -629,6 +630,12 @@ static inline void jokers_update_loop(void)
 
 void game_update()
 {
+    MGBA_DEBUG("GBAlatro: Debug");
+    MGBA_INFO("GBAlatro: Info");
+    MGBA_WARN("GBAlatro: Warn");
+    MGBA_ERROR("GBAlatro: Error");
+    MGBA_FATAL("GBAlatro: Fatal");
```
MathisMartin31: Is it intentional that these messages will be printed on every frame?

MathisMartin31:
Very cool stuff, thanks for this! Debugging will be so much easier ><

Just a couple of questions, but otherwise it's good to go for me

include/mgba_logger.h
ricfehr3: That sounds good to me!

source/game.c
ricfehr3: Oops! This was for testing, I'll remove it.

ricfehr3:
I'm going to merge this one. I'd like to start using it personally.

PR #500, opened 2026-05-23
Move data structures into their own directory

ricfehr3:
First of many smaller PRs to not overwhelm us when reviewing #465 . 

Here, I've moved data structures into their own directories in `include/data_structures` and `source/data_structures`. 

This updates the contributing docs and CI scripts for clang format, now more generic to support more directories and subdirectories.

Tests are updated as well for their makefiles to point to the correct locations.

@MeirGavish @MathisMartin31 **NOTE**: I'm staging to a new branch `refactor-round-staging`. What I'm going to do for this one crazy refactor is make a branch where we can see changes, but won't have massive conflicts between eachother. So, I may be merging these extra fast, but they will be as small of deltas as I can do. Hopefully this keeps everything stable as this big chunk of code is separated.


MathisMartin31 (commented):
Did no file including these headers need to have their include path updated? Just like we include `"game/foo.h"`, should the includes for the data structures become `"data_structures/bar.h"`?

ricfehr3:
> Did no file including these headers need to have their include path updated? Just like we include `"game/foo.h"`, should the includes for the data structures become `"data_structures/bar.h"`?

The `-I` flag was updated to include `[source,include]/data_structures`. `-I` lets the compiler know all files relative to what is passed to it. 

It's purely a style choice. I think using it for `game` files makes sense, but we don't need a "hint" that a `list.h` is anything but a data structure. But we can chose to do it the other way for consistency.

I say, only do it for `game` right now. But I don't feel have any strong feelings about it.

MathisMartin31:
Sounds good to me!

MeirGavish (approved):
Looks good to me

MathisMartin31 (approved):

ricfehr3:
Thanks for the reviews!

PR #505, opened 2026-05-24
Prevent bad colors in case `affine_background_set_color` is called twice by mistake

### LINKED ISSUE #493: Investigate Affine Background Function Behavior
We found that we were setting the background color twice on boss blinds, making a color that was 2x as dark/saturated. See #491. 

In #491, the double call was fixed, but setting the color twice should change to the same color twice, not apply the color modifier on top of the previous color modifier.

MathisMartin31:
Closes #493

@MeirGavish @ricfehr3 I believe this was the way to go, based on our discussion in the related issue?

I tested it by duplicating the call to `affine_background_set_color` and the colors were good.

### DIFF
--- a/source/affine_background.c
+++ b/source/affine_background.c
@@ -63,11 +63,13 @@ void affine_background_update()
 
 void affine_background_set_color(COLOR color)
 {
-    // Reload the palette to reset any previous color scaling
+    // Reload the palette to reset any previous color scaling.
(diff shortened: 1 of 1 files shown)


ricfehr3 (approved):
Great! Looks good to me. I gave it a test and tried to call the function a few times, and the bug seems to be handled.

MeirGavish (approved):
Yep, that's what we talked about, looks good.

PR #495, opened 2026-05-21
Add option to enable a more readable deck, suggestion by @Turro75

MathisMartin31:
@Turro75 @MeirGavish @ricfehr3 
From our discussion in #488, here my version in action ^^

https://github.com/user-attachments/assets/47a0fee6-9d09-4891-a68f-464dd928ba3f

And here's the one from Meir ^^

https://github.com/user-attachments/assets/844d9e9f-a1ff-4df2-86dc-a6fe7f2d172f

PR #511, opened 2026-05-25
Add option to enable a more readable deck, suggestion by @Turro75 (#495)

ricfehr3:
Apologies for this mess. I didn't know how github worked when reverting and I need to recommit changes for #495 . This should add the proper contributions for github, though, if this happens again I'll need to think of a better way :joy:


ricfehr3:
@MathisMartin31 FYI

Issue #498, opened 2026-05-23
Create and use stack implementation from list


ricfehr3:
We use a few stacks as arrays with a tracker of the stack's `top`. They all only hold typed pointers (still pointers, so same size), and have their own `push` an `pop` functions each.

This functionality should be moved into a dedicated data structure.

A `List` could be used, but it is a heavy interface.

MeirGavish:
I like implementing stacks with a linked list since everything is done at O(1) and the size is theoretically unbounded. However some of these stacks may need to be implemented as lists since they actually have arbitrary index accesses...
Maybe the deck itself is the best candidate for stack implementation.

ricfehr3:
> I like implementing stacks with a linked list since everything is done at O(1) and the size is theoretically unbounded. However some of these stacks may need to be implemented as lists since they actually have arbitrary index accesses...
> Maybe the deck itself is the best candidate for stack implementation.

I like that, Meir. I was working on a dedicated stack implementation but I quickly realized that I was rewriting a lot of list functionality. I agree also that since the access is always linear then it should probably just use the existing and proven list.

Access will be a bit slower, but since we know we aren't reaching the vsync time limit at all right now I think we can take structure over efficiency for these lists with less than 100 elements max, if ever.

I also think that we should reuse the list now as well, because I think we can standardize the list operations and offer a static array or linked list backend. This would be a future ticket well after the release, but if we use lists for everything then we can swap out the backend for individual ones. We may or may not do that, but using a list let's that option be a lot more viable.

With that... I still want to properly do this piecewise. So I'll go for the following:

1. Replace stacks with lists in one pr (for the next release)
2. Implement those "stacks" now "lists" as proper "decks". Once again, I'm going to reference #481 as we need to take this into that design.

ricfehr3:
> I like implementing stacks with a linked list since everything is done at O(1) and the size is theoretically unbounded. However some of these stacks may need to be implemented as lists since they actually have arbitrary index accesses... Maybe the deck itself is the best candidate for stack implementation.

@MeirGavish , I think we can take on the technical debt, if you think this issue is a good idea:
#509 . Basically, I think we can support a static array or a linked list backend into our `list.h` interface. So we can confidently use less efficient linked lists temporarily, so long as they don't cause issues.

Issue #481, opened 2026-05-12
Define Generic "Card" -> "Deck" Abstractions to support all card types


ricfehr3:
Currently, there are only jokers and regular playing cards implemented. This has our code conveniently split between Joker only deck functions (that is what is displayed on the screen), and card only deck functions. To support other types of cards and to factorize the logic of rendering and handling cards, a more generic approach needs to be explored. 

This will require exploring the design of abstracting the idea of Cards and Decks. This issue is meant to track that design and implementation.

[Related discussion](https://github.com/GBALATRO/balatro-gba/pull/475#discussion_r3223945719)

MeirGavish:
I'm just quoting the discussion here

@ricfehr3 
> @MathisMartin31 Do you think this structure of interacting with jokers via this file could apply to the regular deck as well? We don't have card upgrades or anything yet, but it would make sense to me that the interface to handle the jokers row will eventually be the same as the deck row when selling/removing/upgrading cards.
> 
> This can be a task for later, but what do you think of describing this functionality more as, "Card Row", and instead pass a configuration of what "Deck" to draw from and where it will be rendered. (This also assumes that a generic "Deck" will hold either jokers or playing cards, or whatever else). This will require some discussion, but looking toward the future this abstraction makes sense to me right now.

@MathisMartin31 

> I think it might, but we won't be able to have a single function for all Card lists. We may be able to factorize some code to use in some jokers_sel_row_..., cards_sel_row_..., consumables_sel_row_... and shop_sel_row_... functions.
> 
> I was thinking about this the other day: right now, the implementation of the Shop works because we only have Jokers available to buy, but we need to be able to mix Consumables and Playing Cards in that row to have a functional shop.
> 
> I think this has to be a task for later, to be done when implementing Tarot, Planet and Spectral cards so that we can actually buy them.
> 
> For this we may need to create an "AbstractCard" struct holding a CardType and pointer to either a PlayingCard, Joker, etc

So, a few thoughts. First off, I don't think "deck" is a good generic name, because the deck very specifically means the deck of playing cards and it doesn't behave at all like jokers or consumables. For one, playing cards can't be sold, and they can't be "reordered" inside the deck. I think the word "deck" should be reserved solely for the deck of playing cards and maybe `CardType` would be better to use.

Also, `Card` in the code currently specifically means playing card, it can (and probably should) be refactored and renamed to `PlayingCard` and that could free up `Card` for a more generic use.

[...]

PR #514, opened 2026-05-28
Introduce card instance

ricfehr3:
I found myself trying to change one thing and finding a couple small things that needed to change, and a couple more from those, all the way down. After a few tries, I think the first thing to do is restructure the `Card` `CardObject` system just a bit. It turns out, that this not only to supports generalized assets: #433 #431 #481, but also supports a generalized hand/pile/list of cards #498. #498 Will support in decoupling a lot of the systems that are a bit spaghetti in `game.c`, letting the rest of #465 be a lot easier to complete. 

This change is the smallest incremental step I could think of introducing. My main goal is to finish the `round` refactor and move onto completing the release. But, this also sets up for thos future issues:

The plan is to generalize all card types (right now only jokers and playing cards, but eventually tarot, planet, etc.) into one `CardInstance` that has a `void* data` pointer and functions that define it's behavior, guarded by an ENUM type. But yeah, that's in the future.


MathisMartin31:
> The plan is to generalize all card types (right now only jokers and playing cards, but eventually tarot, planet, etc.) into one CardInstance that has a void* data pointer and functions that define it's behavior, guarded by an ENUM type. But yeah, that's in the future.

Yeah I was thinking about it when working on the Jokers descriptions. Since all types of cards have a name, description and a rarity/type, I think this data should go into the CardInstance so that the description mechanism can work for any card/pack/voucher in the shop

ricfehr3:
> > The plan is to generalize all card types (right now only jokers and playing cards, but eventually tarot, planet, etc.) into one CardInstance that has a void* data pointer and functions that define it's behavior, guarded by an ENUM type. But yeah, that's in the future.
> 
> Yeah I was thinking about it when working on the Jokers descriptions. Since all types of cards have a name, description and a rarity/type, I think this data should go into the CardInstance so that the description mechanism can work for any card/pack/voucher in the shop

I think that is an important thing to consider. A lot of the joker functionality will be shared to other card types. 

This PR was done while I was deep in the refactoring of `game.c`. There's some similarity to what @MeirGavish has going on with #534 , that may supersede this. But its good to note the connection as we close in on the design.

ricfehr3:
Closing this. Plans have changed a lot and we are already in progress of a different direction.

PR #516, opened 2026-05-29
Refactor/simplify card select

ricfehr3:
I have started to tear down `game.c` probably a dozen times now, haha. This PR is the another step toward the `round.c` refactor: #465 . I think I have a handle on the sequence of this now. I'll be leaving this PR up for a day before merging into my refactor branch.

The TLDR of this is, we have a function: `select_cards_in_played_hand()` that "selects" the card objects in a played hand to move upwards and count toward scoring with the shake animation, etc. The function below can act as a generalized form for everything besides highcard

```c
static inline void select_match(u32 num)
{
    u32 matches[NUM_RANKS] = {0};

    for(int i = 0; i <= played_top; i++)
    {
        matches[played[i]->card->rank]++;
    }
    
    for(int i = 0; i <= played_top; i++)
    {
        if(matches[played[i]->card->rank] >= num)
            card_object_set_selected(played[i], true);
    }
}
```

Currently, the card-array-offset-stack thing needs to be reworked to #498 . A lot of these functions will need to be refactored to decouple the systems. I'll be posting more PRs soon doing what I did here across other parts that use the card stacks.


source/game.c
```diff
@@ -1580,135 +1580,96 @@ static inline void select_flush_and_straight_cards_in_played_hand(void)
...
             card_object_set_selected(played[i], true);
-        }
     }
-}
 
-static inline void select_three_of_a_kind_cards_in_played_hand(void)
-{
-    // find three cards with the same rank
-    for (int i = 0; i <= played_top - 1; i++)
+    /*
```
ricfehr3, line 1609: Please note, I'm leaving these comments in because they will be used after the `played` array is moved to a `list`.

source/game.c
```diff
@@ -2350,7 +2311,8 @@ static inline void select_cards_in_played_hand()
         case FLUSH_HOUSE:
             /* FALL THROUGH */
         case FLUSH_FIVE: // Select all played cards in the hand
-            select_all_five_cards_in_played_hand();
+            // Match all 1-pairs, haha
+            select_match(1); 
             break;
```
MathisMartin31, line 2272: I'm guessing your `select_match(n)` works because we've already computed what hand type we have played? Which means that selecting both a Pair and Two Pair with `select_match(2)` will work because in the Two Pair case we know there are exactly 2 pairs to select, but only one in the Pair case?

If so, I think this expectation should appear somewhere in the function's description

source/game.c
```diff
@@ -1580,135 +1580,96 @@ static inline void select_flush_and_straight_cards_in_played_hand(void)
...
-            break;
+    while((obj = list_itr_next(&itr)))
+    {
+        if(matches[obj->card->rank] >= num)
+            card_object_set_selected(obj, true);
     }
+    */
 }
 
 static inline void select_two_pair_cards_in_played_hand(void)
```
MathisMartin31: I'm not seeing any calls to this function anymore, with the new `select_match(n)` function. Is it necessary to keep it, or are you waiting until the end of the round.c refactor to be sure we can get rid of it?

source/game.c
```diff
@@ -1580,135 +1580,96 @@ static inline void select_flush_and_straight_cards_in_played_hand(void)
     }
 }
 
-static inline void select_all_five_cards_in_played_hand(void)
+static inline void select_match(u32 num)
```
MathisMartin31, line 1594: Add doxygen doc with brief explanation

source/game.c
MathisMartin31, line 2233: Add doxygen doc with brief explanation

MathisMartin31 (commented):
Just a few comments, they're less about functionnality since I trust you've tested the changes, but more about the presentation of the code.

Overall, it does look simpler and more readable, thanks for the change ^^

source/game.c
```diff
@@ -1580,135 +1580,96 @@ static inline void select_flush_and_straight_cards_in_played_hand(void)
...
-            break;
+    while((obj = list_itr_next(&itr)))
+    {
+        matches[obj->card->rank]++;
     }
 
-    for (; i <= played_top - 1; i++) // Find second pair
+    */
+
+    bool found_one_pair, found_two_pair = false;
```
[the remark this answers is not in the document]
MathisMartin31: Aren't booleans initialised to 0/false by default? Still good to write it explicitly though

source/game.c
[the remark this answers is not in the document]
MathisMartin31: I had the same thought lol ><

Also on a related note, Copilot keeps bringing up these errors that should make the build fail, but never pop up. We do have `-Wall -Werror` in the Makefile, so what gives?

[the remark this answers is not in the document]
ricfehr3: I... Don't know. I'll delete this as it is unused, but that is probably worth an issue to track.

source/game.c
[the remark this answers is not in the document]
ricfehr3: Deleting this section, but I agree. I added it to the style guide.

source/game.c
ricfehr3: > I'm guessing your select_match(n) works because we've already computed what hand type we have played? 
**Yep, you got it**, that logic being separated is worth a whole other refactor :) I just noted this, I would like to have all this calculated once eventually : #516 . 

This function is only used to select card objects to be scored, animated, and discarded. `get_hand_type()` only returns one type, So if it already is a pair, then there is no 3 of a kind, and there is no full house. 

`select_match(2)` will match all cards that have a pair. So if it's already a pair, it will get one, if it's a two pair, it will get both sets. 

I think this should be documented better because that is some weird functionality.

source/game.c
```diff
@@ -1580,136 +1580,50 @@ static inline void select_flush_and_straight_cards_in_played_hand(void)
...
-                card_object_set_selected(played[i], true);
-            }
-        }
-    }
-    else // If there are only 4 cards selected we know they match
+    for (int i = 0; i <= played_top; i++)
     {
-        for (int i = 0; i <= played_top; i++)
-        {
+        if (matches[played[i]->card->rank] >= num)
```
MathisMartin31, line 1605: A thought I had, but since the expectation is that we call this with `num == 2` for a Pair, couldn't this condition be `matches[...] == num`?

It shouldn't happen, but as it is it could also select *everything above* a Pair.

It's a nitpick though ^^

source/game.c
```diff
@@ -2323,16 +2240,16 @@ static inline void select_cards_in_played_hand()
...
             break;
         case PAIR:
-            select_pair_cards_in_played_hand();
-            break;
+            select_match(2);
+            /* FALL THROUGH */
         case TWO_PAIR:
-            select_two_pair_cards_in_played_hand();
+            select_match(2);
             break;
```
MathisMartin31, line 2247: Shouldn't the Two Pair's call to `select_match(2)` already select the two sets of 2 cards of the same rank?

I feel like this fall through from the Pair will do a second call for nothing, I think what you wanted to do was take out the call to `select_match` from the Pair case?

MathisMartin31 (commented):
Some new comments, nothing major ^^

PR #521, opened 2026-05-30
Refactor - Moved shop data structures to `shop.c/.h`

MeirGavish:
Since `_shop_jokers_list` and `_avail_jokers_bitset` are mainly used by the shop, it makes more sense to put them in `shop.c/.h`  and have the shop module manage them instead of having them in `game.c/.h` with getters for the shop.

### DIFF
--- a/source/game/shop.c
+++ b/source/game/shop.c
@@ -63,6 +63,9 @@ static const Rect     SHOP_REROLL_RECT      = { 88,  96, UNDEFINED, UNDEFINED };
 static const BG_POINT SHOP_JOKER_SPRITES_INIT_POS = {120, 160};
 // clang-format on
 
+static List _shop_jokers_list = LIST_DEFAULT;
+BITSET_DEFINE(_avail_jokers_bitset, MAX_DEFINABLE_JOKERS)
+
 enum GameShopStates
 {
     GAME_SHOP_INTRO,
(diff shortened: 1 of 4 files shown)


source/game/shop.c
```diff
@@ -136,6 +139,32 @@ static int timer;
...
+    int num_jokers = get_joker_registry_size();
+
+    bitset_clear(&_avail_jokers_bitset);
+    for (int i = 0; i < num_jokers; i++)
+    {
+        bitset_set_idx(&_avail_jokers_bitset, i, true);
+    }
+}
+
+void game_shop_init_run(void)
```
ricfehr3: Nitpick: I think this should be renamed `game_shop_reset`, `game_shop_run_reset`, or similar.

source/game/shop.c
```diff
@@ -63,6 +63,9 @@ static const Rect     SHOP_REROLL_RECT      = { 88,  96, UNDEFINED, UNDEFINED };
 static const BG_POINT SHOP_JOKER_SPRITES_INIT_POS = {120, 160};
 // clang-format on
 
+static List _shop_jokers_list = LIST_DEFAULT;
+BITSET_DEFINE(_avail_jokers_bitset, MAX_DEFINABLE_JOKERS)
```
ricfehr3: Should we remove the `_` underscore from the front? I realize I need to adjust the style guide to be consistent, but there is no hard rule on having it or not. We can make it a rule or just not bother.

ricfehr3 (approved):
Great! This looks like a nice refactor opportunity. Thanks for the PR, I left a couple comments, but approved. Feel free to make the changes to your discretion and merge. Thanks @MeirGavish !

source/game/shop.c
MeirGavish: Maybe I can change it to `s_` prefix and we can make it a rule? I think it would be nicer

ricfehr3: Totally, I agree with that.

source/game/shop.c
MeirGavish: I thought init was more appropriate because it initializes and called on init, but I can rename it reset to be more consistent

MeirGavish: I renamed it to reset

MathisMartin31 (approved):
Nice, thanks for this change! It does indeed make more sense to have these variables here.

Approving

MeirGavish:
> Nice, thanks for this change! It does indeed make more sense to have these variables here.
> 
> Approving

Thanks! 

Since it's fully approved and I addressed all comments I'll just merge.

PR #522, opened 2026-05-31
Replaced lose sound effect with original ripped sound effects

### LINKED ISSUE #515: Fix Game Lose/Win SFX
As of #494 we have game over audio sound effects. The sources of the files have smaller "artifacts" from the recording and can be cleaned up. See https://github.com/GBALATRO/balatro-gba/pull/494#pullrequestreview-4363091690

MeirGavish:
closes #515

I kept the sample rate at 44.1kHz in the files, it sounds better and the ROM size seems unaffected, I guess we can just leave the downsampling to MaxMod.

I think it sound pretty much the same only without the background cards sound effect side effect.

Main

https://github.com/user-attachments/assets/ce45bd37-d428-4957-9495-5c92bf104936

PR

https://github.com/user-attachments/assets/1333b467-28a8-4870-a46c-3f53467c4149

Main with music volume at 0

https://github.com/user-attachments/assets/0ee81642-95c2-4127-8ec9-126a9339d079

PR with music volume at 0

https://github.com/user-attachments/assets/70f683c6-742a-468f-bbd6-6deb23ad9c1c


source/game/game_over.c
```diff
@@ -69,7 +71,9 @@ void game_win_on_init(void)
 
 void game_lose_on_init(void)
 {
-    play_sfx(SFX_GAME_LOSE, MM_BASE_PITCH_RATE, SFX_DEFAULT_VOLUME);
+    play_sfx(SFX_NEGATIVE_LOSE, MM_BASE_PITCH_RATE / 2, GAME_OVER_SFX_VOL);
+    play_sfx(SFX_WHOOSH2_LOSE, GAME_OVER_WHOOSH_RATE, GAME_OVER_SFX_VOL);
```
MeirGavish, line 75: I kept the names of the effects as they are in the original Balatro and just added a "lose" suffix to indicate what they are used for in our project.

source/game/game_over.c
```diff
@@ -32,6 +32,8 @@ static const Rect     NEW_RUN_BTN_SRC_RECT       = {  0,  30,   4,  31};
 // clang-format on
 
 static const u32 GAME_OVER_ANIM_FRAMES = 15;
+#define GAME_OVER_SFX_VOL     178 // 70% of MM_SFX_FULL_VOLUME
+#define GAME_OVER_WHOOSH_RATE 922 // 90% of MM_BASE_PITCH_RATE
```
ricfehr3:
```suggestion
static const u32 GAME_OVER_SFX_VOL = 178; // 70% of MM_SFX_FULL_VOLUME
static const u32 GAME_OVER_WHOOSH_RATE = 922; // 90% of MM_BASE_PITCH_RATE
```

`static const` compiles the same as a macro but adds a type (`u32`). I've been preferring it for the explicit types, but I don't love it or hate it after using it. What do you think?

ricfehr3 (approved):
Nice! Good ear on hearing the issue and tuning the pitch and volume. Sounds great.

I just have one comment on style. Otherwise, merge when you're ready.

source/game/game_over.c
MeirGavish: Agreed, I made them defines out of habit. I just need to change the type since it's not `u32`.

PR #526, opened 2026-06-02
Updates to performance script to work with VSCode

MeirGavish:
This isn't really a PR to be merged, it's very rough, I just wanted to get it out there so you can see the changes I made to the script so it runs in VSCode and Windows. I used AI to make modifications to the script and currently don't remember exactly what I did. 
I don't know if the script currently works on Linux and without VSCode though.
This is up to date with the main branch so maybe if you update #380 and add the script there you can better see the differences.

I have the following in my .vscode/launch.json so it runs when debugging.

```json
            "windows": {
                "miDebuggerPath": "${env:DEVKITARM}/bin/arm-none-eabi-gdb.exe",
                "setupCommands": [
                    {
                        "text": "shell start /b \"\" \"${env:ProgramFiles}/mGBA/mGBA.exe\" -g \"${workspaceFolder}/build/balatro-gba.elf\""
                    },
                    {
                        "text": "source ${workspaceFolder}/scripts/performance/test_performance.gdb"
                    }
                ]
            },
```

Sorry this is so rough, I just wanted to get it out there, I will try to maybe add more comments here later.


MeirGavish:
So I just want to document how I manually profile performance using this as I recently repeated this process.

1. Change the working directory from /build to somewhere I want to save the logs
2. Change the the number of iterations of the outer loop from 50 to just 2 `while $i < 50` -> `while $i < 2`.
This allows me to granularly test short timeframes. 
3. Start running a debug session in VSCode and mGBA.
4. Start performing the action I want to profile, usually it's a discard, then immediately pause with **Ctrl+P**.
5. Pause execution in the debugger.
6. In the debug console enable the profiling breakpoint with `-exec enable 1`
7. Unpause mGBA
8. Continue execution in the debugger, it should break in the enabled profiling breakpoint
9. In the debug console perform `-exec print_frame_tick` to print a single cycle count measurement and repeat however many times until the animation is complete so I have the performance measurements for the entire animation.
10. In the debug console disable the profiling breakpoint with `-exec disable 1` and continue execution to continue running normally.
11. Usually I also repeat for several discards to have a bigger log.

PR #531, opened 2026-06-05
Clarify Linux installation instructions

ricfehr3:
Added note about Ubuntu version requirements for installation. A friend recently tried it on PopOS 22.04 and we found out the glibc version is incompatible with (2.36 required, 2.35 provided). This technically is related to only devkitpro, but it couldn't hurt to note here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -239,7 +239,9 @@ Disregard Steps 4-5 and instead click the green code button on the main reposito
 
 ### Linux (Debian/Arch)
 
-These instructions will be only for Debian (**Ubuntu**) and Arch based systems. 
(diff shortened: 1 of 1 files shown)


MeirGavish (approved):
Seems good. I recently got an Ubuntu laptop and managed to set things up with no problems but I'm pretty sure I installed the 24.04 LTS.

ricfehr3:
> Seems good. I recently got an Ubuntu laptop and managed to set things up with no problems but I'm pretty sure I installed the 24.04 LTS.

That adds up. When she updated to 24.04 it all worked fine. There was also an issue with the version of pillow as well. All good, I'll merge this.

PR #533, opened 2026-06-05
Implement SpriteObject pseudo-inheritance via casting

MeirGavish:
So I was reading about the Liskov Substitution SOLID principle and I realized that this is exactly the object-oriented design principle we need to implement a generic object for #481 and implementing planet cards in the shop etc. 
So I was wondering how it could be implemented in a non-object-oriented language such as C so I discussed it with Gemini and apparently it can implement a sort of inheritance.
Back then in the previous discussion I said,
> If we were using an OOP language like C++ I would define a "Purchasable" interface and have the different card types ?implement it. Also as an aside if I could I would have made JokerObject and CardObject inherit SpriteObject rather than contain it as wrappers.
> Right now SpriteObject is used to implement movement and animation code so it's not duplicated between CardObject and JokerObject but it cannot serve as unified interface to shop/container code.

Apparently C has something special in the standard where casting a pointer to a struct gives a valid pointer to its first member.(Described in Section 6.7.2.1.13 of [WG14/N1124](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf)). This comes from the fact structs are just contiguous memory, the nesting is only in the abstract code definitions. That makes it feel a bit like a hack but if it's in the spec, it's safe to use.

So by turning the `SpriteObject` member into a non-pointer member and making it the first one like so,
```c
typedef struct CardObject
{
    SpriteObject sprite_object;
    Card* card;
    bool selected;
} CardObject;
```
```c
typedef struct JokerObject
{
    SpriteObject sprite_object;
    Joker* joker;
} JokerObject;
```

A `JokerObject` or `CardObject` can serve as a `SpriteObject` and you can do the following
```c
JokerObject* joker_object = (JokerObject*)list_get_at_idx(owned_jokers_list, idx);
sprite_object_set_focus((SpriteObject*)joker_object, true);
```
or even
```c
SpriteObject* joker_object = (SpriteObject*)list_get_at_idx(owned_jokers_list, idx);
sprite_object_set_focus(joker_object, true);
```
And you can even downcast if you know for certain that your `SpriteObject` is a `JokerObject` so you can do something like the following.
```c
typedef int (*GetPriceFunc)(SpriteObject* sprite_object);

int get_joker_object_price(SpriteObject* sprite_object)
{
  return ((JokerObject*)sprite_object)->joker->value / 2;
}
```
And now `get_joker_object_price()` can be assigned to the type `GetPriceFunc`.

This inheritance scheme is generally described [here](https://www.state-machine.com/object-based-programming-in-c).

I think this can potentially be very useful and could allow turning the `SpriteObject` into the generic object like I thought. There are downsides to this too since it really means lumping all the responsibilities onto `SpriteObject` unless we decide to do several levels of inheritance which I'm not sure about...
So this PR is still a draft to show that it works but we still need to think the design through and decide whether we want to do it like this.

Also since `SpriteObject` really functions as an abstract class with no pure instances, I could remove it from the pool in this PR.


source/joker.c
```diff
@@ -288,12 +288,12 @@ bool joker_object_score(
...
         // display the text on top of the card instead of below the Joker for Held Cards effects
         // scored_card cannot be NULL here because of the joker event
-        cursorPosX += fx2int(card_object->sprite_object->x);
+        cursorPosX += fx2int(card_object->sprite_object.x);
         cursorPosY = HELD_CARD_SCORE_TEXT_Y;
     }
     else
     {
-        cursorPosX += fx2int(joker_object->sprite_object->x);
+        cursorPosX += fx2int(((SpriteObject*)joker_object)->x);
```
MeirGavish, line 296: This is not consistent with the other instances of accessing this, I just wanted to test the casting for this.

MeirGavish:
See #536 for how this can actually be used in the shop.

MeirGavish:
Closing this in favor of #536 as it now uses the Item struct, this PR should not be merged as it is. The description of the inheritance idea here is useful though.

PR #536, opened 2026-06-05
Refactor shop code to use generic `Item`

### LINKED ISSUE #534: Create a generic object for `SpriteObject`s and refactor the shop to use it
Following the discussions in #481, I'm opening a separate issue for the generic object implementation and then we can move on to #481 and #501 etc.
I think the first step is to refactor the shop to contain the generic object, this will allow for containing consumables in the shop as well and pave the way for Planet Cards and Tarot Cards.
Related PR with possible initial implementation - #533.

MeirGavish:
Closes #534 

The idea of using inheritance in C is described in the PR this is based on #533.

The design details are explained in the linked issue #534, this PR now uses the Item struct described there.

Setting this as ready for review.
This PR aims to define and implement the core functionalities of `Item` that all items must implement in order to appear in the shop.
It changes many files because it added the inheritance system which affects all `JokerObject`s and `CardObject`s but the most important files are `item.c/.h`, `item_funcs.c/.h`, and `shop.c/.h`
The shop is almost entirely refactored to use `Item` instead of `JokerObject`, the only exception is the descriptions.
I will handle descriptions in a separate issue/PR, or maybe allow for unimplemented descriptions for planet cards at first.
Another thing that's not handled in this PR as discussed in #534 is selling functionality, this will be needed for Consumables but will also wait for another PR.


source/hand.c
```diff
@@ -259,8 +259,8 @@ void reorder_card_sprites_layers(void)
             card_object_set_sprite(hand.cards[i], i);
             sprite_position(
                 card_object_get_sprite(hand.cards[i]),
-                fx2int(hand.cards[i]->sprite_object->x),
-                fx2int(hand.cards[i]->sprite_object->y)
+                fx2int(sprite_object_get_x((SpriteObject*)hand.cards[i])),
+                fx2int(sprite_object_get_y((SpriteObject*)hand.cards[i]))
```
MeirGavish: Note: this code will be removed by #539 either way

include/sprite.h
```diff
@@ -223,6 +224,21 @@ void sprite_object_destroy(SpriteObject** sprite_object);
...
+FIXED sprite_object_get_tx(const SpriteObject* sprite_object);
+void sprite_object_set_tx(SpriteObject* sprite_object, FIXED tx);
+FIXED sprite_object_get_ty(const SpriteObject* sprite_object);
+void sprite_object_set_ty(SpriteObject* sprite_object, FIXED ty);
+void sprite_object_set_vx(SpriteObject* sprite_object, FIXED vx);
+void sprite_object_set_vy(SpriteObject* sprite_object, FIXED vy);
+FIXED sprite_object_get_tscale(const SpriteObject* sprite_object);
+void sprite_object_set_tscale(SpriteObject* sprite_object, FIXED tscale);
+FIXED sprite_object_get_trotation(const SpriteObject* sprite_object);
+void sprite_object_set_trotation(SpriteObject* sprite_object, FIXED trotation);
```
MeirGavish: I added setters and getters for SpriteObject's fields before I started using -fms-extensions so there's no need to access so many nested fields to get to them like `joker_object->item.sprite_object.x` and you can just do `sprite_object_get_x((SpriteObject*)joker_object)`, however now I don't need it and you can just do `joker_object->x` which is so much cleaner so I'm just going to remove these functions and go back to doing things like that.

MeirGavish (commented):
Just adding some comments about the implementation, I'm going to remove the getters and setters now.

source/hand.c
MeirGavish: As part of #551 then...

include/item.h
```diff
@@ -0,0 +1,115 @@
...
+{
+    /**
+     * All items must implement the following since they are called by the shop.
+     * For some of them they can be no-op or return true implementations.
+     */
+    int (*get_buy_price)(Item* item);
+    void (*on_acquired)(Item* item);
+    void (*destroy)(Item** item);
+    bool (*can_acquire)(Item* item);
+    void (*set_available_to_shop)(Item* item, bool available);
```
MeirGavish: So while refactoring and trying to get rid of joker-specific code I'm having more design thoughts here. The issue is the bitsets.
I wanted to refactor it to be generic so I created `set_available_to_shop()`. However not all item types will have an availability set - specifically playing cards don't.
So what I want to do is integrate it into different `create()` (or `create_for_shop()`?) and `dispose()` functions that will replace `destroy()` and `set_available_to_shop()` handle the availability set themselves. `create()` will actually roll the rarity. This means moving the bitsets away from the shop.

include/item.h
```diff
@@ -0,0 +1,151 @@
...
+    // ITEM_TYPE_PACK
+
+    ITEM_NUM_TYPES
+};
+
+typedef struct
+{
+    SpriteObject; // First member struct inheritance
+    // All items in the shop and inventory are SpriteObjects.
+    // Casts from Item* (or inheriting structs) to SpriteObject* are allowed and intentional.
```
MeirGavish: Note the pattern of inheritance here, I opted to use the -fms-extensions so this is an anonymous member,
fully leaning into the is-a relationship so you can directly access the inherited struct members, `item->x = 3;` instead of `item->sprite_object->x = 3;`
and there's no has-a relationship at all so you can't do `item->sprite_object` at all.

include/item.h
```diff
@@ -0,0 +1,153 @@
...
+     * should be capable of appearing in the shop.
+     */
+    Item* (*roll_new)(void);
+    int (*get_buy_price)(Item* item);
+    bool (*can_acquire)(Item* item);
+    void (*acquire)(Item* item);
+    void (*dispose)(Item** item);
+    // TODO: void (*print_description)(Item* item); // or something of the form
+
+    // Optional implementation functions will be added here
```
MeirGavish, line 105: Meaning implementations for functions that are not strictly necessary to appear in the shop for selling items, containing items in packs. etc.

include/item.h
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * All items must implement the following since they are called by the shop and all items
+     * should be capable of appearing in the shop.
+     */
+    Item* (*roll_new)(void);
+    int (*get_buy_price)(Item* item);
+    bool (*can_acquire)(Item* item);
+    void (*acquire)(Item* item);
+    void (*dispose)(Item** item);
+    // TODO: void (*print_description)(Item* item); // or something of the form
```
MeirGavish, line 103: I think this will be the final set of functions that all items must implement so they can appear in the shop.

include/sprite.h
```diff
@@ -194,23 +194,24 @@ void sprite_init(void);
...
- * @brief Allocate and retrieve a pointer to a valid SpriteObject
+ * @brief Initialize a SpriteObject to a default state.
+ * Must be called only once per SpriteObject when it is created.
  *
- * @return A valid pointer to an newly allocated SpriteObject
- *         if successful, othewise return **NULL**.
+ * @param sprite_object - The SpriteObject to initialize
  */
-SpriteObject* sprite_object_new();
+void sprite_object_init(SpriteObject* sprite_object);
```
MeirGavish, line 202: `SpriteObject` has been removed from the pool, it's an abstract struct now with no concrete instances, all its instances are inherited structs.
So I changed new to init, it no longer allocates and only initializes the SpriteObject fields and adds it to the `SpriteObject` list.

include/sprite.h
```diff
@@ -194,23 +194,24 @@ void sprite_init(void);
...
+ * Destroy a SpriteObject by releasing its associated resources (e.g. the sprite).
+ * This invalidates the SpriteObject and it should not be used after destroyed,
+ * a new one should be created instead.
  *
- * @param sprite_object pointer to a pointer of SpriteObject to destroy.
+ * @param sprite_object pointer to a SpriteObject to destroy.
  *        Cannot be **NULL**.
  */
-void sprite_object_destroy(SpriteObject** sprite_object);
+void sprite_object_destroy(SpriteObject* sprite_object);
```
MeirGavish, line 214: Destroy still has the same name and still needs to be called to free the `Sprite` and remove the `SpriteObject` from the list

source/item.c
```diff
@@ -0,0 +1,71 @@
...
+
+#include <tonc.h>
+
+Item* item_roll_new(enum ItemType item_type)
+{
+    if (item_type < 0 && item_type >= ITEM_NUM_TYPES)
+    {
+        MGBA_FUNC_ERROR("Undefined type %d", item_type);
+        return NULL;
+    }
```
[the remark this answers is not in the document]
MeirGavish: Right, missed that

source/item.c
```diff
@@ -0,0 +1,71 @@
...
+
+void item_dispose(Item** item)
+{
+    GBAL_RETURN_IF_NULL_VOID(item);
+    GBAL_RETURN_IF_NULL_VOID(*item);
+    ItemFuncs* item_funcs = get_item_type_funcs((*item)->type);
+    GBAL_RETURN_IF_NULL_VOID(item_funcs);
+    GBAL_RETURN_IF_NULL_VOID(item_funcs->dispose);
+
+    return item_funcs->dispose(item);
```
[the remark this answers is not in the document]
MeirGavish: Right too

source/item_funcs.c
```diff
@@ -0,0 +1,67 @@
...
+    ItemFuncs* ret_func = NULL;
+    if (type >= ITEM_NUM_TYPES || type >= NUM_ELEM_IN_ARR(item_func_table))
+    {
+        MGBA_FUNC_ERROR("Invalid type %d", type);
+    }
+    {
+        ret_func = &item_func_table[type];
+    }
+    return ret_func;
+}
```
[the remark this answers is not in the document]
MeirGavish: Yeah, okay

source/card.c
```diff
@@ -178,5 +181,18 @@ Sprite* card_object_get_sprite(CardObject* card_object)
...
     if (card_object == NULL)
         return NULL;
-    return sprite_object_get_sprite(card_object->sprite_object);
+    return sprite_object_get_sprite((SpriteObject*)card_object);
+}
+
+int card_object_get_buy_price(Item* card_object)
+{
+    return 1;
+}
```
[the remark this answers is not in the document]
MeirGavish: Okay, I guess the item type should really be checked

source/joker.c
```diff
@@ -250,18 +262,160 @@ void joker_object_destroy(JokerObject** joker_object)
...
+    GBAL_RETURN_IF_NULL_VOID(joker_object_item);
+    GBAL_RETURN_IF_NULL_VOID(*joker_object_item);
+    ITEM_RETURN_IF_UNEXPECTED_TYPE_VOID(*joker_object_item, ITEM_TYPE_JOKER);
+
+    JokerObject** joker_object = (JokerObject**)joker_object_item;
+    GBAL_RETURN_IF_NULL_VOID((*joker_object)->joker);
+
+    joker_set_rollable((*joker_object)->joker->id, true);
+    joker_object_destroy((JokerObject**)joker_object_item);
+}
```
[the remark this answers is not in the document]
MeirGavish: Hm... this is a problem... My bad, I need to think about this a little more...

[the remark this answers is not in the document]
MeirGavish: Oh well, I'll take his suggestion, it's a little awkward but it does the job...

source/sprite.c
```diff
@@ -169,28 +171,25 @@ int sprite_get_pb(const Sprite* sprite)
...
     sprite_object_reset_transform(sprite_object);
     sprite_object->focused = false;
 
     list_push_back(&sprite_objects_list, sprite_object);
-
-    return sprite_object;
 }
 
-void sprite_object_destroy(SpriteObject** sprite_object)
+void sprite_object_destroy(SpriteObject* sprite_object)
```
ricfehr3, line 185: It's good to see the pointers-of-pointers not being required anymore :)

include/item.h
```diff
@@ -0,0 +1,153 @@
...
+
+    // Future planned item types
+    // ITEM_TYPE_CONSUMABLE, // Expand to PLANET, TAROT, and SPECTRAL?
+    // ITEM_TYPE_VOUCHER,
+    // ITEM_TYPE_PACK
+
+    ITEM_NUM_TYPES
+};
+
+typedef struct
```
ricfehr3: For this particular struct (and any other you think could be relevant), we should have an entry in doxygen (e.g. what we have for `ListNode` in `list.h`). The how the inheritance works should be described there as well. I don't think it needs too many details of _how_ it works, but it would be good to have an explanation that the extension `fms-extensions` makes `Item`'s members directly accessible.

include/item.h
ricfehr3: I think this will be good too. More can be added as well.

include/item.h
```diff
@@ -0,0 +1,153 @@
...
+{
+    SpriteObject; // First member struct inheritance
+    // All items in the shop and inventory are SpriteObjects.
+    // Casts from Item* (or inheriting structs) to SpriteObject* are allowed and intentional.
+
+    enum ItemType type;
+} Item;
+
+typedef struct item_funcs
```
ricfehr3:
```suggestion
typedef struct
```

For consistency

include/sprite.h
ricfehr3: I think it makes sense this way too. Honestly, it probably should have been `sprite_object_free()` before.

include/game.h
```diff
@@ -4,12 +4,13 @@
...
 #include "game_variables.h"
 #include "graphic_utils.h"
+#include "item.h"
 
 #include <tonc.h>
 
 #define MAX_DECK_SIZE        52
 #define MAX_JOKERS_HELD_SIZE 5 // This doesn't account for negatives right now.
-#define MAX_SHOP_JOKERS      2 // TODO: Make this dynamic and allow for other items besides jokers
+#define MAX_SHOP_ITEMS       2 // TODO: Make this dynamic and allow for other items besides jokers
```
ricfehr3:
```suggestion
#define MAX_SHOP_ITEMS       2 // TODO: Make this dynamic
```

Nitpick on the comment. I think this PR covers the last bit.

source/joker.c
```diff
@@ -250,18 +262,162 @@ void joker_object_destroy(JokerObject** joker_object)
...
     POOL_FREE(JokerObject, *joker_object);
     *joker_object = NULL;
 }
 
+void joker_object_dispose(Item** joker_object_item)
+{
+    GBAL_RETURN_IF_NULL_VOID(joker_object_item);
+    GBAL_RETURN_IF_NULL_VOID(*joker_object_item);
+    ITEM_RETURN_IF_UNEXPECTED_TYPE_VOID(*joker_object_item, ITEM_TYPE_JOKER);
+
```
ricfehr3, line 276: I enjoy how consistent and obvious these checks are now. Good work!

source/joker.c
```diff
@@ -250,18 +262,162 @@ void joker_object_destroy(JokerObject** joker_object)
...
+void joker_reset_rollable_jokers(void)
+{
+    int num_jokers = get_joker_registry_size();
+
+    bitset_clear(&s_rollable_jokers_bitset);
+    for (int i = 0; i < num_jokers; i++)
+    {
+        bitset_set_idx(&s_rollable_jokers_bitset, i, true);
+    }
```
ricfehr3, line 344: I think this should be moved to the bitset function itself as `bitset_set_all()` or similar. Then it can just set the raw values to 0xFFFFFFFF and avoid this loop.

source/joker.c
```diff
@@ -250,18 +262,162 @@ void joker_object_destroy(JokerObject** joker_object)
...
+static int joker_roll_id(void)
+{
+    // Roll for what rarity the joker will be
+    int joker_rarity = joker_get_random_rarity();
+
+    // Now determine how many jokers are available based on the rarity
+    int jokers_avail_size = get_num_rollable_jokers();
+
+    if (jokers_avail_size == 0)
+        return UNDEFINED;
```
ricfehr3:
```suggestion
    // Now determine how many jokers are available based on the rarity
    int jokers_avail_size = get_num_rollable_jokers();

    if (jokers_avail_size == 0)
        return UNDEFINED;
        
    // Roll for what rarity the joker will be
    int joker_rarity = joker_get_random_rarity();
```

I think this should be declared with the other variables below since it doesn't depend on the check.

source/joker.c
```diff
@@ -305,12 +461,13 @@ bool joker_object_score(
     {
         // display the text on top of the card instead of below the Joker for Held Cards effects
         // scored_card cannot be NULL here because of the joker event
-        cursorPosX += fx2int(card_object->sprite_object->x);
+        cursorPosX += fx2int(card_object->x);
```
ricfehr3, line 464: This does look really nice. Nice!

source/item_funcs.c
```diff
@@ -0,0 +1,65 @@
...
+        .acquire = item_acquire_unimplemented,
+        .can_acquire = item_always_can_acquire,
+        .dispose = card_object_dispose
+    }
+};
+// clang-format on
+
+ItemFuncs* get_item_type_funcs(enum ItemType type)
+{
+    if ((int)type < 0 || type >= ITEM_NUM_TYPES || type >= NUM_ELEM_IN_ARR(item_func_table))
```
ricfehr3:
```suggestion
    if ((int)type < 0 || type >= NUM_ELEM_IN_ARR(item_func_table))
```

The check is redundant. The NUM_ELEM_IN_ARR check ensures it's < ITEM_NUM_TYPES.

source/game.c
```diff
@@ -419,7 +425,8 @@ void remove_owned_joker(int owned_joker_idx)
         shortcut_joker_count--;
     }
 
-    game_shop_set_joker_avail(joker_object->joker->id, true);
+    // TODO: Move to site of joker_destroy()?
+    joker_set_rollable(joker_object->joker->id, true);
```
ricfehr3, line 439: I think it makes sense to have this in either place really. But let's keep the TODO in.

source/game/run_setup.c
```diff
@@ -632,7 +632,13 @@ void game_run_setup_on_exit(void)
 static void choose_deck_substate_init(void)
 {
     // Show Deck sprite, name and TODO: description
-    obj_unhide(run_setup_deck->sprite_object->sprite->obj, ATTR0_AFF);
+    {
```
ricfehr3, line 653: Out of curiosity, why wrap this and other operations in this file with scope-guards?

ricfehr3 (changes requested):
Nice work @MeirGavish , I'm pretty sold on the design after (_finally_) getting to this PR. The anonymous members are a pretty good solution for the consumables. The transparent access of the sprite object and usage of the function table glued very well into the existing code. I was initially concerned there would be a learning curve to understanding what was going on, but it's a lot clearer than I was making it out to be. Nice! I'll be back to review more (specifically a better look at `shop.c`), but I have comments already. 

Great refactor! I think this is exemplary of the design books for the SOLID principles you were studying. It was enjoyable to see how flexible the existing code was to the change as well. It's looking really good.

include/sprite.h
MeirGavish: I can rename it to `free()` but then I wonder if other destroy functions need to be renamed to free too for consistency or just this one because it behaves a little differently...

source/game/run_setup.c
MeirGavish: Oh, it's a silly little thing, it's because the `deck_sprite` variable is only needed here and for the NULL check... This is made redundant by #563.

include/game.h
MeirGavish: Right! Good catch!

source/sprite.c
MeirGavish: It's just because it doesn't nullify the struct anymore, they are still required for other destroy functions

source/joker.c
MeirGavish: Thanks! I thought of merging the `ITEM_RETURN_IF_UNEXPECTED_ITEM_TYPE` and `GBAL_RETURN_IF_NULL` macros, I opened issue #585 for this.

source/game.c
```diff
@@ -272,11 +272,17 @@ static inline void held_jokers_update_loop(void)
     {
         // Let the Shop handle the position of this Joker
         if (joker != game_shop_get_description_card())
-            joker->sprite_object->tx = hand_x - int2fx(spacing_lut[jokers_top][i]);
+            joker->tx = hand_x - int2fx(spacing_lut[jokers_top][i]);
         i++;
     }
 }
 
+bool joker_object_can_acquire(Item* joker_object)
```
MathisMartin31, line 290: I see the other `can_acquire` functions are specific to the kind of Item, but not this one which returns whether we can acquire any Joker. Is this intended/taken into account?

include/joker.h
```diff
@@ -112,8 +112,8 @@ typedef struct
 
 typedef struct JokerObject
 {
+    Item; // First member struct inheritance
     Joker* joker;
-    SpriteObject* sprite_object;
 } JokerObject;
```
MathisMartin31, line 117: I still can't get used to the `Item` struct being owned by the `FooObject` and not the other way around ha ha, but I do see the appeal of the inheritence/casting in what you did so if it works let's roll with it :)

MeirGavish:
> Nice work @MeirGavish , I'm pretty sold on the design after (_finally_) getting to this PR. The anonymous members are a pretty good solution for the consumables. The transparent access of the sprite object and usage of the function table glued very well into the existing code. I was initially concerned there would be a learning curve to understanding what was going on, but it's a lot clearer than I was making it out to be. Nice! I'll be back to review more (specifically a better look at `shop.c`), but I have comments already.
> 
> Great refactor! I think this is exemplary of the design books for the SOLID principles you were studying. It was enjoyable to see how flexible the existing code was to the change as well. It's looking really good.

Thanks! Yeah, after long arduous design discussions, the final code is pretty nice and clear which is great because that's how code should be.
I will try to address the review soon.

source/item_funcs.c
MeirGavish: Yeah, you're right

source/joker.c
```diff
@@ -250,18 +262,162 @@ void joker_object_destroy(JokerObject** joker_object)
             UNDEFINED;
     }
 
-    sprite_object_destroy(&(*joker_object)->sprite_object); // Destroy the sprite
-    joker_destroy(&(*joker_object)->joker);                 // Destroy the joker
+    sprite_object_destroy((SpriteObject*)(*joker_object)); // Destroy the sprite
+    joker_destroy(&(*joker_object)->joker);                // Destroy the joker
```
```suggestion
    sprite_object_destroy((SpriteObject*)(*joker_object));
    joker_destroy(&(*joker_object)->joker);
```
Let's remove these redundant comments while we're at it.

source/game.c
MeirGavish: I think my thoughts here were about how this could be generalized for any item that has a rollable set.

source/joker.c
MeirGavish: I think I want to do that in a separate PR, adding something to the bitset calls for unit tests etc.
Edit: Opened an issue and draft PR #587

MeirGavish:
Just noting that I think I might have a bug in joker bitset management in this PR, I will look into it.

MathisMartin31 (commented):
Quick review from me, just one question about the can acquire JokerObject. Nice work with this ^^

include/joker.h
MeirGavish: Think about it as inheritance, 
```c
struct JokerObject
{
   Item;
  ...
};
```
is like
```c
struct JokerObject : public Item
{
  ...
};
```

source/game.c
MeirGavish: Yeah, the item is passed for consistency with the rest of the item functions, but in practice it will probably always check the inventory.
Price checks are done by the shop.
This function is intended to be usable by packs as well.

MeirGavish:
> Just noting that I think I might have a bug in joker bitset management in this PR, I will look into it.

I'm not seeing it now... it may have been due to some mistake when trying to address [this comment](https://github.com/GBALATRO/balatro-gba/pull/536#discussion_r3533149529) in this PR.

include/item.h
MeirGavish: I added Doxygen documentation for the `Item` struct, I also had a small bit in the top of the file, I guess it's okay to leave it in there.
It's a pretty long explanation and documentation duplication is not so good so I will keep it in `Item` only, I suppose anybody who's reading the code for an inheriting struct will reach this code.

MeirGavish:
@ricfehr3 I think I addressed everything, it should be good for another review when you get around to it.

source/joker.c
MeirGavish: So apparently there's an issue with doing this, see https://github.com/GBALATRO/balatro-gba/pull/587#issuecomment-4911347868

source/game.c
MathisMartin31: That makes sense, thanks ^^

include/joker.h
MathisMartin31: Oh yeah, I had understood how the casting worked but writing it like this makes it soooo clear ^^. I think I just did way too much C++ since I got a job 😭

source/joker.c
```diff
@@ -250,18 +262,162 @@ void joker_object_destroy(JokerObject** joker_object)
...
+    if (joker_id == UNDEFINED)
+        return NULL;
+
+    joker_set_rollable(joker_id, false);
+
+    return (Item*)joker_object_new(joker_new(joker_id));
 }
 
-void set_and_shift_text(char* str, int* cursor_pos_x, int* cursor_pos_y, int color_pb)
+static void set_and_shift_text(char* str, int* cursor_pos_x, int* cursor_pos_y, int color_pb)
```
ricfehr3, line 420: I don't think this belongs here. This functionality should be shared in `layout.h`. I don't think it should be fixed in this PR, but we should make an issue to remember it.

source/joker.c
ricfehr3: Okay, yeah. That is outside of this PR 100% then. Thanks for looking into it! This will be fine for this PR.

ricfehr3 (approved):
Awesome! Gave this another review and I think it's good now. Really nice work! I had one comment on moving a function to `layout.h`, but ultimately it's just a style decision, so whatever you choose is fine. Please feel free to merge then whenever you would like. At your discretion I'll let you choose if you want to do this before or after #541

source/joker.c
```diff
@@ -305,12 +461,13 @@ bool joker_object_score(
...
         // scored_card cannot be NULL here because of the joker event
-        cursorPosX += fx2int(card_object->sprite_object->x);
+        cursorPosX += fx2int(card_object->x);
         cursorPosY = HELD_CARD_SCORE_TEXT_Y;
     }
     else
     {
-        cursorPosX += fx2int(joker_object->sprite_object->x);
+        // Adding direct access instead of cast and/or call to getter (same in card_object above)
+        cursorPosX += fx2int(joker_object->x);
```
```suggestion
        cursorPosX += fx2int(joker_object->x);
```

The comment here should be removed... It shouldn't be in the code, it's describing this PR specifically

source/joker.c
MeirGavish: This function is a little weird. I thought it was very specific to joker scoring, and indeed the last 2 lines are very scoring-specific. I guess this function should be changed a bit to be more clear, at the very least renamed and documented.
Then I guess it can be used for card scoring as well and be shared, I guess common_ui.h would be most appropriate, _since layout.h only contains rect definitions_.
Though maybe the function should be changed more than that... it does 2 things and it's a little difficult to understand what these are despite it's brevity.
Set what text? What sort of shift? It actually displays the string on the screen, that's not implied from the name. cursor_pos_x is both an input and an output.
Edit: opened issue #592

MeirGavish:
> Awesome! Gave this another review and I think it's good now. Really nice work! I had one comment on moving a function to `layout.h`, but ultimately it's just a style decision, so whatever you choose is fine. Please feel free to merge then whenever you would like. At your discretion I'll let you choose if you want to do this before or after #541

Thanks! I will merge this first then.

PR #530, opened 2026-06-04
[Feature] Implement Joker descriptions in the Shop when holding B

MathisMartin31:
Closes #219

As I said over in the issue, I have yet to implement the glitched description for the Misprint joker, but it's such a small part of the PR that I'll put this up as ready for review so you guys can get a headstart and take a look at it.

https://github.com/user-attachments/assets/c6ad4a62-d501-4f71-9559-5293a32bf3af

PR #538, opened 2026-06-05
Removed dubious SpriteObject cast from long ago

MeirGavish:
So while working on #533 and #536 I realized there was a cast I meant to fix and missed back when I introduced `SpriteObject` in #63 and actually this is one of the very reasons I introduced it. 🤣
So the cast has been working so far, but just because `CardObject` works, `JokerObject` should in no way depend on `CardObject`.

### DIFF
--- a/source/joker.c
+++ b/source/joker.c
@@ -231,8 +231,7 @@ void joker_object_destroy(JokerObject** joker_object)
 
 void joker_object_update(JokerObject* joker_object)
 {
-    CardObject* card_object = (CardObject*)joker_object;
-    card_object_update(card_object);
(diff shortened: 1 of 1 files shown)


ricfehr3:
lol, nice! That's such a good catch. It's been hiding, waiting for a nightmare when `CardObject` would eventually change :joy: . lgtm. ~~Merging~~ _merging right after #530  ._

ricfehr3 (approved):

MathisMartin31 (approved):

ricfehr3:
Going to merge this one now, as promised.

PR #542, opened 2026-06-07
Add button flash in run setup menu

### LINKED ISSUE #540: When starting a new run "New Run" button is highlighted despite not being an active button
In fact, it doesn't feel like the deck section is what's highlighted, it feels like the "New Run" button is highlighted and pressing it will work even though it doesn't do anything...

Image

ricfehr3:
Closes #540 

Quick implementation of an idea to add flashing buttons for the run setup menu.

It's not very slick looking, but it is a lot more obvious where the cursor is. We don't need to add this, but it's here as at least an example

https://github.com/user-attachments/assets/561e290d-e02f-46df-9853-fdd01065e3a8

### DIFF
--- a/source/game/run_setup.c
+++ b/source/game/run_setup.c
@@ -114,7 +114,9 @@ static Button* active_button = NULL;
 
 static int flash_select_button_timer = 0;
 static bool flash_select_button = false;
-static StateInfo select_flash_state_info[] = {STATE_INFO_INIT_UPDATE_FN(flash_select_button_init, flash_select_button_update)};
+static StateInfo select_flash_state_info[] = {
+    STATE_INFO_INIT_UPDATE_FN(flash_select_button_init, flash_select_button_update)
+};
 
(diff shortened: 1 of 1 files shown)


MeirGavish:
I think the better thing to do is just remove the highlight from the "New Run" button.

MeirGavish:
Or, make the "New Run" button flash and the other highlights behave the same as the other button highlights in the game...

MathisMartin31:
I agree with Meir here, I can't say I'm a big fan of the blinking buttons, it's a bit agressive compared to the general chill vide of the game.

Tabs blinking, and slower, would be better imo.

Also the Deck "button" will be more visible in the future since there will be arrows that will show up on either side to signal that we can cycle through the different Decks

ricfehr3:
Fair points. It does look bad, I agree, haha. I'll close this, but I'll reference it in the issue so we can consider if we would like to have flashing buttons at all.

PR #546, opened 2026-06-10
Temporarily remove "new run" and "resume" buttons

### LINKED ISSUE #540: When starting a new run "New Run" button is highlighted despite not being an active button

ricfehr3:
@MeirGavish Yes, I did push to main with this commit on accident. No idea how. I think I setup the branch tracking in some cursed way. I reverted it in main and made this pr.

Anyway! Here is the quick fix for #540 . We can simply revert this commit and add it back to the status it was. 

If this one is approved I'll merge #539. Then a play test day and I think it's time.


MathisMartin31 (approved):

MeirGavish:
> @MeirGavish Yes, I did push to main with this commit on accident. No idea how. I think I setup the branch tracking in some cursed way. I reverted it in main and made this pr.

It seems like GitHub allows administrators to push to protected branches, I checked "Do not allow bypassing the above settings" to prevent this. I also added a requirement that PRs can't be merged unless workflow checks have passed, for not I just put the "Build ROM" and "Run Unit Tests" there just to see how it works, but I guess the clang-format should also be added there.

MeirGavish (approved):

MeirGavish:
> I also added a requirement that PRs can't be merged unless workflow checks have passed, for not I just put the "Build ROM" and "Run Unit Tests" there just to see how it works, but I guess the clang-format should also be added there.

@ricfehr3 
I removed this, I couldn't get it to register the actual workflows run, and it waited for a report even after they passed, it could be nice to get it to work sometime.

ricfehr3:
@MeirGavish Sounds good, thanks for checking into those settings. I'll also look into setting that up properly. I'll also just be more aware myself going forward.

PR #548, opened 2026-06-11
[Enhancement] Game Over screen rework

### LINKED ISSUE #527: [Feature] Display the seed on game over
As of #482, we allow seeded runs by entering a seed when starting a game but we don't display the seed used in a randomly chosen seed run so the player can only use seeds that he guesses.
We need to display the seed on game over so players can keep the seed from their run and use it.

MathisMartin31:
Closes #527

I did go a bit further than strictly showing the seed on the Game Over menus but I thought it made sense to implement the rest of the stuff shown on there. The info displayed is a bit limited but it's something ^^'

You also get the option to reuse the current seed for the next run, but it still takes you to the Run Setup screen so you can choose your Deck (not implemented yet) or choose your Seed if you change your mind.

https://github.com/user-attachments/assets/175050ea-7083-48a8-8a10-d6eaa739cdb2


include/card.h
```diff
@@ -7,7 +7,7 @@
 #include <maxmod.h>
 #include <tonc.h>
 
-#define MAX_CARDS           (NUM_SUITS * NUM_RANKS)
+#define MAX_CARDS           128 // Arbitrary high value so we don't have issues
```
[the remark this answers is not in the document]
MeirGavish: Damn it, Copilot!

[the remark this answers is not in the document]
MathisMartin31: Ha ha, well at least we know it doesn't waste tokens reading all of our conversations xD (although you would think that it would at least synthesize what was said in the linked issue and add it to its context 🤔)

source/game/run_setup.c
```diff
@@ -1162,6 +1180,8 @@ static void play_on_pressed(void)
     else
         rng_shuffle_seed();
```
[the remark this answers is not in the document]
MathisMartin31: Hmm, even if the seed is empty/0, I don't think we should use a random one. Once the box has been ticked I think we should use the current fixed seed no matter what.

MeirGavish:
I do want to note that adding this to the release could be a little risky because of the return to main menu feature, it needs to be tested to make sure nothing breaks there.

MathisMartin31:
> I do want to note that adding this to the release could be a little risky because of the return to main menu feature, it needs to be tested to make sure nothing breaks there.

Yes I agree, I think this can wait until the release is out

MathisMartin31:
Same thing as the skip tags PR, I'll test this later tonight and make sure it's ready to go

MeirGavish:
> Same thing as the skip tags PR, I'll test this later tonight and make sure it's ready to go

Cool, I don't know when I'll get to it but I want to review this next.

MathisMartin31:
> Cool, I don't know when I'll get to it but I want to review this next.

It's good for review, but no rush ^^

include/game.h
```diff
@@ -76,6 +76,7 @@ List* get_jokers_list(void);
 List* get_expired_jokers_list(void);
 List* get_discarded_jokers_list(void);
 
+Card* deck_pop(void);
```
MeirGavish: It seems like one of the rebases added a second declaration for `deck_pop()`, remove it so there's only one.

source/game/game_over.c
```diff
@@ -21,52 +34,148 @@ enum EndCondition
...
+        g_game_vars.playing_blind_token,
+        DEFEATED_BY_TOKEN_INIT_POS.x,
+        DEFEATED_BY_TOKEN_INIT_POS.y
+    );
+
+    // Highlight New Run button
+    game_over_selection_grid.selection = GAME_OVER_INIT_SEL;
+    button_set_highlight(&game_over_buttons[GAME_OVER_SEED_ROW][0], false);     // Seed
+    button_set_highlight(&game_over_buttons[GAME_OVER_RUN_MENU_ROW][0], true);  // New Run
+    button_set_highlight(&game_over_buttons[GAME_OVER_RUN_MENU_ROW][1], false); // Main Menu
```
MeirGavish: These are magic numbers, remove the comments and define `GAME_OVER_NEW_RUN_BTN_COL` etc. or similar so the comments are not needed and the code is self-descriptive and maintainable.

include/game_variables.h
```diff
@@ -28,6 +28,10 @@
 
 #define DEFAULT_HAND_SIZE 8
 
+// There are 13 different hand types in total.
+// Should have been in hand.h but we'll avoid the extra dependency
+#define HAND_TYPE_MAX 13
```
MeirGavish: Is it problematic to include hand.h here? Does it break compilation? Because if not I think it would be better to include it here rather than have duplicated defines.

source/game/game_over.c
```diff
@@ -76,67 +185,168 @@ void game_lose_on_init(void)
...
+    const Selection* new_selection
+)
+{
+    // Remove highlight from previous button
+    if (row_idx == prev_selection->y)
+        button_set_highlight(game_over_get_button_from_sel(prev_selection), false);
+
+    // Highlight new button
+    if (row_idx == new_selection->y)
+        button_set_highlight(game_over_get_button_from_sel(new_selection), true);
```
MeirGavish: These comments are redundant and should be removed, the code is self-descriptive enough

source/game/game_over.c
```diff
@@ -76,67 +185,168 @@ void game_lose_on_init(void)
...
+            g_game_vars.playing_blind_token->pos.y - TILE_SIZE
+        );
     }
+
+    // Print values and buttons text
     else if (timer == GAME_OVER_ANIM_FRAMES)
     {
-        switch (condition)
+        char best_hand_str[UINT_MAX_DIGITS + 1];
+        truncate_uint_to_suffixed_str(g_game_vars.best_played_hand, 5, best_hand_str);
```
MeirGavish: 5 is a magic number here, why not use `rect_width(&BEST_HAND_VALUE_RECT)/TTE_CHAR_SIZE` like the other calls instead?

source/game.c
```diff
@@ -95,11 +96,15 @@ static StateMachine game_sm = STATE_MACHINE_DEFINE(state_info, GAME_STATE_MAX);
...
 GameVariables g_game_vars = {
-    .timer = 0, .rng_info = {0, 0},
+    .timer = 0,
+    // Setting the seed to an invalid value so that the Run Setup screen knows we're not reusing a previous Run's seed
+    .rng_info = {UNDEFINED, 0},
 
     .round = 0, .ante = 0, .money = 0, .hand_size = DEFAULT_HAND_SIZE,
     .deck = DECK_TYPE_RED,
 
+    .best_played_hand = 0, .nb_played_hands = {0},
```
MeirGavish: The variable name `best_played_hand` is confusing as it implies tracking the hand itself rather than the score, it should be renamed `best_hand_score` which better conveys the intent of the variable.

source/game/round.c
```diff
@@ -1069,6 +1072,10 @@ static inline void game_round_process_input_and_state(void)
             lerped_temp_score = int2fx(temp_score);
             lerped_score = int2fx(g_game_vars.score);
 
+            // Register high score
+            if (temp_score > g_game_vars.best_played_hand)
+                g_game_vars.best_played_hand = temp_score;
```
MeirGavish: This comment is redundant, I think the code is self-descriptive enough and it should be removed. This also applies for the not-in-this-PR "protect against score overflow" above, the function is already documented to do that.

MeirGavish (changes requested):
Looks great! Good job! This pretty much completes the seed system.
I have a few comments so I'm requesting changes, nothing big, just a few small comments about names,  magic numbers, and documentation etc.

However, I did see 2 bugs when testing this,
1. After restarting the run multiple times with the same seed it crashed... I can't reproduce this now unfortunately... If you can't reproduce it either, don't spend too much time trying, I think we can merge this PR and open a separate issue to continue investigating it.
2. This is actually the first time I'm testing the seed system, it seems like while using the same seed on separate runs does draw the same cards, the jokers in the shop are different. I don't know if it's like this in main too? If it is, this PR can be merged since it's unrelated and a separate issue can be opened.

include/game_variables.h
MathisMartin31: Yeah I think I just completely forgot about hand.h for a hot second there, my bad ><

MathisMartin31:
@MeirGavish Thanks for the review ^^

I had also tried doing a few runs reusing the same seed and hadn't encountered any crashes, I'll try again tonight because that's really bad if it's related ><

I will also look into the Jokers stuff, ~maybe it's not using the RNG system we made yet?~ It does so that's not the issue here

MeirGavish:
> I had also tried doing a few runs reusing the same seed and hadn't encountered any crashes, I'll try again tonight because that's really bad if it's related ><

We've seen before that touching IWRAM and pools can cause crashes, so maybe it's because of the `MAX_CARDS` increase, maybe bringing it back down to 64 will help. But we still need to reliably reproduce to check.

MathisMartin31:
@MeirGavish I did around 10 runs back to back using the same seed and I didn't crash, and I rolled the same Jokers on the first shop every time so I don't really know what happened on your end there 🤔

MeirGavish:
> @MeirGavish I did around 10 runs back to back using the same seed and I didn't crash, and I rolled the same Jokers on the first shop every time so I don't really know what happened on your end there 🤔 

Huh, interesting, I'll give it another try, thanks!

MathisMartin31:
> > @MeirGavish I did around 10 runs back to back using the same seed and I didn't crash, and I rolled the same Jokers on the first shop every time so I don't really know what happened on your end there 🤔
> 
> Huh, interesting, I'll give it another try, thanks!

Well nevermind, I did two more runs and I did get different Jokers with the same seed as before... It seems to happen to me more when I reset mgba between runs. I always got the same Jokers when starting a new run after loosing the last one

MathisMartin31:
Huh, looking at the code, it seems `rng_get_u32` is called for things like getting a slightly different pitch when drawing and swapping cards, which means depending on how you select cards during the round, you will get different rng in the following shop. See round.c:592/612

I think that, for this kind of fluff, we should use lookup tables big enough so that we don't notice the repetition, and reserve the use of the rng for actual gameplay.

Edit: I replaced these calls to `rng_get_u32` and the Jokers do seem to be same each time now

MeirGavish:
> Huh, looking at the code, it seems `rng_get_u32` is called for things like getting a slightly different pitch when drawing and swapping cards, which means depending on how you select cards during the round, you will get different rng in the following shop. See round.c:592/612
> 
> I think that, for this kind of fluff, we should use lookup tables big enough so that we don't notice the repetition, and reserve the use of the rng for actual gameplay.
> 
> Edit: I replaced these calls to `rng_get_u32` and the Jokers do seem to be same each time now

All of this is definitely for another issue/PR, but recently there has been a very good video giving an overview of the Balatro code on YouTube by Howdy. Here is a [link to the timestamp talking about the RNG](https://youtu.be/54w9crNNThU&t=739). 
Basically it uses what he calls a "per key seeded RNG" where there's a separate sequence of random numbers generated for each key so different types of random events don't affect each other. You can take a look at the Balatro code, I think it's the `pseudoseed()` function that does that.
I don't know how feasible it would be for us to do but that would be the solution that would work best.
Otherwise rerolling the shop would still affect the deck shuffle for the next round for example.

MathisMartin31:
> Basically it uses what he calls a "per key seeded RNG" where there's a separate sequence of random numbers generated for each key so different types of random events don't affect each other. You can take a look at the Balatro code, I think it's the pseudoseed() function that does that.
> I don't know how feasible it would be for us to do but that would be the solution that would work best.
Otherwise rerolling the shop would still affect the deck shuffle for the next round for example.

That's interesting, the only way I can see to do that would be storing multiple RngInfo structs and restoring the rng sequence every time we need to roll from a different one.

It would be rather easy to do it, but the big problem with that is that the longer the run goes on, the more cycles it would take to restore the position in the rng sequence since the only way to do that is to get random numbers until we're at the same point we were (c.f. `load_game` function)

MeirGavish:
> > Basically it uses what he calls a "per key seeded RNG" where there's a separate sequence of random numbers generated for each key so different types of random events don't affect each other. You can take a look at the Balatro code, I think it's the pseudoseed() function that does that.
> > I don't know how feasible it would be for us to do but that would be the solution that would work best.
> > Otherwise rerolling the shop would still affect the deck shuffle for the next round for example.
> 
> That's interesting, the only way I can see to do that would be storing multiple RngInfo structs and restoring the rng sequence every time we need to roll from a different one.
> 
> It would be rather easy to do it, but the big problem with that is that the longer the run goes on, the more cycles it would take to restore the position in the rng sequence since the only way to do that is to get random numbers until we're at the same point we were (c.f. `load_game` function)

For now I think it's okay to leave a naive implementation and explore a better one in another issue.

MathisMartin31:
@MeirGavish Good for another round of review ^^

MathisMartin31:
Also, since we won't be needing the 3x3 rect used for the game over screen in the round background anymore, I have a plan to use the extra free space and add Jokers descriptions to the round he he he :3

MeirGavish (approved):
Great! Approving!

ricfehr3 (approved):
Finally got around to this. Looks good! It looks good and well refined by the point I got to it. I played around with it and it looks good! The random numbers thing is interesting, but obviously for another issue at this point. I was able to consistently get the same hand with the same seed. Approving. I'll merge after a day, but I see @MeirGavish Has already approved though. 💯

MeirGavish:
This has been fully approved so I'm merging

PR #550, opened 2026-06-11
[Bug] Fix missing Deck size/max text after showing Joker description in the Shop

MathisMartin31:
I noticed it earlier, but when showing a Joker's description I clean the right part of the screen from top to bottom, but after hiding the description I forgot to display that text again. This should now be done ^^

https://github.com/user-attachments/assets/b0748dd3-83b4-4289-b7b7-ba68ed68ca64

I'm adding this to the v0.2.0 milestone since this is a bug related to one of the main features of that release.

### DIFF
--- a/include/game.h
+++ b/include/game.h
@@ -96,6 +96,7 @@ int get_deck_top(void);
 int get_num_discards_remaining(void);
 int get_num_hands_remaining(void);
 
+void display_deck_size_max(void);
 u32 get_chips(void);
 void set_chips(u32 new_chips);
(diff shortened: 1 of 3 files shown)


ricfehr3 (approved):
Nice catch! LGTM + I tested `main` and this branch and it fixes the issue. Good job.

MeirGavish (approved):
I haven't tested it but looks good to me, good catch!

ricfehr3:
Merging this, thank you!

PR #552, opened 2026-06-12
Refine controls section in README.md

ricfehr3:
Pretty simple, I was adding the joker description button and realized we should probably clean up the controls description a bit.

I tried to simplify it as much as possible. I removed that the d-pad is for navigation because I feel that is intuitive enough.

_adding to milestone as well_

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ https://github.com/user-attachments/assets/54a9e2e9-1a02-48d5-bb9d-5ab257a7e03b
 
 (B: Deselect All Cards)
 
-(L: Play Hand / Sell Joker)
+(L: Sell Joker  / Play Hand)
 
 (R: Discard Hand)
 
--- a/README.md
+++ b/README.md
@@ -39,21 +39,17 @@ https://github.com/user-attachments/assets/54a9e2e9-1a02-48d5-bb9d-5ab257a7e03b
 
 ### Controls:
 
-(A: Pick Card/Make Selection)
-
-(L: Play Hand)
-
-(R: Discard Hand)
+(A: Pick Card / Make Selection)
 
 (B: Deselect All Cards)
 
-(Hold A + D-Pad: Move Owned Jokers / Cards In Hand)
+(L: Play Hand / Sell Joker)
 
-#### Shop Only Controls
+(R: Discard Hand)
 
-(L: Sell Joker)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,22 +37,23 @@ See **_CONTRIBUTING.md_** for build instructions for Docker, Linux, MacOS, and W
...
 (R: Discard Hand)
 
 (B: Deselect All Cards)
 
-#### When on the joker row in the shop or during round
+(Hold A + D-Pad: Move Owned Jokers / Cards In Hand)
+
+#### Shop Only Controls
+
 (L: Sell Joker)
```
MeirGavish: These are good changes, the only thing is that it's not really true that selling a joker is shop only, you can sell it during the round as well...

MeirGavish (changes requested):

README.md
ricfehr3: Oops. I just simplified / reworded it again. I think it should now make sense.

ricfehr3: At some point, maybe we should add a "controls" screen to the game. Not to change anything, just display a single screen with the controls on it.

MeirGavish: Yeah, that's a good idea, we just need to think of where to display it...

README.md
```diff
@@ -37,22 +37,19 @@ See **_CONTRIBUTING.md_** for build instructions for Docker, Linux, MacOS, and W
...
 
-(A: Pick Card/Make Selections)
+(A: Pick Card / Make Selection)
 
-#### When on the hand row during round
-(L: Play Hand)
+(B: Deselect All Cards)
 
-(R: Discard Hand)
+(L: Play Hand / Sell Joker)
```
MeirGavish:
```suggestion
(L: Sell Joker  / Play Hand)
```

Maybe switch this? 
I just want to put the Sell Joker option first so it's clearer to the player that they can sell a joker and how, while the play hand hotkey is secondary and not so bad if missed.

MeirGavish (commented):
Just uncertain about one thing, but otherwise I think this should be clear an concise enough.

README.md
ricfehr3: That sounds good to me too.

MeirGavish (approved):
Cool, as said, seems concise and clear enough for me, approving!

ricfehr3:
Thanks for the review @MeirGavish !

Issue #509, opened 2026-05-25
Restructure List to have Backends (linked list & array)


ricfehr3:
This is a post-release issue. I also would like to note this and keep a discussion going, as we are going to be using `List` for multiple items: e.g. #498 

Right now, our "`List`" is a doubly-linked list. The term `List` is rather generic, but I think we should take advantage of that and add a bit of abstraction to its design to let it be an easy to swap between "backends."

Since we have a good interface in `list.h` right now, including iterators, push/pop #506 , and arbitrary access calls. I think we should use those interface functions and wrap the list a backend set of functions, either being the doubly linked-list, or a static array.

Declaration would be the only difference, otherwise, all types and functions will be the same, but with the flexibility of choosing the implementation at compile time. 

The reason to centralize this functionality in this API is because, 1: If you're using a `List` instead of just a static array, you've already given up efficiency for some sort of convenience, justifying the abstraction a bit. 2: You can swap backends at compile time. The second is the most important for me, as there is nothing stopping us from having a lot of arbitrary access to a linked list backend, or random insertions and deletions in an array backend. However, if we identify we could benefit from either type, it can be easily swapped. 

This will also let us take on a bit of technical debt when refactoring as we can take this design into account and confidently carry some unoptimized linked lists that will be easily optimized later.

PR #553, opened 2026-06-12
Implement a global SpriteObject update loop

### LINKED ISSUE #535: Call `SpriteObjectUpdate` for all `SpriteObject`s once per frame
Keep a list of all `SpriteObjects` and call `SpriteObjectUpdate()` on all of them once every frame. This will allow removing `JokerObjectUpdate` and `CardObjectUpdate`.
#534 Could help with that but I'm not sure it's strictly necessary.
This is also somewhat related to #501 since it will at least remove the calls to `JokerObjectUpdate` from there.

### LINKED ISSUE #568: [BUG] Joker Sprite Flicker In Top Left Corner On Reroll
**Describe the bug**
When rerolling a joker sprite flickers in the top left corner of the screen for a single frame.

**To Reproduce**
Go to the shop and reroll, it doesn't seem to be affected by how many jokers you already own.
This doesn't seem to always reproduce, not sure what affects it.

**Expected behavior**
The new joker sprites need to be placed in the shop slots when rerolling without flickering elsewhere.

**Screenshots**

https://github.com/user-attachments/assets/9d14042c-e7ab-43a2-a9f6-339f772d67e8

Frame by frame

Image

Image

Image

**What version**
Bug reproduced on the following,
- Main branch with latest commit [Fixed argument on calls to obj_unhide (#562)](https://github.com/GBALATRO/balatro-gba/commit/d2cdc7af1be3bd4ec44ef6049ed29238d077b446)
Also tested and reproduced 2 commits back.
- QOL + Feature Release v0.2.0

Bug reproduced on the following,
- mGBA
- GBA hardware with EZFLASH cart

MathisMartin31:
Closes #535

I did keep some updates to certain sprites, these are done to avoid flickering by showing a newly created sprite at position (0,0) for a single frame before it gets updated.

Here is the game running without any visible issue:

https://github.com/user-attachments/assets/01efef09-a1a0-40f6-b637-8cdab522ad97

Edit: I did notice a slight flicker of the cards when playing them, I moved some updates around and it turns out I don't need the few calls to `sprite_object_update` I kept after all ^^

### DIFF
--- a/include/pool.h
+++ b/include/pool.h
@@ -62,9 +62,9 @@
 
 #define POOL_GET(type)           pool_get_##type()
 #define POOL_FREE(type, obj)     pool_free_##type(obj)
-#define POOL_IDX(type, obj)      pool_idx_##type(obj) // the index of the object
-#define POOL_AT(type, idx)       pool_at_##type(idx)  // the object at
-#define POOL_VALID_AT(type, idx) pool_valid_at_##type(idx)  // the object at idx is currently valid
+#define POOL_IDX(type, obj)      pool_idx_##type(obj)      // the index of the object
+#define POOL_AT(type, idx)       pool_at_##type(idx)       // the object at
+#define POOL_VALID_AT(type, idx) pool_valid_at_##type(idx) // the object at idx is currently valid
 
 #define POOL_ENTRY(name, capacity) POOL_DECLARE_TYPE(name);
 #include POOLS_DEF_FILE
--- a/include/pool.h
+++ b/include/pool.h
@@ -21,8 +21,7 @@
     type* pool_get_##type();          \
     void pool_free_##type(type* obj); \
     int pool_idx_##type(type* obj);   \
-    type* pool_at_##type(int idx);    \
-    bool pool_valid_at_##type(int idx);
+    type* pool_at_##type(int idx);
 
 #define POOL_DEFINE_TYPE(type, capacity)                                \
     BITSET_DEFINE(type##_bitset, capacity)                              \
--- a/include/pool.h
+++ b/include/pool.h
@@ -21,7 +21,8 @@
     type* pool_get_##type();          \
     void pool_free_##type(type* obj); \
     int pool_idx_##type(type* obj);   \
-    type* pool_at_##type(int idx);
+    type* pool_at_##type(int idx);    \
+    bool pool_valid_at_##type(int idx);
 
 #define POOL_DEFINE_TYPE(type, capacity)                                \
     BITSET_DEFINE(type##_bitset, capacity)                              \
@@ -53,12 +54,17 @@
         if (idx < 0 || idx >= (type##_pool.bitset)->cap)                \
             return NULL;                                                \
         return &type##_pool.objects[idx];                               \
+    }                                                                   \
+    bool pool_valid_at_##type(int idx)                                  \
+    {                                                                   \
+        return bitset_get_idx(type##_pool.bitset, idx);                 \
     }
 
-#define POOL_GET(type)       pool_get_##type()
-#define POOL_FREE(type, obj) pool_free_##type(obj)
-#define POOL_IDX(type, obj)  pool_idx_##type(obj) // the index of the object
-#define POOL_AT(type, idx)   pool_at_##type(idx)  // the object at
+#define POOL_GET(type)           pool_get_##type()
+#define POOL_FREE(type, obj)     pool_free_##type(obj)
+#define POOL_IDX(type, obj)      pool_idx_##type(obj) // the index of the object
+#define POOL_AT(type, idx)       pool_at_##type(idx)  // the object at
+#define POOL_VALID_AT(type, idx) pool_valid_at_##type(idx)  // the object at idx is currently valid
 
 #define POOL_ENTRY(name, capacity) POOL_DECLARE_TYPE(name);
 #include POOLS_DEF_FILE
--- a/include/pool.h
+++ b/include/pool.h
@@ -54,17 +54,12 @@
         if (idx < 0 || idx >= (type##_pool.bitset)->cap)                \
             return NULL;                                                \
         return &type##_pool.objects[idx];                               \
-    }                                                                   \
-    bool pool_valid_at_##type(int idx)                                  \
-    {                                                                   \
-        return bitset_get_idx(type##_pool.bitset, idx);                 \
     }
 
-#define POOL_GET(type)           pool_get_##type()
-#define POOL_FREE(type, obj)     pool_free_##type(obj)
-#define POOL_IDX(type, obj)      pool_idx_##type(obj)      // the index of the object
-#define POOL_AT(type, idx)       pool_at_##type(idx)       // the object at
-#define POOL_VALID_AT(type, idx) pool_valid_at_##type(idx) // the object at idx is currently valid
+#define POOL_GET(type)       pool_get_##type()
+#define POOL_FREE(type, obj) pool_free_##type(obj)
+#define POOL_IDX(type, obj)  pool_idx_##type(obj) // the index of the object
+#define POOL_AT(type, idx)   pool_at_##type(idx)  // the object at
 
 #define POOL_ENTRY(name, capacity) POOL_DECLARE_TYPE(name);
 #include POOLS_DEF_FILE
--- a/source/game.c
+++ b/source/game.c
@@ -195,7 +195,7 @@ static const Rect DECK_SIZE_RECT            = {200,     152,    240,       160  
 static const Rect ROUND_TEXT_RECT           = {48,      144,    UNDEFINED, UNDEFINED };
 static const Rect ANTE_TEXT_RECT            = {8,       144,    UNDEFINED, UNDEFINED };
 
-static const BG_POINT CARD_DRAW_POS         = {210,     118};
+static const BG_POINT CARD_DRAW_POS         = {208,     118};
 static const BG_POINT CARD_DISCARD_PNT      = {240,     70};
 static const BG_POINT HAND_START_POS        = {120,     90};
 static const BG_POINT HAND_PLAY_POS         = {120,     70};
--- a/source/game.c
+++ b/source/game.c
@@ -549,10 +549,11 @@ void game_update()
 
     g_game_vars.timer++;
 
-    sprite_object_update_all();
     jokers_update_loop();
 
     state_machine_update();
+
+    sprite_object_update_all();
 }
 
 void game_change_state(enum GameState new_game_state)
@@ -2258,9 +2259,6 @@ static inline void game_playing_discarded_cards_loop(void)
             discarded_card_object->sprite_object->ty = int2fx(112);
             discarded_card_object->sprite_object->x = int2fx(240);
             discarded_card_object->sprite_object->y = int2fx(80);
-
-            // Update immediately to avoid flickering sprites at position (0,0)
-            sprite_object_update(discarded_card_object->sprite_object);
         }
         else
         {
--- a/source/game.c
+++ b/source/game.c
@@ -195,7 +195,7 @@ static const Rect DECK_SIZE_RECT            = {200,     152,    240,       160  
 static const Rect ROUND_TEXT_RECT           = {48,      144,    UNDEFINED, UNDEFINED };
 static const Rect ANTE_TEXT_RECT            = {8,       144,    UNDEFINED, UNDEFINED };
 
-static const BG_POINT CARD_DRAW_POS         = {204,     114};
+static const BG_POINT CARD_DRAW_POS         = {210,     118};
 static const BG_POINT CARD_DISCARD_PNT      = {240,     70};
 static const BG_POINT HAND_START_POS        = {120,     90};
 static const BG_POINT HAND_PLAY_POS         = {120,     70};
--- a/source/game.c
+++ b/source/game.c
@@ -195,7 +195,7 @@ static const Rect DECK_SIZE_RECT            = {200,     152,    240,       160  
 static const Rect ROUND_TEXT_RECT           = {48,      144,    UNDEFINED, UNDEFINED };
 static const Rect ANTE_TEXT_RECT            = {8,       144,    UNDEFINED, UNDEFINED };
 
-static const BG_POINT CARD_DRAW_POS         = {208,     110};
+static const BG_POINT CARD_DRAW_POS         = {204,     114};
 static const BG_POINT CARD_DISCARD_PNT      = {240,     70};
 static const BG_POINT HAND_START_POS        = {120,     90};
 static const BG_POINT HAND_PLAY_POS         = {120,     70};
@@ -467,7 +467,6 @@ static inline void discarded_jokers_update_loop(void)
 
     while ((joker_object = list_itr_next(&itr)))
     {
-        joker_object_update(joker_object);
         if (joker_object->sprite_object->x == joker_object->sprite_object->tx &&
             joker_object->sprite_object->y == joker_object->sprite_object->ty)
         {
@@ -499,7 +498,6 @@ static inline void held_jokers_update_loop(void)
         if (joker != game_shop_get_description_card())
             joker->sprite_object->tx = hand_x - int2fx(spacing_lut[jokers_top][i]);
         i++;
-        joker_object_update(joker);
     }
 }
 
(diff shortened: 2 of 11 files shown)


MathisMartin31:
@MeirGavish Sorry for the update while you're reviewing, I did notice a slight flicker when playing cards so I moved the call to `sprite_object_update_all` to the end of the game update loop and it's fixed now.
I also could remove the last calls to `sprite_object_update` as there was no flicker left

MeirGavish:
No problem, I still haven't started reviewing, I just requested a review from myself to mark this as waiting for review from me.

source/sprite.c
```diff
@@ -304,6 +304,17 @@ IWRAM_CODE void sprite_object_update(SpriteObject* sprite_object)
     sprite_position(sprite_object->sprite, fx2int(sprite_object->x), fx2int(sprite_object->y));
 }
 
+IWRAM_CODE void sprite_object_update_all(void)
+{
+    SpriteObject* sprite_object = NULL;
+    for (int i = 0; i < MAX_SPRITE_OBJECTS; i++)
+    {
+        sprite_object = POOL_AT(SpriteObject, i);
+        if (POOL_VALID_AT(SpriteObject, i) && sprite_object != NULL)
```
ricfehr3: Nitpick: I think we should change `POOL_VALID_AT`to `!POOL_FREE_AT`. Very small nitpick, I may have thought on it too hard, but it makes more sense to me. Feel free to disregard.

source/sprite.c
```diff
@@ -304,6 +304,17 @@ IWRAM_CODE void sprite_object_update(SpriteObject* sprite_object)
     sprite_position(sprite_object->sprite, fx2int(sprite_object->x), fx2int(sprite_object->y));
 }
 
+IWRAM_CODE void sprite_object_update_all(void)
+{
+    SpriteObject* sprite_object = NULL;
+    for (int i = 0; i < MAX_SPRITE_OBJECTS; i++)
```
ricfehr3: Can you put a comment in the header or here saying that as long as `MAX_SPRITE_OBJECTS` is reasonably sized, there shouldn't be an issue with going over every pool entry.

ricfehr3 (changes requested):
I dig this. I initially thought there would need to be some list of active `SpriteObject`s to update, which would require some logic to manage the list, plus requiring all `SpriteObject` usage to fit to that new style. But this is rather elegant, IMO, as it uses of the pool's properties to be that list, with the existing system in place.

I left a few comments and I would like some time to test as well, but I think this is a good change once all the update calls are in order.

MeirGavish:
Sorry for the delay, I had a busy day yesterday and didn't get a chance to properly comment.
I actually prefer to have a list of `SpriteObject`s and hook managing it into `sprite_object_new()` and `sprite_object_destroy()`. 
One reason is I think performance would be better since you iterate over a smaller list.
But the main reason is that I think the memory management system should be decoupled from the update cycle,
I was actually removing `SpriteObject` from the pool in my `Item` PR.
Theoretically we should be able to change back the memory management system to malloc without having to change anything in the uodate cycle.

ricfehr3:
> Sorry for the delay, I had a busy day yesterday and didn't get a chance to properly comment. I actually prefer to have a list of `SpriteObject`s and hook managing it into `sprite_object_new()` and `sprite_object_destroy()`. One reason is I think performance would be better since you iterate over a smaller list. But the main reason is that I think the memory management system should be decoupled from the update cycle, I was actually removing `SpriteObject` from the pool in my `Item` PR. Theoretically we should be able to change back the memory management system to malloc without having to change anything in the uodate cycle.

Oh, true. Very good point @MeirGavish . Decoupling the systems is important. It is nice that the pool provides some of the functionality already, but I do agree that decoupling is more important. 

Maybe we could hook into `sprite_object_new` instead and remove the new POOL functions and just manage the list there?

MeirGavish (changes requested):
Just marking my review as request changes for the reasons in my [previous comment](https://github.com/GBALATRO/balatro-gba/pull/553#issuecomment-4697542379)

MathisMartin31:
@MeirGavish @ricfehr3 The thing that worries me with Lists isn't the performance when updating, but the performance when destroying SpriteObjects because we need to traverse the list and pop the right element every time..

For instance, when we draw or play cards (when they enter/exit our hand), we rearrange or sort the cards, which involves destroying all sprites and recreating them in the right order, which is going to be a lot of comparisons.

I'll try and see if it works out in the end.

MeirGavish:
> @MeirGavish @ricfehr3 The thing that worries me with Lists isn't the performance when updating, but the performance when destroying SpriteObjects because we need to traverse the list and pop the right element every time..
> 
> For instance, when we draw or play cards (when they enter/exit our hand), we rearrange or sort the cards, which involves destroying all sprites and recreating them in the right order, which is going to be a lot of comparisons.
> 
> I'll try and see if it works out in the end.

That's a good point, I guess this can benefit from #509.

MathisMartin31:
Huh... I did it with a List and huuuuh, it's not looking too good ^^'

Screenshot_20260614_001712

Here's the code that changed, apart from `pool.c` that I reverted:

```
static List sprite_objects_list = LIST_DEFAULT;
```

```
SpriteObject* sprite_object_new()
{
    SpriteObject* sprite_object = POOL_GET(SpriteObject);
    sprite_object->sprite = NULL;
    sprite_object_reset_transform(sprite_object);
    sprite_object->focused = false;

    list_push_back(&sprite_objects_list, sprite_object);

    return sprite_object;
}

void sprite_object_destroy(SpriteObject** sprite_object)
{
    if (*sprite_object == NULL)
        return;

    list_remove_data(&sprite_objects_list, *sprite_object);

    sprite_destroy(&(*sprite_object)->sprite);
    POOL_FREE(SpriteObject, *sprite_object);
    *sprite_object = NULL;
}
```

```
IWRAM_CODE void sprite_object_update_all(void)
{
    int nb_sprites = list_get_len(&sprite_objects_list);
    for (u32 i = 0; i < nb_sprites; i++)
    {
        sprite_object_update(list_get_at_idx(&sprite_objects_list, i));
    }
}
```

I did try to do the update loop properly but it ended up weird too, like any variable we print gives an empty string

```
IWRAM_CODE void sprite_object_update_all(void)
{
    SpriteObject* sprite_object = NULL;
    ListItr itr = list_itr_create(&sprite_objects_list);
    while ((sprite_object = list_itr_next(&itr)))
    {
        sprite_object_update(sprite_object);
    }
}
```

https://github.com/user-attachments/assets/7654bc0a-7b00-4267-bd9c-0b87cc0a1121

MathisMartin31:
It's crashing at a place completely unrelated, we're not even deleting/creating SpriteObjects when selecting a card, we're just shifting them up.

Is there any way to check if we're not running out of memory?

ricfehr3:
Odd behavior... Sounds like the pool get/free pointers may be different?

Can you push your branch or give me a patch from `git diff main`? I'd like to hop in the debugger.

We should never run out of memory because we have everything statically allocated. I figure something is dereferencing a NULL pointer elsewhere or who knows.

MathisMartin31:
> Can you push your branch or give me a patch from git diff main? I'd like to hop in the debugger.

Done ^^

And just so you don't have to search for it, it seems to crash when calling `tte_printf` on `chips_str_buff` in the `display_chips(void)` function than ends up being called after selecting a card

MeirGavish:
Seems like a memory corruption error. Check if things are working correctly with the pool.

MeirGavish:
@MathisMartin31 @ricfehr3 
The crashes you're seeing may be related to #560

MathisMartin31:
> @MathisMartin31 @ricfehr3 The crashes you're seeing may be related to #560

Maybe, I don't really know because the address the error message gives me is waaaay out of bounds from either EWRAM or IWRAM (`0x0203FFFF` and `0x03007FFF` respectively) at `0xA02004E2`

That plus the crash is not happening when accessing the List directly, but when printing Chips on the screen.

MathisMartin31:
Hmm, so I removed the `IWRAM_CODE` directive from the `sprite_object_update_all` function and it worked :thinking:

MathisMartin31:
Also another thing I'm thinking of since I was looking at the memory pools, is that now that we put pointer to every existing SpriteObject into a List, it means that we automatically have at least as many ListNodes as there are SpriteObjects/Sprites in memory, but we have `MAX_LIST_NODES = 128`, so there will be a point where, if we have too many Sprites in memory, and since we have Lists for other things, we will run out of space in the pool for new ListNodes before hitting 128 sprites.

include/pool.h
```diff
@@ -21,7 +21,8 @@
     type* pool_get_##type();          \
     void pool_free_##type(type* obj); \
     int pool_idx_##type(type* obj);   \
-    type* pool_at_##type(int idx);
+    type* pool_at_##type(int idx);    \
+    bool pool_valid_at_##type(int idx);
```
MeirGavish: Is this still being used/defined? If not the declaration should be removed.

source/game.c
```diff
@@ -195,7 +195,7 @@ static const Rect DECK_SIZE_RECT            = {200,     152,    240,       160
 static const Rect ROUND_TEXT_RECT           = {48,      144,    UNDEFINED, UNDEFINED };
 static const Rect ANTE_TEXT_RECT            = {8,       144,    UNDEFINED, UNDEFINED };
 
-static const BG_POINT CARD_DRAW_POS         = {208,     110};
+static const BG_POINT CARD_DRAW_POS         = {204,     114};
```
MeirGavish: This change is only to make things look better, right?

source/sprite.c
```diff
@@ -304,6 +311,16 @@ IWRAM_CODE void sprite_object_update(SpriteObject* sprite_object)
     sprite_position(sprite_object->sprite, fx2int(sprite_object->x), fx2int(sprite_object->y));
 }
 
+void sprite_object_update_all(void)
```
MeirGavish, line 311: Maybe just add a TODO here that it could be good to put this in IWRAM?

MeirGavish (changes requested):
Looks great! Great job! This refactor really cleans up the code by moving the update responsibility to a single place. I tested it and it seems to run nicely and there are no performance issues.  Though thinking about it maybe running a performance profile test is due...
Of course #427 still exists but it doesn't seem to have gotten worse.
Just a few small comments, requesting changes mainly to address what seems like unused leftover code.

include/pool.h
MathisMartin31: Whoops, nope not used anymore I missed it when reverting stuff

source/game.c
MathisMartin31: Yes it is, I found that the cards didn't quite spawn in the right place

MathisMartin31: I tweaked it a bit more, now I think it's good ^^

MeirGavish (approved):
Looks good! Approving.

I also did the profiling, as described [here](https://github.com/GBALATRO/balatro-gba/pull/526#issuecomment-4728315866) for 4 discards and here are the results,

[gdb-output-553.txt](https://github.com/user-attachments/files/29042512/gdb-output-553.txt)

```
./scripts/performance/parse_gdb_logs.sh ./logs/gdb-output-553.txt
avg = 74463.87
min = 50687
max = 185623
```

As compared to [main](https://github.com/GBALATRO/balatro-gba/issues/427#issuecomment-4728410473), it seems like there's no performance degradation. In fact, the results seem slightly better though I think this is within uncontrolled random variation.

MathisMartin31:
> ```
> ./scripts/performance/parse_gdb_logs.sh ./logs/g
> db-output-553.txt
> avg = 74463.87
> min = 50687
> max = 185623
> ```

Wait, I thought our cycles budget was around 55k cycles though, isn't an average of 75k pretty bad ?

MeirGavish:
> > ```
> > ./scripts/performance/parse_gdb_logs.sh ./logs/g
> > db-output-553.txt
> > avg = 74463.87
> > min = 50687
> > max = 185623
> > ```
> 
> Wait, I thought our cycles budget was around 55k cycles though, isn't an average of 75k pretty bad ?

Quoting my [comment](https://github.com/GBALATRO/balatro-gba/pull/454#issuecomment-4374780521)
> One thing I want to note about the numbers though, is that the 280000 cycle number is for an entire frame including both VBlank and VDraw. But our code doesn't run in VDraw, only in VBlank. In `main()` you can see our code runs by waiting for VBlank with `VBlankIntrWait()` and only then running everything, including `update()` that runs all the logic, and `draw()` being last. We don't want `draw()` to run during VDraw (probably explains why performance issues manifest as sprite flickers) so everything has to happen within VBlank, leaving us only with the 83000 or so cycles of VBlank to do everything.
> 
> It also seems like all Tonc demos work like this and never try to do anything during VDraw. I tried playing around with the order of calls in `main()` to maybe allow things to be done during VDraw but it created different sprite flicker issues.
> Edit: Also the times are from here https://gbadev.net/tonc/video.html#sec-blanks

So, theoretically our budget is about 83K.
When performing these tests (including [these](https://github.com/GBALATRO/balatro-gba/issues/427#issuecomment-4728410473)) I perform discards with 5 jokers owned, I think that's why the performance is worse than [these results](https://github.com/GBALATRO/balatro-gba/pull/454#pullrequestreview-4217005236).
And I *think* spikes above 83K are what causes the sprite tear, though maybe we are more tolerant to spikes there than I thought, though we are also seeing spikes above 180K which is pretty significant.

MathisMartin31:
> > > ```
> > > ./scripts/performance/parse_gdb_logs.sh ./logs/g
> > > db-output-553.txt
> > > avg = 74463.87
> > > min = 50687
> > > max = 185623
> > > ```
> > 
> > Wait, I thought our cycles budget was around 55k cycles though, isn't an average of 75k pretty bad ?
> 
> Quoting my [comment](https://github.com/GBALATRO/balatro-gba/pull/454#issuecomment-4374780521)
> > One thing I want to note about the numbers though, is that the 280000 cycle number is for an entire frame including both VBlank and VDraw. But our code doesn't run in VDraw, only in VBlank. In `main()` you can see our code runs by waiting for VBlank with `VBlankIntrWait()` and only then running everything, including `update()` that runs all the logic, and `draw()` being last. We don't want `draw()` to run during VDraw (probably explains why performance issues manifest as sprite flickers) so everything has to happen within VBlank, leaving us only with the 83000 or so cycles of VBlank to do everything.
> > 
> > It also seems like all Tonc demos work like this and never try to do anything during VDraw. I tried playing around with the order of calls in `main()` to maybe allow things to be done during VDraw but it created different sprite flicker issues.
> > Edit: Also the times are from here https://gbadev.net/tonc/video.html#sec-blanks
> 
> So, theoretically our budget is about 83K.
> When performing these tests (including [these](https://github.com/GBALATRO/balatro-gba/issues/427#issuecomment-4728410473)) I perform discards with 5 jokers owned, I think that's why the performance is worse than [these results](https://github.com/GBALATRO/balatro-gba/pull/454#pullrequestreview-4217005236).
> And I *think* spikes above 83K are what causes the sprite tear, though maybe we are more tolerant to spikes there than I thought, though we are also seeing spikes above 180K which is pretty significant.

Okay thanks, I had my numbers mixed up.

Yes it looks like the spikes over 180k could be one of the reasons why we have screen tearing.

And at 75k on average we're cutting it pretty close so I think in the near future we should do a pass over all big functions, see which take the most cycles to do some targeted optimizations.

ricfehr3:
> Looks good! Approving.
> 
> I also did the profiling, as described [here](https://github.com/GBALATRO/balatro-gba/pull/526#issuecomment-4728315866) for 4 discards and here are the results,
> 
> [gdb-output-553.txt](https://github.com/user-attachments/files/29042512/gdb-output-553.txt)
> 
> ```
> ./scripts/performance/parse_gdb_logs.sh ./logs/gdb-output-553.txt
> avg = 74463.87
> min = 50687
> max = 185623
> ```
> 
> As compared to [main](https://github.com/GBALATRO/balatro-gba/issues/427#issuecomment-4728410473), it seems like there's no performance degradation. In fact, the results seem slightly better though I think this is within uncontrolled random variation.

Nice! @MeirGavish Really nice work profiling the game here. I wanted to also say (as @MathisMartin31  said) that whatever that 185K spike is would probably explain the screen tearing. I'm excited that we have some information to _why_ it's happening now.

ricfehr3 (approved):
Looks great! @MathisMartin31 I'll go ahead and merge this since @MeirGavish also agreed it's good.

PR #554, opened 2026-06-13
Remove automatic releases on tags

ricfehr3:
Remove the workflow to automatically deploy releases. This workflow doesn't really fit how we work, since we prefer to have a more fine-tuned release cycle. 

I need this to start tagging. Since it will make a release on any tag with a `v` after it.


ricfehr3:
@MathisMartin31 @MeirGavish I'm going to just merge this. Since it's removing something we don't use. I don't think we want to use it really either. If we do need it, we can revert this commit, but I want the freedom to make a tag without accidentally doing a release right now.

PR #557, opened 2026-06-14
Add logging macros that include the function name

### LINKED ISSUE #555: Add error logging macros that include the function name
Currently we have excellent macros to print to the mGBA log. I find it useful to always include the function name in the log, that way we  can log an error such as "Called with NULL parameter" and automatically have the function name.
C99 has a `__func__`, a [Predefined Identifier](https://www.geeksforgeeks.org/c/predefined-identifier-func-in-c/) that always has the current function name.
Incorporating it in macros will help achieve that.

MeirGavish:
Closes #555 

I opted to not replace the existing log functions and macros to allow logging without the function name.
I extracted `mgba_vprintf` that accepts a `va_list` from `mgba_printf`, that way I can call it from my other function.
I am wondering whether to change the names of the log macros to from `MGBA_ERROR`, `MGBA_WARN`, etc, to `MGBA_LOG_ERROR`, `MGBA_LOG_WARN` to make it a little clearer that they write to a log.
That way I could also change my added macros `MGBA_FUNC_ERROR`, `MGBA_FUNC_WARN` to `MGBA_FUNC_LOG_ERROR`, `MGBA_FUNC_LOG_WARN` which I think would make them clearer.

### DIFF
--- a/include/mgba_logger.h
+++ b/include/mgba_logger.h
@@ -17,7 +17,7 @@
  * mgba -l 14 game.rom # INFO, WARN, and ERROR
  * ```
  *
- * @note You have to fight with other logs in mgba and DEBUG can get messy.
+ * @note You have to fight with other logs in mgba and INFO can get messy.
  *
  * @note FATAL does kill the game. Use with care.
  */
--- a/include/mgba_logger.h
+++ b/include/mgba_logger.h
@@ -63,7 +63,6 @@ void mgba_printf(MgbaLogLevel level, const char* fmt, ...);
  * @param ... variadic arguments
  *
  * @note for all logs, it's cutoff at the hard mgba limit of 0x100
- * @note The function name may be dropped if the resulting string exceeds this limit.
  */
 void mgba_func_printf(MgbaLogLevel level, const char* func_name, const char* fmt, ...);
 
--- a/include/mgba_logger.h
+++ b/include/mgba_logger.h
@@ -101,18 +101,18 @@ void mgba_func_printf(MgbaLogLevel level, const char* func_name, const char* fmt
 #endif
 // clang-format on
 
-#define CHECK_NULL_ARG_VOID(param)\
-    if ((param) == NULL)\
-    {\
+#define CHECK_NULL_ARG_VOID(param)                                 \
+    if ((param) == NULL)                                           \
+    {                                                              \
         MGBA_FUNC_ERROR("Called with NULL '%s' argument", #param); \
-        return;\
+        return;                                                    \
     }
 
-#define CHECK_NULL_ARG_RET(param, ret_val)\
-    if ((param) == NULL)\
-    {\
+#define CHECK_NULL_ARG_RET(param, ret_val)                         \
+    if ((param) == NULL)                                           \
+    {                                                              \
         MGBA_FUNC_ERROR("Called with NULL '%s' argument", #param); \
-        return ret_val;\
+        return ret_val;                                            \
     }
 
 #endif
--- a/include/mgba_logger.h
+++ b/include/mgba_logger.h
@@ -56,17 +56,34 @@ void mgba_printf(MgbaLogLevel level, const char* fmt, ...);
 
 // clang-format off
 #ifdef MGBA_LOGGING
+
 #define MGBA_FATAL(...) mgba_printf(MGBA_LOG_FATAL, __VA_ARGS__)
 #define MGBA_ERROR(...) mgba_printf(MGBA_LOG_ERROR, __VA_ARGS__)
 #define MGBA_WARN(...)  mgba_printf(MGBA_LOG_WARN,  __VA_ARGS__)
 #define MGBA_INFO(...)  mgba_printf(MGBA_LOG_INFO,  __VA_ARGS__)
 #define MGBA_DEBUG(...) mgba_printf(MGBA_LOG_DEBUG, __VA_ARGS__)
+
+#define MGBA_FUNC_LOG(level, ...) mgba_func_printf(level, __func__, __VA_ARGS__)
+
+#define MGBA_FUNC_FATAL(...) MGBA_FUNC_LOG(MGBA_LOG_FATAL, __VA_ARGS__)
+#define MGBA_FUNC_ERROR(...) MGBA_FUNC_LOG(MGBA_LOG_ERROR, __VA_ARGS__)
+#define MGBA_FUNC_WARN(...)  MGBA_FUNC_LOG(MGBA_LOG_WARN,  __VA_ARGS__)
+#define MGBA_FUNC_INFO(...)  MGBA_FUNC_LOG(MGBA_LOG_INFO,  __VA_ARGS__)
+#define MGBA_FUNC_DEBUG(...) MGBA_FUNC_LOG(MGBA_LOG_DEBUG, __VA_ARGS__)
+
 #else
+          
 #define MGBA_FATAL(...) ((void)0)
 #define MGBA_ERROR(...) ((void)0)
 #define MGBA_WARN(...) ((void)0)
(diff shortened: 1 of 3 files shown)


include/mgba_logger.h
MeirGavish: By the way, when testing I saw  it was actually the INFO tab that's being spammed by mGBA and didn't see anything in the DEBUG.

include/mgba_logger.h
```diff
@@ -54,20 +54,78 @@ bool mgba_logger_init(void);
...
+ * us for a void function.
+ */
+#define CHECK_NULL_ARG_RET(param, ret_val)                         \
+    if ((param) == NULL)                                           \
+    {                                                              \
+        MGBA_FUNC_ERROR("Called with NULL '%s' argument", #param); \
+        return ret_val;                                            \
+    }
+
 #endif
```
MeirGavish: I also added these macros, I think they will be very nice especially for https://github.com/GBALATRO/balatro-gba/issues/558 as they can really simplify the argument checking boilerplate.
I wasn't sure where exactly to put them but I figured this is a good enough place. I thought of util.h but I don't want it to depend on mgba_logger.h as it should not depend on GBA specific code.

source/mgba_logger.c
```diff
@@ -26,18 +27,35 @@ bool mgba_logger_init(void)
...
+void mgba_func_printf(MgbaLogLevel level, const char* func_name, const char* fmt, ...)
+{
+    char func_name_fmt_buff[MGBA_LOG_BUFFER_SIZE];
+    snprintf(func_name_fmt_buff, sizeof(func_name_fmt_buff), "%s(): %s", func_name, fmt);
+
+    va_list args;
+    va_start(args, fmt);
+    mgba_vprintf(level, func_name_fmt_buff, args);
+    va_end(args);
 }
```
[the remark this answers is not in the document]
MeirGavish: That's a valid point, I don't want to solve it like this, I still want to call `mgba_vprintf()`. I will add checks.

include/mgba_logger.h
```diff
@@ -54,20 +54,78 @@ bool mgba_logger_init(void);
...
+ * Useful for checking arguments to a function.
+ * This version is for a void function, while @ref CHECK_NULL_ARG_RET is for one with
+ * a return value.
+ */
+#define CHECK_NULL_ARG_VOID(param)                                 \
+    if ((param) == NULL)                                           \
+    {                                                              \
+        MGBA_FUNC_ERROR("Called with NULL '%s' argument", #param); \
+        return;                                                    \
+    }
```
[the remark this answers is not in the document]
MeirGavish: Generally I try to use curly braces in all my conditional execution statements and avoid code of the following form that can be problematic,

```c
if (condition)
    CHECK_NULL_ARG_VOID(arg);
else
    do_something();
```

But it's better to keep my macros safe so I'll take it.

include/mgba_logger.h
```diff
@@ -54,20 +54,78 @@ bool mgba_logger_init(void);
...
+ * @param ret_val The value to return in case @p param is NULL.
+ * This version is for a function that returns a value while @ref CHECK_NULL_ARG_VOID
+ * us for a void function.
+ */
+#define CHECK_NULL_ARG_RET(param, ret_val)                         \
+    if ((param) == NULL)                                           \
+    {                                                              \
+        MGBA_FUNC_ERROR("Called with NULL '%s' argument", #param); \
+        return ret_val;                                            \
+    }
```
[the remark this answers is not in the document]
MeirGavish: See https://github.com/GBALATRO/balatro-gba/pull/557#discussion_r3411440522

include/mgba_logger.h
ricfehr3:
```suggestion
 * @note You have to fight with other logs in mgba and INFO can get messy.
```

I think we should change it, I double checked and you are correct.

include/mgba_logger.h
ricfehr3: I dig this. I usually do this in other codebases and I'm a huge fan of the pattern. 

I also feel this should go into `util.h`, and I think we should consider using these macros in a couple ways. It's convenient to put some other conditionally compiled error handling here. 

For example, I have been using `exit(1)` quickly test if I'm reaching certain conditions. I would love to have a way to add a central `gbal_exit()` function. The debugger can break and get back traces on failed "assertions". 

Because the logging and debugger error handling is not dependent on each other, I think we should keep these in `util.h`. Then we can use `#ifdef` blocks to conditionally add logging and/or debugger breakpoints.

source/mgba_logger.c
```diff
@@ -26,18 +27,50 @@ bool mgba_logger_init(void)
...
+    int chars_used =
+        snprintf(func_name_fmt_buff, sizeof(func_name_fmt_buff), "%s(): %s", func_name, fmt);
+
+    char* printed_str = func_name_fmt_buff;
+
+    if ((size_t)chars_used > sizeof(func_name_fmt_buff) - 1)
+    {
+        // The buffer is not large enough, format has been truncated, drop the function name
+        printed_str = fmt;
+    }
```
ricfehr3: I do think it's a cool idea to do this check, I'm questioning if all of this is worth it to conditionally display the function name. It makes sense to me to always have the function name, and though our function names can get rather long, 256 characters is still a lot to work with.

For example here is 256 characters:
```
Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia volupta
```

If we go over that limit, I think it's reasonable to just make a second print call, and instead we can simply truncate the string.

ricfehr3 (changes requested):
Sweet! I always like seeing things added that make our lives easier 😅 . Great work! I have some comments. I don't feel too strong about the truncate vs. function name dropping thing, but I wanted to share my thoughts. Thanks @MeirGavish

include/mgba_logger.h
MeirGavish: Done, I moved it to util.h with some more conditionally compiled code

source/mgba_logger.c
MeirGavish: I changed it to cutoff the string instead of dropping the function. The result however is that I no longer use the `mgba_vprintf()`. I could delete it but I thought I'd leave it in case it becomes useful.

include/mgba_logger.h
MeirGavish: @ricfehr3 I made another change to this. I thought it would be more useful to make it generic and not specific to function arguments so it can check any error values during control flow and return on error. For example for return values for other functions.

```
static inline void card_draw(void)
{
    CardObject* card_object = card_object_new(deck_pop());
    GBAL_RETURN_IF_NULL_VOID(card_object);
...
}
```

And I thought it would make sense to make it generic to any error values, not just `NULL`, such as `UNDEFINED`, but looking at the code I didn't see much use cases for it so I reverted that.

ricfehr3 (approved):
Nice cleanup and good addition. Thanks! I gave it another review and some testing and I think it's good to go now. I also wanted to say we could add the file and line number with `__FILE__` and `__LINE__` provided by gcc, but just the function name makes the most sense for space, but that is an option. 

Please merge when you would like, it's all good from my end.

MeirGavish:
> Nice cleanup and good addition. Thanks! I gave it another review and some testing and I think it's good to go now. I also wanted to say we could add the file and line number with `__FILE__` and `__LINE__` provided by gcc, but just the function name makes the most sense for space, but that is an option.
> 
> Please merge when you would like, it's all good from my end.

Thanks! Yes, I've been debating `__FILE__` and `__LINE__`  myself but was hesitant due to buffer length limit... I guess we can always add them if we see they would be useful.
Merging!

PR #559, opened 2026-06-14
Implement new `sprite_object_position` function

### LINKED ISSUE #543: Make easier function to set `SpriteObject` position
~~Pretty self-explanatory. We need to set a few variables up when creating `SpriteObject`s. It would be nice to have some default value/constructor to simplify things.~~ https://github.com/GBALATRO/balatro-gba/issues/543#issuecomment-4643788318 , **Correction**:

Setting a `SpriteObjects` position requires setting the target of the `SpriteObject`. We should add a helper function so It's easier.

```c
void sprite_object_set_position(SpriteObject*, x, y)
{
  sprite_object->x = x;
  sprite_object->y = y;
  sprite_object->tx = x;
  sprite_object->ty = y
}
```

MathisMartin31:
Closes #543.

Function is inlined for efficiency just line `sprite_position`

Although I'm starting to wonder if we should do a `position` and `target` variants of this function because in the end I couldn't use it much, since the position and target position are usually not the same to achieve some sort of animation.

### DIFF
--- a/source/game/run_setup.c
+++ b/source/game/run_setup.c
@@ -583,12 +583,11 @@ void game_run_setup_on_init(void)
     // both cards do not exist at the same time, one is destroyed before the other is created.
     card_object_set_sprite_face_down(run_setup_deck, g_game_vars.deck, 0);
 
-    run_setup_deck->sprite_object->tx = int2fx(RUN_SETUP_DECK_SPRITE_T_X);
-    run_setup_deck->sprite_object->ty = int2fx(RUN_SETUP_DECK_SPRITE_T_Y);
-
-    // Need to do -1 here so that the sprite isn't at the target position and gets updated
-    run_setup_deck->sprite_object->x = run_setup_deck->sprite_object->tx - 1;
-    run_setup_deck->sprite_object->y = run_setup_deck->sprite_object->ty - 1;
+    sprite_object_position(
+        run_setup_deck->sprite_object,
+        RUN_SETUP_DECK_SPRITE_T_X,
+        RUN_SETUP_DECK_SPRITE_T_Y
+    );
 
     card_object_update(run_setup_deck);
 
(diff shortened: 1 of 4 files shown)


source/game/run_setup.c
```diff
@@ -583,12 +583,11 @@ void game_run_setup_on_init(void)
...
-    run_setup_deck->sprite_object->ty = int2fx(RUN_SETUP_DECK_SPRITE_T_Y);
-
-    // Need to do -1 here so that the sprite isn't at the target position and gets updated
-    run_setup_deck->sprite_object->x = run_setup_deck->sprite_object->tx - 1;
-    run_setup_deck->sprite_object->y = run_setup_deck->sprite_object->ty - 1;
+    sprite_object_position(
+        run_setup_deck->sprite_object,
+        RUN_SETUP_DECK_SPRITE_T_X,
+        RUN_SETUP_DECK_SPRITE_T_Y
+    );
```
MeirGavish, line 592: Does it no longer need to be - 1 here?

MeirGavish (commented):
Looks good! I bet  there are other places where it can be used but this is a good start.
I just have this one question to make sure the change is okay.

source/game/run_setup.c
MathisMartin31: Nope, I don't really know what I was smoking there but it sure was good lol.

I tested it and the sprite is placed correctly

MeirGavish: lol Okay, I'll approve then.

MeirGavish (approved):

MeirGavish:
@ricfehr3 
I'll merge this one, it's pretty small, I hope it's okay, and that maybe it removes a bit of the PR load from you.

ricfehr3:
@MathisMartin31 Great lil' change. I love to see it. @MeirGavish By all means, if you think a PR is small enough to just merge, go for it. I'll see it anyway and I trust your judgement. Also, less PR work is nice :sweat_smile: . Thanks!

PR #541, opened 2026-06-06
[Feature] Implement a few Skip Tags

MathisMartin31:
Does close #353 though not all Tags in scope are implemented here.

This PR adds only 8 Skip Tags: Investment, Boss, Handy, Garbage, Double, Top Up, Speed, and Economy. This isn't a lot, but it's still a lot of code so I think it'd be good to start with this and add some more later, when the refactors are over and the code is more stable.

Here's an example of a combination of the Double and Economy Tags working together:

https://github.com/user-attachments/assets/0eb95f02-8f41-4119-ae3e-665d242e10ee

PR #561, opened 2026-06-14
Moved pools and bitsets to EWRAM

### LINKED ISSUE #560: Move pools to EWRAM
As seen in some PRs recently (https://github.com/GBALATRO/balatro-gba/pull/563, https://github.com/GBALATRO/balatro-gba/pull/553), making some changes creates unexpected crashes.
These crashes may be related to running out of IWRAM as we are close to filling it (see #565).

Running get_memory_map.sh shows the pools are mapped to IWRAM,
```
Sprite          | 0x030032ac | 2560       | 40         | 32         
SpriteObject    | 0x03002f0c | 896        | 40         | 32         
Joker           | 0x030026a4 | 96         | 40         | 32         
JokerObject     | 0x03002704 | 64         | 36         | 32         
Card            | 0x03002744 | 104        | 36         | 32         
CardObject      | 0x03002dcc | 192        | 56         | 32         
ListNode        | 0x030027ac | 1536       | 40         | 32         
Total bytes used: 5960
```

Since the pools are our main memory allocation management system, I think it doesn't make sense for them to be in IWRAM, as IWRAM is small and fast and we want to be able to store many things so we need a large memory section for the pool and I think EWRAM is much more suited.

Or a better solution would be to update the pool API to be able to determine what section each pool is placed.

MeirGavish:
So the issues I've seen with this are sprite flickers when rerolling.
Edit: seems like this isn't caused by this change: #568

https://github.com/user-attachments/assets/95c250b8-6a30-4a2a-8703-207010171abb

Frame by frame

image
image
image

### DIFF
--- a/include/bitset.h
+++ b/include/bitset.h
@@ -196,8 +196,8 @@ int bitset_itr_next(BitsetItr* itr);
  * @param capacity the capacity of the bitset
  */
 #define BITSET_DEFINE(name, capacity)                  \
-    static uint32_t name##_w[BITSET_ARRAY_SIZE] = {0}; \
-    static Bitset name = {                             \
+    EWRAM_DATA static uint32_t name##_w[BITSET_ARRAY_SIZE] = {0}; \
+    EWRAM_DATA static Bitset name = {                             \
         .w = name##_w,                                 \
         .nbits = BITSET_BITS_PER_WORD,                 \
         .nwords = BITSET_ARRAY_SIZE,                   \
--- a/include/pool.h
+++ b/include/pool.h
@@ -25,8 +25,8 @@
 
 #define POOL_DEFINE_TYPE(type, capacity)                                \
     BITSET_DEFINE(type##_bitset, capacity)                              \
-    EWRAM_DATA static type type##_storage[capacity];                               \
-    EWRAM_DATA static type##Pool type##_pool = {                                   \
+    EWRAM_DATA static type type##_storage[capacity];                    \
+    EWRAM_DATA static type##Pool type##_pool = {                        \
         .bitset = &type##_bitset,                                       \
         .objects = type##_storage,                                      \
     };                                                                  \
--- a/include/pool.h
+++ b/include/pool.h
@@ -25,8 +25,8 @@
 
 #define POOL_DEFINE_TYPE(type, capacity)                                \
     BITSET_DEFINE(type##_bitset, capacity)                              \
-    static type type##_storage[capacity];                               \
-    static type##Pool type##_pool = {                                   \
+    EWRAM_DATA static type type##_storage[capacity];                               \
+    EWRAM_DATA static type##Pool type##_pool = {                                   \
         .bitset = &type##_bitset,                                       \
         .objects = type##_storage,                                      \
     };                                                                  \


MeirGavish:
Oh, and the unit tests don't build with this change...

MathisMartin31:
I think tests don't build because iirc, `EWRAM_DATA` is a define from libtonc?
Maybe doing this at the top of the test source file could help, basically redefining it to do nothing:

```
#ifdef EWRAM_DATA
#undef EWRAM_DATA
#define EWRAM_DATA
#endif
```

MeirGavish:
> I think tests don't build because iirc, `EWRAM_DATA` is a define from libtonc? Maybe doing this at the top of the test source file could help, basically redefining it to do nothing:
> 
> ```
> #ifdef EWRAM_DATA
> #undef EWRAM_DATA
> #define EWRAM_DATA
> #endif
> ```

But doing that will undefined EWRAM_DATA and not put it in the section we want when building the ROM...
The issue can be fixed with the preprocessor, I just need to clear up my mind and think it up.

MathisMartin31:
> But doing that will undefined EWRAM_DATA and not put it in the section we want when building the ROM...
> The issue can be fixed with the preprocessor, I just need to clear up my mind and think it up.

I meant specifically for the test, which targets a regular x86 PC, not the ROM targeting the GBA

MeirGavish:
> > But doing that will undefined EWRAM_DATA and not put it in the section we want when building the ROM...
> > The issue can be fixed with the preprocessor, I just need to clear up my mind and think it up.
> 
> I meant specifically for the test, which targets a regular x86 PC, not the ROM targeting the GBA

Right, sorry, I missed that.
That would help there, but actually pool.h doesn't include tonc.h so it would need to get EWRAM_DATA from somewhere else...

MathisMartin31:
Also let's be weary of EWRAM, as it is less performant than IWRAM
https://gbadev.net/gbadoc/memory.html#external-work-ram

MeirGavish:
> Also let's be weary of EWRAM, as it is less performant than IWRAM https://gbadev.net/gbadoc/memory.html#external-work-ram

That's true, but it's larger. I think it makes more sense to put general purpose memory in EWRAM and reserve IWRAM for things that specifically need to be fast.

ricfehr3:
> > Also let's be weary of EWRAM, as it is less performant than IWRAM https://gbadev.net/gbadoc/memory.html#external-work-ram
> 
> That's true, but it's larger. I think it makes more sense to put general purpose memory in EWRAM and reserve IWRAM for things that specifically need to be fast.

I'm not by my computer right now. But one thing I had thought of previously is allowing a pool to be defined in any section of memory. Then we can add another set of macros or pass it as an argument or whatever works.

MathisMartin31:
> > > Also let's be weary of EWRAM, as it is less performant than IWRAM https://gbadev.net/gbadoc/memory.html#external-work-ram
> > 
> > 
> > That's true, but it's larger. I think it makes more sense to put general purpose memory in EWRAM and reserve IWRAM for things that specifically need to be fast.
> 
> I'm not by my computer right now. But one thing I had thought of previously is allowing a pool to be defined in any section of memory. Then we can add another set of macros or pass it as an argument or whatever works.

Yeah, for example I think all the graphics stuff should be in IWRAM, while generic ListNodes and abstract objects can reside in EWRAM.

But I don't know how mixing and matching this kind of stuff like that will turn out, because if we're iterating over a List to access SpriteObjects, will we be limited to the EWRAM speed in the end? And is transferring data from one to the other slower than moving data within each one?

But as Meir said, the main purpose for which we want to use EWRAM is because it's big. I'm worring over RAM speed and performance but we're not doing things crazy enough like 3D to warrant these concerns ><

MeirGavish:
> > > Also let's be weary of EWRAM, as it is less performant than IWRAM https://gbadev.net/gbadoc/memory.html#external-work-ram
> > 
> > 
> > That's true, but it's larger. I think it makes more sense to put general purpose memory in EWRAM and reserve IWRAM for things that specifically need to be fast.
> 
> I'm not by my computer right now. But one thing I had thought of previously is allowing a pool to be defined in any section of memory. Then we can add another set of macros or pass it as an argument or whatever works.

Yeah, makes sense. I'm not sure whether I will continue working on this PR and do that... Though I might depending on how often I encounter these crashes...

ricfehr3:
Since we will be covering this in #541, should we close this PR?

MeirGavish:
> Since we will be covering this in #541, should we close this PR?

I still want to give a deeper look to that in that PR but I think we can close this one

PR #562, opened 2026-06-14
Fixed mode argument on calls to obj_unhide

MeirGavish:
Apparently `obj_hide()` and `obj_unhide()` override the mode bits (regular, affine, etc.) of the sprite, because they are the same bits. i.e. a sprite mode can only be one of `ATTR0_REG`, `ATTR0_AFF`, `ATTR0_HIDE`, `ATTR0_AFF_DBL`. 
Therefore `obj_unhide` accepts a mode parameter to determine which mode to return the sprite to.
So far the code just passed 0 on all calls, but that's actually a magic number and it needs to be in line with the type of the sprite so I changed that in this PR.
A more correct way to do this, is to save the sprite mode in the `Sprite` struct but that crashes so I made a separate draft PR #563  for that and this is just a quick fix in the meantime.

### DIFF
--- a/source/game/blind_select.c
+++ b/source/game/blind_select.c
@@ -494,7 +494,7 @@ void game_blind_select_change_background(void)
 {
     for (int i = 0; i < NUM_BLINDS_PER_ANTE; i++)
     {
-        obj_unhide(blind_select_tokens[i]->obj, 0);
+        obj_unhide(blind_select_tokens[i]->obj, ATTR0_REG);
     }
 
     // Default y position for the blind select tokens. 12 is the amount of tiles the background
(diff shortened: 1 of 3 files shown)


MathisMartin31:
Yeah I realized that `obj_unhide` was messing with the mode when at some point I tried to hide SkipTags, but could not change their scaling and rotation after calling `obj_unhide(0)` lol.

Both your PRs bring very usefull changes because we won't need to worry about modes when manipulating existing sprites anymore

ricfehr3 (approved):
Nice, good catch. I don't see any other use of `obj_unhide` in the code, and this is pretty straight forward. Thanks! Merging.

PR #563, opened 2026-06-14
Add mode field to Sprite

### LINKED ISSUE #589: Change Sprite.idx into an s16
Since it indexes MAX_SPRITES, there's no reason for it to be larger. With alignment/padding this could conserve pool space and IWRAM space.
This will require changing the type from int in a few places in the code.

_Originally posted by @MeirGavish in https://github.com/GBALATRO/balatro-gba/pull/563#discussion_r3508678549_

MeirGavish:
See #562.

As said, this PR crashes, still not entirely sure why, but it seems like moving the pools to EWRAM helps?
See #560

Edit: I rebased on main and it seems to not crash now...? So I'm setting as ready for review.

This PR just adds a sprite hiding API to save the mode in the sprite instead of always calling `obj_hide` and needing to pass the correct mode, ensuring correctness and making it less error-prone.

### DIFF
--- a/include/sprite.h
+++ b/include/sprite.h
@@ -189,10 +189,17 @@ bool sprite_get_width(Sprite* sprite, int* width);
  */
 int sprite_get_pb(const Sprite* sprite);
 
-// TODO: Document
+/**
+ * @brief Hides the sprite by manipulating ATTR0_HIDE in OAM.
+ * @param sprite The sprite to hide
+ */
 void sprite_hide(Sprite* sprite);
 
-// TODO: Documents
+/**
+ * @brief Unhides the sprite by manipulating ATTR0_HIDE in OAM.
+ * The sprite's ATTR0_MODE is maintained from the sprite's creation with @ref sprite_new()
+ * @param sprite The sprite to unhide
+ */
 void sprite_unhide(Sprite* sprite);
 
 /**
@@ -236,8 +243,17 @@ void sprite_object_destroy(SpriteObject* sprite_object);
  */
 void sprite_object_set_sprite(SpriteObject* sprite_object, Sprite* sprite);
 
+/**
+ * @brief Hides the SpriteObject by manipulating ATTR0_HIDE in OAM.
+ * @param sprite_object The SpriteObject to hide
+ */
 void sprite_object_hide(SpriteObject* sprite_object);
 
+/**
+ * @brief Unhides the SpriteObject by manipulating ATTR0_HIDE in OAM.
+ * The sprite's ATTR0_MODE is maintained from the sprite's creation with @ref sprite_new()
+ * @param sprite_object The SpriteObject to unhide
+ */
 void sprite_object_unhide(SpriteObject* sprite_object);
 
 /**
--- a/include/sprite.h
+++ b/include/sprite.h
@@ -45,7 +45,7 @@ typedef struct
     /**
      * @brief Sprite index in memory managed by GBAlatro
      */
-    int idx;
+    s16 idx;
 
     /**
      * @brief The mode of the sprite (regular, affine, etc.), set when the sprite is created
@@ -127,7 +127,7 @@ typedef struct
  * @return Valid Sprite if allocations are successful.
  *         Otherwise, return **NULL**.
  */
-Sprite* sprite_new(u16 a0, u16 a1, u32 tid, u32 pb, int sprite_index);
+Sprite* sprite_new(u16 a0, u16 a1, u32 tid, u32 pb, s16 sprite_index);
 
 /**
  * @brief Destroy Sprite
@@ -143,7 +143,7 @@ void sprite_destroy(Sprite** sprite);
  *
  * @return Index of sprite in object buffer if `sprite` is valid, otherwise **UNDEFINED**.
  */
-int sprite_get_layer(Sprite* sprite);
+s16 sprite_get_layer(Sprite* sprite);
 
 /**
  * @brief Get a Sprite's width and height
--- a/include/sprite.h
+++ b/include/sprite.h
@@ -46,6 +46,12 @@ typedef struct
      * @brief Sprite index in memory managed by GBAlatro
      */
     int idx;
+
+    /**
+     * @brief The mode of the sprite (regular, affine, etc.), set when the sprite is created
+     * corresponds to A0 & ATTR0_MODE_MASK
+     */
+    u16 mode;
 } Sprite;
 
 /**
@@ -183,6 +189,12 @@ bool sprite_get_width(Sprite* sprite, int* width);
  */
 int sprite_get_pb(const Sprite* sprite);
 
+// TODO: Document
+void sprite_hide(Sprite* sprite);
+
+// TODO: Documents
+void sprite_unhide(Sprite* sprite);
+
 /**
  * @brief Initialize GBAlatro sprite system
  */
@@ -223,6 +235,10 @@ void sprite_object_destroy(SpriteObject* sprite_object);
  *                      Cannot be **NULL**.
  */
 void sprite_object_set_sprite(SpriteObject* sprite_object, Sprite* sprite);
+
+void sprite_object_hide(SpriteObject* sprite_object);
+
+void sprite_object_unhide(SpriteObject* sprite_object);
 
 /**
  * @brief Reset SpriteObject's transform back to default values.
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -165,19 +165,19 @@ void sprite_draw()
 
 int sprite_get_pb(const Sprite* sprite)
 {
-    CHECK_NULL_ARG_RET(sprite, UNDEFINED);
+    GBAL_RETURN_IF_NULL_RET(sprite, UNDEFINED);
     return (sprite->obj->attr2 & ATTR2_PALBANK_MASK) >> ATTR2_PALBANK_SHIFT;
 }
 
 void sprite_hide(Sprite* sprite)
 {
-    CHECK_NULL_ARG_VOID(sprite);
+    GBAL_RETURN_IF_NULL_VOID(sprite);
     obj_hide(sprite->obj);
 }
 
 void sprite_unhide(Sprite* sprite)
 {
-    CHECK_NULL_ARG_VOID(sprite);
+    GBAL_RETURN_IF_NULL_VOID(sprite);
     obj_unhide(sprite->obj, sprite->mode);
 }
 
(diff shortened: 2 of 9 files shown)


MathisMartin31:
~What's the crash exactly here?~
I pulled your branch, looks like it's in the exact same place I was crashing in #553, and yes adding `EWRAM_DATA` to the pool structs declarations does resolve the crash.
It looks more and more like we are indeed running out of IWRAM memory

MathisMartin31:
I just did the math for this, and doing `sizeof(Sprite) * MAX_SPRITES + ...` yields 5448 bytes, or ~43.6Kb, so the memory pool alone more than fills the entire IWRAM

MeirGavish:
> I just did the math for this, and doing `sizeof(Sprite) * MAX_SPRITES + ...` yields 5448 bytes, or ~43.6Kb, so the memory pool alone more than fills the entire IWRAM

In fact we have a really nice script `get_memory_map.sh` to print the pools and their sizes which is run on every PR as part of the Build ROM workflow.

Here's the output from #562 which doesn't change the pools.
```
Sprite          | 0x030032ac | 2560       | 40         | 32         
SpriteObject    | 0x03002f0c | 896        | 40         | 32         
Joker           | 0x030026a4 | 96         | 40         | 32         
JokerObject     | 0x03002704 | 64         | 36         | 32         
Card            | 0x03002744 | 104        | 36         | 32         
CardObject      | 0x03002dcc | 192        | 56         | 32         
ListNode        | 0x030027ac | 1536       | 40         | 32         
Total bytes used: 5960
```

So we can see the pools are mapped to IWRAM and indeed the pool sizes themselves add up to 5448 bytes, but that's actually ~5.3KiB. Along with the bitsets it's about 5.5KiB which is a big chunk of IWRAM but still not everything.

The script uses readelf but apparently we can easily check how much IWRAM is used with a size utility.
On the main branch,
```
arm-none-eabi-size -A ./build/balatro-gba.elf
./build/balatro-gba.elf  :
section                size        addr
.crt0                   528   134217728
.init                    12   134218256
.plt                      0   134218268
.text                132968   134218272
.fini                    12   134351240
.rodata             4901212   134351256
.version                 26   139252468
.ctors                    2   139252494
.dtors                    0   139252496
.eh_frame                 4   139252496
.gcc_except_table         0   139252500
.iwram                 6240    50331648
.bss                  13604    50337888
.data                  8628    50351492
.init_array               4    50360120
.fini_array               4    50360124
.iwram0                   0    50360128
.iwram1                   0    50360128
.iwram2                   0    50360128

...

Total               5579681
```

The IWRAM sections here add up to about 28,472 bytes ~ 27.8KiB, which is most of the IWRAM.

For this PR,
```
./build/balatro-gba.elf  :
section                size        addr
.crt0                   528   134217728
.init                    12   134218256
.plt                      0   134218268
.text                132928   134218272
.fini                    12   134351200
.rodata             4901212   134351216
.version                 26   139252428
.ctors                    2   139252454
.dtors                    0   139252456
.eh_frame                 4   139252456
.gcc_except_table         0   139252460
.iwram                 6240    50331648
.bss                  14124    50337888
.data                  8628    50352012
.init_array               4    50360640
.fini_array               4    50360644
.iwram0                   0    50360648
.iwram1                   0    50360648
.iwram2                   0    50360648
...

Total               5579880
```

Here it adds up to 28.3KiB so it's still not everything but still most of it.

I'm not sure if the crashes are being triggered by being close to filling up the IWRAM as opposed to actually filling it but we do need to clear up some IWRAM.

I'll open a new issue for this.

MathisMartin31:
Huh, I was sure I read 32 kilobits and not kilobytes in the docs, must have been real tired ha ha >_<

MeirGavish:
I rebased on main and it seems to not crash now...? So I'm setting as ready for review.
Though we do need to look into these memory issues...

source/sprite.c
```diff
@@ -161,13 +164,22 @@ void sprite_draw()
...
-    }
+    GBAL_RETURN_IF_NULL_RET(sprite, UNDEFINED);
     return (sprite->obj->attr2 & ATTR2_PALBANK_MASK) >> ATTR2_PALBANK_SHIFT;
 }
 
+void sprite_hide(Sprite* sprite)
+{
+    GBAL_RETURN_IF_NULL_VOID(sprite);
+    obj_hide(sprite->obj);
+}
```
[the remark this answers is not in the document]
MeirGavish: Well, the intent is to keep the mode from creation and not allow changing the sprite mode during its lifetime that's also how it's documented... 
Though maybe we should allow changing the mode of the sprite? I don't know if I see a reason to...

include/sprite.h
```diff
@@ -46,6 +46,12 @@ typedef struct
      * @brief Sprite index in memory managed by GBAlatro
      */
     int idx;
```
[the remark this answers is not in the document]
MeirGavish: That is an interesting point, this is probably a good idea. But it's a bigger refactor than this, there are several places that use layer/index as an int and they all need to be replaced.
Also we will need to use `s16` and not `u16` to allow `UNDEFINED`.

source/sprite.c
```diff
@@ -161,13 +164,22 @@ void sprite_draw()
 
 int sprite_get_pb(const Sprite* sprite)
 {
-    if (sprite == NULL)
-    {
-        return UNDEFINED;
-    }
+    GBAL_RETURN_IF_NULL_RET(sprite, UNDEFINED);
     return (sprite->obj->attr2 & ATTR2_PALBANK_MASK) >> ATTR2_PALBANK_SHIFT;
 }
```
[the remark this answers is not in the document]
MeirGavish: It's true that many other functions guard `sprite->obj` but not all places do that and now I wonder if they should... `sprite_object->sprite` can definitely be NULL, but not sure about `sprite->obj`
I will think about it.

source/sprite.c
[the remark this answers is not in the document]
MathisMartin31: I also don't see a reason to do that either. Some sprites are affine, some don't need to be and if the mode changes, to me that means the sprite should be recreated.

source/sprite.c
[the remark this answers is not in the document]
MeirGavish: I think it's okay to not guard it in general since it's set by the constructor...

include/sprite.h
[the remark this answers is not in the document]
MeirGavish: Done.

include/sprite.h
```diff
@@ -183,6 +189,19 @@ bool sprite_get_width(Sprite* sprite, int* width);
  */
 int sprite_get_pb(const Sprite* sprite);
```
ricfehr3, line 190:
```suggestion
bool sprite_get_dimensions(Sprite* sprite, s32* width, s32* height);

/**
* @brief Get a Sprites's height
*
* @param sprite pointer to Sprite, cannot be **NULL**
* @param height pointer to variable to be set, cannot be **NULL**
*
* @return **true** is successful, **false** if otherwise. Upon success,
*         `height` contains valid data, otherwise, the variable is unchanged.
*/
bool sprite_get_height(Sprite* sprite, s32* height);

/**
* @brief Get a Sprite's width
*
* @param sprite pointer to Sprite, cannot be **NULL**
* @param width pointer to variable to be set, cannot be **NULL**
*
* @return **true** is successful, **false** if otherwise. Upon success,
*         `width` contains valid data, otherwise, the variable is unchanged.
*/
bool sprite_get_width(Sprite* sprite, s32* width);

/**
* @brief Get the palette bank of a Sprite
*
* @param sprite pointer to extract associated palette bank. Cannot be **NULL**.
*
* @return The palette bank of the Sprite if successful, otherwise return **UNDEFINED**.
*/
s32 sprite_get_pb(const Sprite* sprite);
```

Not a needed change, but for consistency, we should consider having the same types (matching s16). Really though, this is just a nitpick. These eventually get called from functions that use `int` type as well (obj_get_width/height). I'll let you have the final say since it's your work @MeirGavish .

source/sprite.c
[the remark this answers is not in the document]
ricfehr3: IMHO, This is one of those times where putting faith in the programmer makes more sense than any number of checks. I agree, if it's in the constructor, it should be good. Additionally, the tonc functions don't check for NULL pointers, which is a fair baseline. 

My only comment is on consistency. See cross-referenced comment [here](https://github.com/GBALATRO/balatro-gba/pull/563/changes#r3619493019)

source/sprite.c
ricfehr3, line 151: These functions should switch to the standard `GBAL_RETURN_IF_NULL_RET` and similar macros. In addition, here we do NULL checks on `sprite->obj`. I don't think this is necessary as discussed [here](https://github.com/GBALATRO/balatro-gba/pull/563/changes#r3508614870). But we should pick one or the other for consistency.

ricfehr3 (changes requested):
@MeirGavish Finally came around to reviewing this, sorry about the delay. I've just been busy, unfortunately :/ . But this is good. I like seeing your changes come together from the previous PRs with the macros and inheritance. I have a couple comments on that, just looking to use the Macros more and some questions on consistency.

source/sprite.c
MeirGavish: Yeah, you're right, it makes sense to add the proper NULL-check macros throughout.

include/sprite.h
MeirGavish: I think I prefer to leave these as `int` since they call `obj_get_width/height()` which also return `int` to be consistent with that type... The weird thing is that `oam_sizes` and the return types of `obj_get_size()` are `u8`, so there's a bit of an inconsistency on tonc's side here...

source/sprite.c
MeirGavish: I've added comprehensive NULL-checks

MeirGavish:
> @MeirGavish Finally came around to reviewing this, sorry about the delay. I've just been busy, unfortunately :/ . But this is good. I like seeing your changes come together from the previous PRs with the macros and inheritance. I have a couple comments on that, just looking to use the Macros more and some questions on consistency.

Thanks! No problem, I've been busy myself lately. I addressed the review and requesting another.

source/sprite.c
MeirGavish: I'm starting to think maybe it isn't the best idea to have NULL-checks on every level of the call hierarchy.
Maybe have a NULL-check only for pointer variables that the function actually accesses and if they are only passed to another function, trust it to perform the check, that way only the lowest level functions perform the check.
The benefit would be performance and less clutter in the code. 
The downside would be that the logs will show only the bottom of the call stack which is a bit delayed from when the error first occurred.

ricfehr3: I think we should keep them in for now. As you said, they are good for safety. I also think we shouldn't worry about the optimization on the checks (yet). Since we know most of the time we are under the vsync time limit besides some odd spikes.

That being said, we can still be smart with it going forward and treat every situation uniquely. It's easier to just have the checks there, just in case. But ideally, it would be as efficient as possible. It's a tradeoff we will need to review each time.

ricfehr3 (approved):
@MeirGavish Looks good! Feel free to merge when you are ready. I left one more comment on the NULL check usage, but it's not blocking for this review. Thanks for the PR again!

source/sprite.c
MeirGavish: Yeah, you're right, giving another look at the code, it seems like it's better to fail early in probably all cases.

MeirGavish:
> @MeirGavish Looks good! Feel free to merge when you are ready. I left one more comment on the NULL check usage, but it's not blocking for this review. Thanks for the PR again!

Thanks! Gave it a test run and it seems fine, merging

PR #570, opened 2026-06-18
[Refactor] Separate Round logic into its own file

### LINKED ISSUE #532: Extract round code from game.c into round.c/.h
Opening a sub-issue to record this is part of #465.
As per discussion #518, this could be problematic to do separately.

MathisMartin31:
Closes #532

~Still a draft until I properly clean, format and document as much as possible~
Good to go ^^


include/game/round.h
```diff
@@ -0,0 +1,63 @@
...
+ *         current Blind's score requirement, and applies the flaming effect if needed.
+ */
+void toggle_flaming_score(void);
+
+/**
+ * @brief Get the index of the last played card
+ *
+ * @return int
+ */
+int get_played_top(void);
```
ricfehr3, line 22: My problem I was having was overengineering the removal/restructuring of the stacks. This is far more sane, just keep it as is, nice work!

source/game/round.c
```diff
@@ -0,0 +1,2040 @@
...
+    PLAY_STARTING,
+    PLAY_BEFORE_SCORING,
+    PLAY_SCORING_CARDS,
+    PLAY_SCORING_CARD_JOKERS,
+    PLAY_SCORING_HELD_CARDS,
+    PLAY_SCORING_INDEPENDENT_JOKERS,
+    PLAY_SCORING_HAND_SCORED_END,
+    PLAY_ENDING,
+    PLAY_ENDED
+};
```
ricfehr3, line 199: I want to avoid additional refactor work in this PR, but this should be moved to a state machine eventually. It's better to have the transitions handled in a central way. Again, this was something I got stuck on, and keeping it this way makes more sense for now. I will make a issue for this though. I tried to do this previously, but it was getting complicated, this PR should make it easier.

ricfehr3:
Cool! I'm in the middle of reviewing it right now. Just wanted to say, nice work. Seeing the leftover `game.c` and having a clearer view of each files responsibilities helps a lot for future design. We definitely will need to simplify the new `round.c` file, but the decoupling here is good. Nice work!

MeirGavish:
I will try to give this a review, probably not a full review, more of an overview

source/game.c
```diff
@@ -412,7 +186,7 @@ void game_init()
     g_game_vars.ante = STARTING_ANTE;
     g_game_vars.money = STARTING_MONEY;
     g_game_vars.score = STARTING_SCORE;
-    g_game_vars.round = 0;
+    g_game_vars.round = STARTING_ROUND;
```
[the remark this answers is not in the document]
MathisMartin31: I'm sure it's reset somewhere else but it doesn't hurt to add it here as well

include/game/round.h
```diff
@@ -0,0 +1,63 @@
+/**
+ * @file round.h
+ * @brief API relative to the Rounds we play
+ */
+
+#ifndef GAME_ROUND_H
+#define GAME_ROUND_H
+
```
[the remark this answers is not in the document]
MathisMartin31: True

source/game/round.c
```diff
@@ -0,0 +1,2040 @@
...
+
+                set_hand_top(get_hand_top() - 1);
+                // This technically isn't drawing cards, I'm just reusing the variable
+                cards_drawn++;
+                sound_played = false;
+                g_game_vars.timer = TM_ZERO;
+
+                *hand_y = hand[card_idx]->sprite_object->y;
+                *hand_x = hand[card_idx]->sprite_object->x;
+            }
```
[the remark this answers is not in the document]
MathisMartin31: I think this works because immediately after calling `card_object_destroy`, `hand[card_idx]` is indeed NULL, but calling `reorder_card_sprites_layers()` pushes that NULL value at the end of the array, and the value at index `card_idx` is no longer NULL.

There is still an edge case though, which is when `card_idx == hand_top`. In that case, there are no non-NULL cards after it and so it will stay in place, which in turn means that `hand[card_idx]` is still NULL.

Trying to fix the code as Copilot suggested results in weird sprite jitter so more debugging is required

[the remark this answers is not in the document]
MathisMartin31: Well, doing this to protect again the edge case I saw seems to work xD
I don't know why it was working before but now we're sure it won't crash on us at some point

```
if (hand[card_idx] != NULL)
{
    *hand_y = hand[card_idx]->sprite_object->y;
    *hand_x = hand[card_idx]->sprite_object->x;
}
```

source/game/round.c
```diff
@@ -0,0 +1,2040 @@
...
+            // 0, 0);
+            // Set the sprite for the discarded card object
+            card_object_set_sprite(discarded_card_object, 0);
+            sprite_object_reset_transform(discarded_card_object->sprite_object);
+
+            discarded_card_object->sprite_object->tx = int2fx(204);
+            discarded_card_object->sprite_object->ty = int2fx(112);
+            discarded_card_object->sprite_object->x = int2fx(240);
+            discarded_card_object->sprite_object->y = int2fx(80);
+        }
```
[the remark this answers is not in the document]
MathisMartin31: I changed the condition to undiscard a card to `discard_top >= 0 && discarded_card_object == NULL` and it doesn't seem to affect the round end "undiscard" animation.

source/game/round.c
```diff
@@ -0,0 +1,2040 @@
...
+    }
+
+    return false;
+}
+
+static inline void play_starting_played_cards_update(int played_idx)
+{
+    bool card_selected = card_object_is_selected(played[played_top - scored_card_index]);
+    if (played_idx == played_top && (g_game_vars.timer % FRAMES(10) == 0 || !card_selected) &&
+        g_game_vars.timer > FRAMES(40))
```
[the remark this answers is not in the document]
MathisMartin31: Somehow the game doesn't crash when reading out of bounds from the `played` array but this retains the functionality while protecting from invalid reads:

```
bool card_selected = (played_top < scored_card_index)
    ? false
    : card_object_is_selected(played[played_top - scored_card_index]);
```

From the few tests I did, we do need to go inside this function with `scored_card_index` starting at `played_top+1` so that when we start, no card is selected after we move then to the center of the screen, then we select the scoring cards one by one.

source/game/round.c
```diff
@@ -0,0 +1,2040 @@
...
+ * @brief This is the reverse of PLAY_STARTING. The cards get reset back to their neutral position
+ *         sequentially.
+ *
+ * @param played_idx index of the card currently considered
+ */
+static inline void play_ending_played_cards_update(int played_idx)
+{
+    bool card_selected = card_object_is_selected(played[played_top - scored_card_index]);
+    if (played_idx == played_top && (g_game_vars.timer % FRAMES(10) == 0 || !card_selected) &&
+        g_game_vars.timer > FRAMES(40))
```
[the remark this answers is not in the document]
MathisMartin31: Same out of bounds, same fix ^^

source/game/round.c
```diff
@@ -0,0 +1,2040 @@
...
+
+        for (int i = 0; i <= played_top; i++)
+        {
+            if (played[i]->card->rank != played[(i + 1) % played_top]->card->rank &&
+                played[i]->card->rank != played[(i + 2) % played_top]->card->rank)
+            {
+                unmatched_index = i;
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
MathisMartin31: I have played a 4OAK with a few combinations of 4 and 5 five cards played (no extra card, or an extra card at the beginning, end, or somewhere in the middle) and have encountered no issue whatsoever.

I guess Copilot didn't understand what's going on here?

Although there still is one thing to fix here: to go to the code Copilot is talking about, the condition is `played_top >= 3` with a comment saying "5 cards played", but when we play 5 cards, the index of the last one/top of the stack is 4, not 3.

MathisMartin31:
@MeirGavish I have addressed all comments from Copilot's review, feel free to review my fixes and resolve the conversations if you find them satisfactory ^^

MathisMartin31:
I just thought about it @ricfehr3, but I'm touching things that are already included in your PR #516. Should I mark this PR as blocked by yours, or yours blocked by mine?

ricfehr3:
> I just thought about it @ricfehr3, but I'm touching things that are already included in your PR #516. Should I mark this PR as blocked by yours, or yours blocked by mine?

Let's block #516 on this PR. There will be conflicts and I think it makes more sense to put the smaller change on top of the larger one in that case.

source/game.c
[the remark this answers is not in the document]
MeirGavish: Seems like they only reset when calculating the score for the played hand...
I think it's good to initialize them here as well.

source/game/round.c
```diff
@@ -749,17 +750,24 @@ static inline void card_in_hand_loop_handle_discard_and_shuffling(
...
+
+                // Remove discarded card from hand and shift the ones after it
                 card_object_destroy(&hand[card_idx]);
                 reorder_card_sprites_layers();
-
                 set_hand_top(get_hand_top() - 1);
-                // This technically isn't drawing cards, I'm just reusing the variable
-                cards_drawn++;
+
+                cards_discarded++;
```
MeirGavish, line 792: This was much needed (referring to the removal of the "reuse" of `cards_drawn`)

source/game/round.c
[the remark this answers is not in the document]
MeirGavish: Copilot is right here though, there is no case that this code fails because but it really is incorrect. The mix of size and top really makes things confusing and it's probably better to only use size and not use top. I will make another suggestion for a fix [here](https://github.com/GBALATRO/balatro-gba/pull/570/changes#r3468450056).

source/game/round.c
```diff
@@ -0,0 +1,2039 @@
...
+        }
+    }
+    else // If there are only 4 cards selected we know they match
+    {
+        for (int i = 0; i <= played_top; i++)
+        {
+            card_object_set_selected(played[i], true);
+        }
+    }
+}
```
MeirGavish, line 940: Referring to [my reply](https://github.com/GBALATRO/balatro-gba/pull/570/changes#r3468380084),
This only uses size and never uses top, I tested it a bit and it should work...
You need to define
```c
int get_played_size(void)
{
    return played_top + 1;
}
```
I opened issue #573 to do this for all the code.

```suggestion
static inline void select_four_of_a_kind_cards_in_played_hand(void)
{
    int played_size = get_played_size();
    // find four cards with the same rank
    // If there are 5 cards selected we just need to find the one card that doesn't match, and
    // select the others
    if (played_size >= 5)
    {
        int unmatched_index = -1;
        for (int i = 0; i < played_size; i++)
        {
            if (played[i]->card->rank != played[(i + 1) % played_size]->card->rank &&
                played[i]->card->rank != played[(i + 2) % played_size]->card->rank)
            {
                unmatched_index = i;
                break;
            }
        }

        for (int i = 0; i < played_size; i++)
        {
            if (i != unmatched_index)
            {
                card_object_set_selected(played[i], true);
            }
        }
    }
    else // If there are only 4 cards selected we know they match
    {
        for (int i = 0; i < played_size; i++)
        {
            card_object_set_selected(played[i], true);
        }
    }
}
```

MathisMartin31: Done ^^.

I also took the liberty to change `played_top + 1` to `played_size` in a couple of other places in the code

source/game/round.c
MathisMartin31: Yeah, I figured the compiler was smart enough to determine if a variable/register can be reused :)

include/hand.h
```diff
@@ -138,6 +138,20 @@ int get_hand_top(void);
...
+ * @return true if yes, false otherwise
+ */
+bool can_discard_hand(void);
+
+/**
+ * @brief Determines whether the "Play" button can be pressed or not
+ *
+ * @return true if yes, false otherwise
+ */
+bool can_play_hand(void);
```
MeirGavish: I think these functions should be in round.c, they are only specifically relevant to the round and aren't really generic to the hand despite the term "hand" appearing in them.

MeirGavish (changes requested):
Looks great, good job! Requesting changes just for the one comment about the functions.

Other than that I'm debating whether `chips`, `mult`, and along with them `joker_object_score()` and `score` should be in round.c rather than in `g_game_vars` because they are specific to the round but I think this should be in another issue/PR and left as out of scope for this one.

MathisMartin31:
Thanks for the review ^^

> Other than that I'm debating whether `chips`, `mult`, and along with them `joker_object_score()` and `score` should be in round.c rather than in `g_game_vars` because they are specific to the round but I think this should be in another issue/PR and left as out of scope for this one.

They aren't entirely exclusive to the round, planet cards and celestial packs will use them too iirc

MeirGavish:
> Thanks for the review ^^
> 
> > Other than that I'm debating whether `chips`, `mult`, and along with them `joker_object_score()` and `score` should be in round.c rather than in `g_game_vars` because they are specific to the round but I think this should be in another issue/PR and left as out of scope for this one.
> 
> They aren't entirely exclusive to the round, planet cards and celestial packs will use them too iirc
> 

It's true but I thought to maybe separate the display functions `display_chips(u32 chips)` etc. that's global and could be used by planet cards and `display_curr_chips(void)`etc. that's round exclusive... (Though `display_curr_round()` would be global in game.c/.h).

This is for another issue.

include/hand.h
MathisMartin31: Yeah you're right ^^

MathisMartin31:
> > Thanks for the review ^^
> > > Other than that I'm debating whether `chips`, `mult`, and along with them `joker_object_score()` and `score` should be in round.c rather than in `g_game_vars` because they are specific to the round but I think this should be in another issue/PR and left as out of scope for this one.
> > 
> > 
> > They aren't entirely exclusive to the round, planet cards and celestial packs will use them too iirc
> 
> It's true but I thought to maybe separate the display functions `display_chips(u32 chips)` etc. that's global and could be used by planet cards and `display_curr_chips(void)`etc. that's round exclusive... (Though `display_curr_round()` would be global in game.c/.h).
> 
> This is for another issue.

Oh yeah, so basically doing what we had before, but when you put it like that it makes sense. Also agree that this could be its own small issue.

ricfehr3 (approved):
Hey, sorry for just getting around to this. I've been keeping track of it and I agree with all the changes caught so far. Nice work on this! I have a couple of comments, but they are not changes. This is a another great refactor.

source/game/round.c
MathisMartin31: Yeah I saw the associated issue but I didn't feel like opening that can of worms right now ><

MeirGavish (approved):
Great! Approving!
Edit: I didn't look at **all** the code in this PR, but I think the reviews were comprehensive enough?

include/game/round.h
MeirGavish: There's still room to change the stack implementation but this and especially `get_played_size()` is a good first step that decouples the implementation a bit.

MeirGavish:
> Other than that I'm debating whether `chips`, `mult`, and along with them `joker_object_score()` and `score` should be in round.c rather than in `g_game_vars` because they are specific to the round but I think this should be in another issue/PR and left as out of scope for this one.

On second thought, I think they can stay in the globals, it's consistent with the rest of the values and it seems to be how Balatro does it.
But I did open #574 for other related issues mentioned.

ricfehr3:
> Great! Approving! Edit: I didn't look at **all** the code in this PR, but I think the reviews were comprehensive enough?

I think we are good to go. I have been through it a few times and a lot of it is the same code carried over from `game.c`. Looks like it's time to merge. Thanks @MathisMartin31 and @MeirGavish for the PR and review + making the issue.

include/game/round.h
ricfehr3: Agreed, the stack implementation needs some work.

PR #580, opened 2026-07-02
[Refactor] Moved `joker_object_score()` to round.c

### LINKED ISSUE #592: Refactor set_and_shift_text()
This function is a little weird. I thought it was very specific to joker scoring, and indeed the last 2 lines are very scoring-specific. I guess this function should be changed a bit to be more clear, at the very least renamed and documented.
Then I guess it can be used for card scoring as well and be shared, I guess common_ui.h would be most appropriate.
Though maybe the function should be changed more than that... it does 2 things and it's a little difficult to understand what these are despite its brevity.
Set what text? What sort of shift? It actually displays the string on the screen, that's not implied from the name. `cursor_pos_x` is both an input and an output.

 _Originally posted by @MeirGavish in [#536](https://github.com/GBALATRO/balatro-gba/pull/536/changes/01a437662f92a493e17ea6d69d73e8c8ca742c11..af70fc984705b4e2fdc49637b0e9c887a5bb7bdd#r3549253734)_

MeirGavish:
`joker_object_score()` does a lot of very round-specific scoring rather than joker-generic functionality so I think it should be in round.c

### DIFF
--- a/source/game/round.c
+++ b/source/game/round.c
@@ -53,10 +53,13 @@
 #define GAME_PLAYING_HAND_SEL_Y 1
 
 // Pixel sizes
-#define CARD_FOCUSED_UNSEL_Y 10
-#define CARD_UNFOCUSED_SEL_Y 15
-#define CARD_FOCUSED_SEL_Y   20
-#define SCORED_CARD_TEXT_Y   48
+#define CARD_FOCUSED_UNSEL_Y   10
+#define CARD_UNFOCUSED_SEL_Y   15
+#define CARD_FOCUSED_SEL_Y     20
+#define SCORED_CARD_TEXT_Y     48
+#define JOKER_SCORE_TEXT_Y     48
+#define HELD_CARD_SCORE_TEXT_Y 108
+#define MAX_CARD_SCORE_STR_LEN 2
 
 // clang-format off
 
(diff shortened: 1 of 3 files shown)


source/game/round.c
```diff
@@ -1472,6 +1475,130 @@ static inline void game_round_process_flaming_score(void)
...
+static bool joker_object_score(
+    JokerObject* joker_object,
+    CardObject* card_object,
+    enum JokerEvent joker_event
+)
+{
+    if (joker_object == NULL)
+    {
+        return false;
+    }
```
ricfehr3: Opportunity to use new macros: `GBAL_RETURN_IF_NULL_RET`, also we need to check for `card_object` as well.

source/game/round.c
```diff
@@ -1472,6 +1475,130 @@ static inline void game_round_process_flaming_score(void)
  * CARD/JOKER SCORING LOGIC
  ******************************************************************************/
 
+static void joker_scoring_set_and_shift_text(
+    char* str,
+    int* cursor_pos_x,
+    int* cursor_pos_y,
```
ricfehr3: Can you add a NULL check for these? Or... since it's just in this file, at least note they cannot be NULL. Though, I think it would make sense to just check for NULL here, since this won't be called every frame.

ricfehr3 (dismissed):
Thanks @MeirGavish . This is a good refactor. I have a couple comments addressing some existing things in the code. Nothing major, and once you're happy with it feel free to merge, or let me know if you want another look. Thanks!

source/game/round.c
MeirGavish: Oh dear, this is #592, I may resort to cleaning things up a bit here...

MeirGavish: I addressed the issue, so I'm linking it. I extracted the text printing logic from `set_and_shift_text()` and replaced it with a single call to `tte_printf()` instead. I replaced `set_and_shift_text()` with the function `joker_scoring_increment_text_cursor()` that does only one thing.

source/game/round.c
```diff
@@ -1473,6 +1476,141 @@ static inline void game_round_process_flaming_score(void)
...
+    {
+        return false;
+    }
+
+    if (effect_flags_ret & JOKER_EFFECT_FLAG_RETRIGGER)
+    {
+        set_retrigger(joker_effect->retrigger);
+    }
+
+    // joker_effect.message will have been set if the Joker had anything custom to say
```
MathisMartin31: Could you remove this comment? It is already present at line 1586, where it ought to be with the handling of the message

MathisMartin31 (approved):
Looks great, thanks ^^. Just one small nitpick but I can't see any other issue so I'll approve, but we'll see what Rickey has to say about it :)

source/game/round.c
MeirGavish: Right, good call

MeirGavish:
```suggestion
```
Remove the comment

MeirGavish:
@ricfehr3 I want to merge this PR soon in case you still want to get around to re-review it beforehand.

PR #587, opened 2026-07-07
Added bitset_set_all()

### LINKED ISSUE #586: Add `bitset_set_all()` to the bitset
`reset_shop_jokers()` uses `bitset_clear()` and then does a for loop to set all the bits in the bitset.
This should be moved to the bitset function itself as `bitset_set_all()` or similar. Then it can just set the raw values to 0xFFFFFFFF and avoid this loop.

_Originally posted by @ricfehr3 in https://github.com/GBALATRO/balatro-gba/pull/536#discussion_r3533149529_

MeirGavish:
Closes #586

### DIFF
--- a/include/bitset.h
+++ b/include/bitset.h
@@ -128,6 +128,13 @@ int bitset_set_next_free_idx(Bitset* bitset);
 void bitset_clear(Bitset* bitset);
 
 /**
+ * @brief Set the bitset, all to 1
+ *
+ * @param bitset A @ref Bitset to operate on
+ */
+void bitset_set_all(Bitset* bitset);
+
+/**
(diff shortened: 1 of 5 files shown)


MeirGavish:
Reproduced the bug, I'll try to figure this out.
image
image

MeirGavish:
So the problem is that the bitset has a fixed size and it's larger than the joker registry, when you use `bitset_set_all()` and all the possible bits are set, the following function returns a number bigger than the joker registry and invalid joker IDs can be rolled.
```c
static inline int get_num_shop_jokers_avail(void)
{
    return bitset_num_set_bits(&s_avail_jokers_bitset);
}
```

PR #597, opened 2026-07-23
Removed call to `obj_aff_copy()` from `sprite_draw()`

MeirGavish:
Since `obj_aff_buffer` just points to `obj_buffer` ([with interleaved content](https://gbadev.net/tonc/regobj.html#sec-oam)), the call to `oam_copy()` in `sprite_draw()` copies both the regular sprite data and the affine data, overriding the call to `obj_aff_copy()` and making it redundant.
So I removed that call to improve performance a little bit.

### DIFF
--- a/source/sprite.c
+++ b/source/sprite.c
@@ -157,7 +157,6 @@ void sprite_init()
 
 void sprite_draw()
 {
-    obj_aff_copy(obj_aff_mem, obj_aff_buffer, MAX_AFFINES);
     oam_copy(oam_mem, obj_buffer, MAX_SPRITES);
 }
 


MathisMartin31 (approved):
I checked and `oam_copy` does indeed copy the whole buffer byte per byte so we don't need `obj_aff_copy`. Nice catch ^^

ricfehr3 (approved):
I agree with @MathisMartin31 and your research. The game seems to play without any issues as well. I'm going to approve and merge this one since it is only one line, though, we should totally check to see if there's a performance improvement here.

ricfehr3:
Thanks! @MeirGavish

MeirGavish:
> I agree with @MathisMartin31 and your research. The game seems to play without any issues as well. I'm going to approve and merge this one since it is only one line, though, we should totally check to see if there's a performance improvement here. 

True, I didn't get a chance to check

PR #599, opened 2026-07-24
[Bug] Implement independent RNG sequences to avoid interferences

### LINKED ISSUE #588: [BUG] Using the same seed does not produce the same Jokers in the shop
While using the same seed on separate runs does draw the same cards on the first round, the jokers in the shop are different.
This is due to the fact the RNG system is used to randomize sound effects and the same one is used for all random numbers, so just drawing or selecting cards will affect the jokers in the shop.
There are 2 possible solutions to this,
1. Use a pre-calculated LUT of random numbers for the sound effects and leave the RNG system to only handle gameplay-related events such as Jokers in the shop and deck shuffling. This is the easier solution and it's a decent one but the only downside is that there is still dependence between gameplay events - that is, rerolling the jokers in the shop will affect the deck shuffling in the next round.
2. Use a per-key RNG system that would separate the sequence of random numbers for the sound effects, jokers, deck shuffling, and anything else. This is the complex solution and it's what the original Balatro does [as explained in this video](https://youtu.be/54w9crNNThU&t=739).

If the issue is resolved in method 1, we can open a separate issue for method 2.

_Originally posted by @MeirGavish in https://github.com/GBALATRO/balatro-gba/pull/548#pullrequestreview-4610651695_

MathisMartin31:
Closes #588

I replaced almost all calls to the glibc's RNG to use my own so that we can easily manage several, independent RNG sequences. The default RNG is only used when resetting the seed: we call `srand` on it, and generate the starting states of our RNG using `rand`.

Also, if we could wait for the Skip Tags to be merged before merging this It would be great, I'd rather rebase this on main than the skip tags ><'

### DIFF
--- a/include/random.h
+++ b/include/random.h
@@ -12,6 +12,17 @@
 
 #include <tonc.h>
 
+enum RngType
+{
+    RNG_TYPE_CARD_SHUFFLE,
+    RNG_TYPE_BLIND,
+    RNG_TYPE_SHOP_ITEMS,
+    RNG_TYPE_SKIP_TAGS,
+    RNG_TYPE_JOKER_EFFECTS,
+    RNG_TYPE_MISC,
+    RNG_TYPE_MAX
+};
+
 /**
  * @brief Information to track and restore RNG state
  */
@@ -19,8 +30,8 @@ typedef struct
 {
     /** Initial seed */
     u32 seed;
-    /** Position in the rng sequence. */
-    u32 step;
+    /** Individual states for independent rng sequences */
+    u32 states[RNG_TYPE_MAX];
 } RngInfo;
 
 /**
@@ -52,15 +63,20 @@ void rng_set_seed(u32 seed);
 void rng_shuffle_seed(void);
 
 /**
- * @brief Get the next "randomly" generated number in the sequence from the current seed.
+ * @brief Get the next "randomly" generated number in the sequence corresponding to the given type.
+ *
+ * @note Custom rng had to be implemented to be able to manage several independent sequences, since
+ *        `initstate` and `setstate` are POSIX and not available on GBA via devkitpro.
+ *
+ * @param type rng key to the sequence we need to pull a random number from
  *
  * @return u32
  */
-u32 rng_get_u32(void);
+u32 rng_get_u32(enum RngType type);
 
 /**
- * @brief Restore RNG info struct in the GameVariables. Sets the `seed` and seeks the
- *         position `step` in the rng sequence.
+ * @brief Restore RNG info struct in the GameVariables. Sets the `seed` and restores the state for
+ *         each independent sequence
  *
  * @param info RngInfo struct applied
  *
--- a/include/random.h
+++ b/include/random.h
@@ -46,7 +46,7 @@ typedef struct
 } RngInfo;
 
 /**
- * @brief Starts counting CPU cycles, this will be used by rng_shuffle_seed to
+ * @brief Starts counting CPU cycles, which will be used by rng_shuffle_seed to
  *         generate a more random seed. To be called once on game start.
  */
 void rng_init(void);
@@ -66,10 +66,15 @@ void rng_update(void);
 void rng_set_seed(u32 seed);
 
 /**
- * @brief Uses the CPU cycles counter to randomize the RNG seed as much as possible.
- *         This will be called by the main menu and the game over screens so that
- *         the next run's seed isn't the same as the last's.
- *         rng_start_sampling needs to have been called, and will stop the profiling.
+ * @brief Sets a new randomized seed for the RNG using a source of entropy.
+ *
+ * This needs to be called at least once before generating anything random but not on the very
+ * first cycle because it uses a CPU timer for entropy.
+ *
+ * This would also be used to generate the new seed when starting a new non-player-seeded run
+ *
+ * @note `rng_init` needs to have been called, so that the CPU profiling can provide entropy.
+ * @sa rng_init
  */
 void rng_shuffle_seed(void);
 
--- a/include/random.h
+++ b/include/random.h
@@ -22,17 +22,13 @@ enum RngSequence
     RNG_SEQ_SHOP_ITEMS,
     RNG_SEQ_SKIP_TAGS,
 
-    /**
-     * Each Joker with a random effect has its own independent RNG sequence
-     */
+    // Each Joker with a random effect has its own independent RNG sequence
     RNG_SEQ_JOKER_MISPRINT,
     RNG_SEQ_JOKER_RESERVED_PARKING,
     RNG_SEQ_JOKER_BUSINESS_CARD,
 
-    /**
-     * @brief For non-gameplay related things such as sound effects or visual effects, so as to not
-     *         interfere with important stuff like Shop rolls or Joker effects.
-     */
+    // For non-gameplay related things such as sound effects or visual effects, so as to not
+    // interfere with important stuff like Shop rolls or Joker effects.
     RNG_SEQ_MISC,
 
     RNG_SEQ_MAX
--- a/include/random.h
+++ b/include/random.h
@@ -23,20 +23,10 @@ enum RngSequence
     RNG_SEQ_SKIP_TAGS,
 
     /**
-     * @brief Used for rolling the Misprint Joker value.
-     *
-     * Each Joker with a random effect has its own indpendent RNG sequence
+     * Each Joker with a random effect has its own independent RNG sequence
      */
     RNG_SEQ_JOKER_MISPRINT,
-
-    /**
-     * @brief Used to determine if the Reserved Parking Joker triggers.
-     */
     RNG_SEQ_JOKER_RESERVED_PARKING,
-
-    /**
-     * @brief Used to determine if the Business Card Joker triggers.
-     */
     RNG_SEQ_JOKER_BUSINESS_CARD,
 
     /**
--- a/include/random.h
+++ b/include/random.h
@@ -12,15 +12,40 @@
 
 #include <tonc.h>
 
-enum RngType
+/**
+ * @brief Keys to different independent RNG sequences
+ */
+enum RngSequence
 {
-    RNG_TYPE_CARD_SHUFFLE,
-    RNG_TYPE_BLIND,
-    RNG_TYPE_SHOP_ITEMS,
-    RNG_TYPE_SKIP_TAGS,
-    RNG_TYPE_JOKER_EFFECTS,
-    RNG_TYPE_MISC,
-    RNG_TYPE_MAX
+    RNG_SEQ_CARD_SHUFFLE,
+    RNG_SEQ_BLIND,
+    RNG_SEQ_SHOP_ITEMS,
+    RNG_SEQ_SKIP_TAGS,
+
+    /**
+     * @brief Used for rolling the Misprint Joker value.
+     *
+     * Each Joker with a random effect has its own indpendent RNG sequence
+     */
+    RNG_SEQ_JOKER_MISPRINT,
+
+    /**
+     * @brief Used to determine if the Reserved Parking Joker triggers.
+     */
+    RNG_SEQ_JOKER_RESERVED_PARKING,
+
+    /**
+     * @brief Used to determine if the Business Card Joker triggers.
+     */
+    RNG_SEQ_JOKER_BUSINESS_CARD,
+
+    /**
+     * @brief For non-gameplay related things such as sound effects or visual effects, so as to not
+     *         interfere with important stuff like Shop rolls or Joker effects.
+     */
+    RNG_SEQ_MISC,
+
+    RNG_SEQ_MAX
 };
 
 /**
@@ -31,7 +56,7 @@ typedef struct
     /** Initial seed */
     u32 seed;
     /** Individual states for independent rng sequences */
-    u32 states[RNG_TYPE_MAX];
+    u32 states[RNG_SEQ_MAX];
 } RngInfo;
 
 /**
@@ -65,14 +90,11 @@ void rng_shuffle_seed(void);
 /**
  * @brief Get the next "randomly" generated number in the sequence corresponding to the given type.
  *
- * @note Custom rng had to be implemented to be able to manage several independent sequences, since
- *        `initstate` and `setstate` are POSIX and not available on GBA via devkitpro.
- *
- * @param type rng key to the sequence we need to pull a random number from
+ * @param key key to the RNG sequence we need to pull a random number from
  *
  * @return u32
  */
-u32 rng_get_u32(enum RngType type);
+u32 rng_get_u32(enum RngSequence key);
 
 /**
  * @brief Restore RNG info struct in the GameVariables. Sets the `seed` and restores the state for
--- a/source/game/run_setup.c
+++ b/source/game/run_setup.c
@@ -918,7 +918,7 @@ static inline void reroll_seed_str(void)
     // Also, don't use the shuffled seed as is, or we'll just end up with sequential seeds when
     // rolling multiple times.
     rng_shuffle_seed();
-    u32 new_seed = rng_get_u32();
+    u32 new_seed = rng_get_u32(RNG_TYPE_MISC);
     rng_set_seed(new_seed);
     u32_to_base36(new_seed, s_seed_str);
     update_seed_text();
--- a/source/game/run_setup.c
+++ b/source/game/run_setup.c
@@ -918,8 +918,6 @@ static inline void reroll_seed_str(void)
     // Also, don't use the shuffled seed as is, or we'll just end up with sequential seeds when
     // rolling multiple times.
     rng_shuffle_seed();
-    u32 new_seed = rng_get_u32(RNG_TYPE_MISC);
-    rng_set_seed(new_seed);
     u32_to_base36(g_game_vars.rng_info.seed, s_seed_str);
     update_seed_text();
     s_seed_cursor_pos = BASE36_MAX_DIGITS;
@@ -1191,8 +1189,9 @@ static void seed_on_pressed(void)
  */
 static void play_on_pressed(void)
 {
-    // Apply provided Seed if enabled
-    if (use_seed)
+    // Apply provided Seed if enabled, and if we entered one. This prevents us from always using
+    // seed "ZZZZZZ" if we enter the Seed menu and hit Play without typing anything.
+    if (use_seed && strlen(s_seed_str) > 0)
         rng_set_seed(base36_to_u32(s_seed_str));
     else
         rng_shuffle_seed();
--- a/source/game/run_setup.c
+++ b/source/game/run_setup.c
@@ -920,7 +920,7 @@ static inline void reroll_seed_str(void)
     rng_shuffle_seed();
     u32 new_seed = rng_get_u32(RNG_TYPE_MISC);
     rng_set_seed(new_seed);
-    u32_to_base36(new_seed, s_seed_str);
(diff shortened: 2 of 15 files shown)


MeirGavish:
Good job tackling this! Implementing the RNG algorithm ourselves rather than relying on a library implementation feels a little iffy but if it's an implementation of a standard algorithm and it works, I think it's good enough for our purposes since we're not really concerned with security or someone cheating the game.

> Also, if we could wait for the Skip Tags to be merged before merging this It would be great, I'd rather rebase this on main than the skip tags ><'

As for this, I actually think this should go in first. The main reason being that @ricfehr3 wanted make another small release for the recent improvements. Having this in will complete the seeded runs feature that was added in the previous release.
The skip tags, however, is a new big feature that warrants a bigger release and may introduce bugs we missed so it requires more testing.
The other reason is that this is a smaller PR so it's easier to review and integrate while the skip tags is a big PR with lots of changes that may take a little longer. I know resolving conflicts kind of sucks but it's just part of the development process and I think that in general the order of PRs being merged should be dictated by other things, such which one is ready first without too much dependency between them, or what should go in the next release.
That being said, I'm okay with giving priority to the skip tags PR if we do one of the following,
1. Make the release now with a seed system that's still a bit off.
2. Not make a release soon with the recent improvements and wait longer for a bigger release that includes the Skip Tags feature.

@ricfehr3 I honestly prefer making the release with this PR in first but what do you think?

MathisMartin31:
Yeaaaah you're right, and I did squash a lot of skip tags commits anyway so it shouldn't be too bad :)

MeirGavish:
I'll try to get to this soon but I'll request a copilot review in the meantime.

MeirGavish:
It's a busy time for me so reviews will take a while, I'm going to prioritize this PR for the release, I'm starting to look into reviewing it, I started by researching Balatro's RNG a little deeper.
Apparently, the sequence of generated items for the shop is separate to the sequence generated for packs, which is interesting but also makes sense pretty much. I don't think it will change this PR in any way (or at least not in a significant way) but it's still something to be considered.

MathisMartin31:
> It's a busy time for me so reviews will take a while, I'm going to prioritize this PR for the release, I'm starting to look into reviewing it, I started by researching Balatro's RNG a little deeper.

No worries take the time you need don't burn yourself out ^^

> Apparently, the sequence of generated items for the shop is separate to the sequence generated for packs, which is interesting but also makes sense pretty much. I don't think it will change this PR in any way (or at least not in a significant way) but it's still something to be considered.

That's interesting, we can add an argument to the joker roll function when we implement packs so that the ids are rolled from the right sequence. Adding sequences is as simple as adding a new entry to the `RngType` enum so on this PR's side of things the change is trivial

source/random.c
```diff
@@ -23,29 +23,38 @@ void rng_update(void)
...
+    // method used will stay stuck.
+    u32 capped_seed = seed % (MAX_BASE36 + 1);
+    g_game_vars.rng_info.seed = (capped_seed == 0) ? MAX_BASE36 : capped_seed;
     srand(g_game_vars.rng_info.seed);
+
+    // Generate rng states for all RngTypes categories using the given seed
+    for (enum RngType type = 0; type < RNG_TYPE_MAX; type++)
+    {
+        g_game_vars.rng_info.states[type] = rand();
+    }
```
MeirGavish: Express yourself in code rather than in comments, either remove the comment (the code seems self-descriptive enough) or extract the for loop into a function  with a name derived from the comment (`init_rng_states()`?) for an even clearer self-descriptive code.

source/game/run_setup.c
```diff
@@ -918,9 +918,9 @@ static inline void reroll_seed_str(void)
     // Also, don't use the shuffled seed as is, or we'll just end up with sequential seeds when
     // rolling multiple times.
     rng_shuffle_seed();
-    u32 new_seed = rng_get_u32();
+    u32 new_seed = rng_get_u32(RNG_TYPE_MISC);
```
MeirGavish: I think it _might_ be better to just use `rand()` here instead of using our sequence when it's only been partially initialized.
Using it here is at the very least misleading since it's not really needed.

source/random.c
```diff
@@ -23,29 +23,38 @@ void rng_update(void)
...
-    g_game_vars.rng_info.step++;
-    return rand();
+    // Xorshift32
+    uint32_t old_state = g_game_vars.rng_info.states[type];
+    old_state ^= old_state << 13;
+    old_state ^= old_state >> 17;
+    old_state ^= old_state << 5;
+
+    g_game_vars.rng_info.states[type] = old_state;
+    return old_state;
```
MeirGavish: Why did you call it `old_state` if there isn't a `new_state`? It's a little confusing, just call it `state`.
Also extract the operation into a `xorshift32()` function, it avoids the need for a comment to explain the algorithm and separates the algorithm operation from the state update.

source/random.c
```diff
@@ -23,29 +23,38 @@ void rng_update(void)
...
     rng_set_seed(s_timer_acc);
 }
 
-u32 rng_get_u32(void)
+u32 rng_get_u32(enum RngType type)
 {
-    g_game_vars.rng_info.step++;
-    return rand();
+    // Xorshift32
+    uint32_t old_state = g_game_vars.rng_info.states[type];
```
MeirGavish:
```suggestion
    u32 old_state = g_game_vars.rng_info.states[type];
```

Also use u32 to be consistent with the file.

include/random.h
```diff
@@ -12,15 +12,26 @@
...
 #include <tonc.h>
 
+enum RngType
+{
+    RNG_TYPE_CARD_SHUFFLE,
+    RNG_TYPE_BLIND,
+    RNG_TYPE_SHOP_ITEMS,
+    RNG_TYPE_SKIP_TAGS,
+    RNG_TYPE_JOKER_EFFECTS,
+    RNG_TYPE_MISC,
```
MeirGavish:
```suggestion
    RNG_TYPE_MISC,  // For non-gameplay related things such as sound effects or visual effects
```
I think this one needs a comment to clearly specify what it's for. I think anything that's gameplay related will need its own sequence.

include/random.h
```diff
@@ -52,15 +63,20 @@ void rng_set_seed(u32 seed);
 void rng_shuffle_seed(void);
 
 /**
- * @brief Get the next "randomly" generated number in the sequence from the current seed.
+ * @brief Get the next "randomly" generated number in the sequence corresponding to the given type.
+ *
+ * @note Custom rng had to be implemented to be able to manage several independent sequences, since
+ *        `initstate` and `setstate` are POSIX and not available on GBA via devkitpro.
```
MeirGavish: This note is more of an implementation detail and not really relevant to the caller of this function.
It would be better to put it where xorshift32 is used to explain why we're not using a library rand().
I'm also not sure the POSIX functions have to be mentioned, the implementation discussion is already documented on Github.

include/random.h
```diff
@@ -12,15 +12,26 @@
...
+enum RngType
+{
+    RNG_TYPE_CARD_SHUFFLE,
+    RNG_TYPE_BLIND,
+    RNG_TYPE_SHOP_ITEMS,
+    RNG_TYPE_SKIP_TAGS,
+    RNG_TYPE_JOKER_EFFECTS,
+    RNG_TYPE_MISC,
+    RNG_TYPE_MAX
+};
```
MeirGavish: Maybe rename to `RngSequence`/`RNG_SEQUENCE_*` (maybe even abbreviate `RngSeq`/`RNG_SEQ_*`) or `RngKey`/`RNG_KEY_*`, I think "type" is a little less fitting and clear here.

source/joker.c
```diff
@@ -360,7 +360,7 @@ static int joker_roll_id(void)
     int joker_rarity = joker_get_random_rarity();
 
     int matching_joker_ids[jokers_avail_size];
-    int fallback_random_idx = rng_get_u32() % jokers_avail_size;
+    int fallback_random_idx = rng_get_u32(RNG_TYPE_SHOP_ITEMS) % jokers_avail_size;
```
MeirGavish: I think it makes sense for this PR to make the RNG sequence a parameter to the the Item `roll_new()` function, including `joker_object_roll_new()`, `joker_roll_id()`, and `joker_get_random_rarity()` so the sequence they use is propagated down.
The current code just assumes all jokers are generated for the shop but it will not stay that way.

source/game/run_setup.c
```diff
@@ -918,7 +918,7 @@ static inline void reroll_seed_str(void)
...
     rng_set_seed(new_seed);
     u32_to_base36(new_seed, s_seed_str);
```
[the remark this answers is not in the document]
MeirGavish: @MathisMartin31
Referring to [Copilot's comment](https://github.com/GBALATRO/balatro-gba/pull/599#discussion_r3651857328)
This was an issue before this PR, wasn't it?
It feels like this code is a little convoluted and hard to follow and it makes it more error-prone, you have to look at the function implementations to understand it.
For this specific error you can extract the code that caps the number to `MAX_BASE36` to a function and then you can just use it here with `new_seed` instead of relying on `rng_set_seed()` to cap it when updating the global seed.
Other than that it feels like the code here needs some more refactoring to make more sense but I'm not exactly sure what and it should probably wait for another refactor PR...

source/joker_effects.c
```diff
@@ -1301,7 +1301,7 @@ static u32 reserved_parking_joker_effect(
 
     u32 effect_flags_ret = JOKER_EFFECT_FLAG_NONE;
 
-    if ((rng_get_u32() % 2 == 0) && card_is_face(scored_card))
+    if ((rng_get_u32(RNG_TYPE_JOKER_EFFECTS) % 2 == 0) && card_is_face(scored_card))
```
```suggestion
    if (card_is_face(scored_card) && (rng_get_u32(RNG_TYPE_JOKER_EFFECTS) % 2 == 0))
```
Swap so the short-circuit condition checks if it's a face card before rolling the random number so the sequence is only advanced on face cards and not on every card in hand.

source/joker_effects.c
```diff
@@ -1323,7 +1323,7 @@ static u32 business_card_joker_effect(
...
```
```suggestion
    if (card_is_face(scored_card) && (rng_get_u32(RNG_TYPE_JOKER_EFFECTS) % 2 == 0))
```

Likewise swap the check so the sequence is not advanced on every card scored.

include/random.h
```diff
@@ -12,15 +12,26 @@
 
 #include <tonc.h>
 
+enum RngType
+{
+    RNG_TYPE_CARD_SHUFFLE,
+    RNG_TYPE_BLIND,
+    RNG_TYPE_SHOP_ITEMS,
+    RNG_TYPE_SKIP_TAGS,
+    RNG_TYPE_JOKER_EFFECTS,
```
MeirGavish: So in Balatro every joker with a random effect has its own random sequence.
With this single sequence implementation, say, if you have both Reserved Parking and Business card and one should trigger and the other should not, their order affects which one triggers.
I guess it's not such a big deal for these 2 jokers but let's say we had Riff-Raff too, the jokers it would generate would depend on how many face cards you played in the previous round and that's a little weird.
On the other hand to really match Balatro we would probably need over a hundred sequences in the end and I'm wondering if it's too much.

source/random.c
```diff
@@ -23,29 +23,38 @@ void rng_update(void)
...
 void rng_restore(RngInfo info)
 {
     g_game_vars.rng_info = info;
-
-    srand(g_game_vars.rng_info.seed);
-    for (u32 i = 0; i < g_game_vars.rng_info.step; i++)
-    {
-        (void)rng_get_u32();
-    }
 }
```
MeirGavish, line 80: It's nice to be able to remove this

MeirGavish (changes requested):
Good job implementing this! I thought it would be more complicated but your solution is nice and straightforward.
I made a pretty thorough review but I still haven't tested this, posting my review now and I will come back later to test it.

source/game/run_setup.c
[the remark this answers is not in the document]
MathisMartin31: I think it's better to just use `rng_shuffle_seed` here and get the seed it comes up with, since it uses the clock counter to generate a rather good random number

include/random.h
MathisMartin31: Yeah I think it's better to give every Joker with a random effect its own sequence, it's more consistent with the seeded random

MathisMartin31:
@MeirGavish Thanks for the extensive review ^^, it's good for another round, hope I didn't miss anything

include/random.h
```diff
@@ -12,15 +12,51 @@
...
+
+    /**
+     * @brief Used to determine if the Reserved Parking Joker triggers.
+     */
+    RNG_SEQ_JOKER_RESERVED_PARKING,
+
+    /**
+     * @brief Used to determine if the Business Card Joker triggers.
+     */
+    RNG_SEQ_JOKER_BUSINESS_CARD,
```
MeirGavish:
```suggestion
    /**
     * Each Joker with a random effect has its own independent RNG sequence
     */
    RNG_SEQ_JOKER_MISPRINT,
    RNG_SEQ_JOKER_RESERVED_PARKING,
    RNG_SEQ_JOKER_BUSINESS_CARD,
```

It would be much better to not document every joker in the enum, the comments are redundant since they are fully inferred by the name and they take up a lot of vertical space in an enum that is expected to grow a lot.
Also, typo fix "independent".

source/joker.c
```diff
@@ -538,7 +538,7 @@ Sprite* joker_object_get_sprite(JokerObject* joker_object)
 int joker_get_random_rarity()
 {
     int joker_rarity = 0;
-    int rarity_roll = rng_get_u32() % 100;
+    int rarity_roll = rng_get_u32(RNG_SEQ_SHOP_ITEMS) % 100;
```
MeirGavish: Pass the sequence key here too.

source/random.c
```diff
@@ -21,31 +21,57 @@ void rng_update(void)
     s_timer_acc += (u32)REG_TM1D;
 }
 
-void rng_set_seed(u32 seed)
+/**
+ * @brief Reset all independent RNG sequences to their initial states using the
+ *         provided custom seed.
```
MeirGavish:
```suggestion
 * @brief Reset all independent RNG sequences to their initial states using the
 *         global custom seed.
```

To clarify the documentation since the seed is not a parameter

MeirGavish (changes requested):
Looks good, I only had a little bit of time today so this is a partial review, just a few small comments but I still want to take a closer look at `reroll_seed_str()`, `rng_shuffle_seed()`, and `rng_set_seed()`. I just figured I should  post it now instead of keeping you waiting.
Also, I still want to test this...

source/joker.c
```diff
@@ -345,6 +345,37 @@ void joker_reset_rollable_jokers(void)
...
+        joker_rarity = RARE_JOKER;
+    }
+    else if (rarity_roll < COMMON_JOKER_CHANCE + UNCOMMON_JOKER_CHANCE + RARE_JOKER_CHANCE +
+                               LEGENDARY_JOKER_CHANCE)
+    {
+        joker_rarity = LEGENDARY_JOKER;
+    }
+
+    return joker_rarity;
+}
```
MeirGavish: I'm not sure `joker_get_random_rarity()` should be static (but I'm okay with it), but if you do make it static don't move the function up, use a forward declaration. The code is more readable when the called function is below the calling function.
I used to do this too because of how C works and to avoid forward declarations but I've been reading the book Clean Code (by Robert C. Martin) and it pointed out that that makes the code more confusing to read so I'm thoroughly convinced.

source/random.c
```diff
@@ -21,31 +21,57 @@ void rng_update(void)
...
+    *(state) ^= *(state) << 13;
+    *(state) ^= *(state) >> 17;
+    *(state) ^= *(state) << 5;
+}
+
+u32 rng_get_u32(enum RngSequence key)
+{
+    xorshift32(&g_game_vars.rng_info.states[key]);
+    return g_game_vars.rng_info.states[key];
 }
```
MeirGavish: While we're at it, let's flip the order of caller-called with forward declarations for these functions too?

source/game/run_setup.c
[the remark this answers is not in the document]
MeirGavish: Replying to [comment](https://github.com/GBALATRO/balatro-gba/pull/599#discussion_r3690108356)
Yeah, I think it's okay with your latest changes, you do need to call `rng_shuffle_seed()` at least once before generating anything random to get actual random numbers, my problem was with the mixed use of `new_seed`.

include/random.h
```diff
@@ -52,15 +74,17 @@ void rng_set_seed(u32 seed);
 void rng_shuffle_seed(void);
```
MeirGavish, line 79: So now that I understand it, I see that the documentation for this function is confusing and stale, `rng_start_sampling` doesn't exist.
How about this one? (Github doesn't let me suggest it as a change because these lines didn't change)
"Sets a new randomized seed for the RNG using a source of entropy.
This needs to be called at least once before generating anything random but not on the very first cycle because it uses a timer for entropy.
This would also be used to generate the new seed when starting a new non-player-seeded run"

MeirGavish (changes requested):
Okay, looking good! Just a few more changes to documentation and function ordering, and we're good to go.
I also tested it and it seems to be working fine.

MathisMartin31:
> Okay, looking good! Just a few more changes to documentation and function ordering, and we're good to go.
> I also tested it and it seems to be working fine.

Thanks again for the review, should be good to go for another round ^^

source/random.c
```diff
@@ -34,22 +47,18 @@ static void init_rng_states(void)
...
-    u32 capped_seed = seed % (MAX_BASE36 + 1);
-    g_game_vars.rng_info.seed = (capped_seed == 0) ? MAX_BASE36 : capped_seed;
-    init_rng_states();
-}
-
 void rng_shuffle_seed(void)
 {
     srand(s_timer_acc);
     rng_set_seed(rand());
 }
```
MeirGavish: By the same principle, `rng_shuffle_seed()` needs to be above `rng_set_seed()`

source/joker.c
```diff
@@ -449,6 +419,37 @@ Item* joker_object_roll_new(enum RngSequence key)
     return (Item*)joker_object_new(joker_new(joker_id));
 }
 
+/**
+ * @brief Get a random rarity a Joker will be rolled from
+ *
+ * @param key to the RNG sequence used
+ * @return a random Joker rarity
+ */
+static inline int s_joker_get_random_rarity(enum RngSequence key)
```
MeirGavish: I'm not sure I would add the `s_` prefix to this particular function since as I said I'm not sure it would stay static.

MeirGavish (approved):
Looks good! I left 2 last comments but it's been a long back-and-forth and they're small nitpicks by now so I'll approve this.
I will want to merge this soon and make a release with the completed seed system.
@ricfehr3 Let me know if you want to take a look to or have any thoughts.
Edit: I will merge this in a day or two.

MathisMartin31:
> Looks good! I left 2 last comments but it's been a long back-and-forth and they're small nitpicks by now so I'll approve this. I will want to merge this soon and make a release with the completed seed system. @ricfehr3 Let me know if you want to take a look to or have any thoughts.

Nice, thanks for this ^^
I did go ahead and fix the issues you mentioned :)

MeirGavish:
@ricfehr3
Mentioning here since it's the last PR for the release, we can move to a discussion or issue if needed.
I made a draft for the release, feel free to change anything if you feel the need and let me know.
I also tested the current main branch before the release. I played 2 runs with the same seed. There were slight variations later in the run but I think they were mostly due to owning different jokers.
I will make the release in a day or two.

PR #612, opened 2026-08-10
[DRAFT FOR @juanvillacortac] Joker Round description

MathisMartin31:
Here it is ^^

The relevant code for you @juanvillacortac is in `item_description.c`. You should also look up `background_gfx.png` to see how the layout is done to free up space.

If you need ask anything about this code feel free to do so in the comments here, I'll give you as many pointers as I can.

PR #613, opened 2026-08-10
[Refactor] Port Jokers descriptions to the Item API in the Shop

### LINKED ISSUE #594: Add description functionality to `Item` and refactor Joker description code to use it
This will allow consumables to implement descriptions. I guess this doesn't have to be a prerequisite, consumables can forego descriptions at first but I think it's still better to do this first.

MathisMartin31:
Closes #594

Extracted from my draft PR #612, and removed any Skip Tags code inherited from #541 but I kept the description states refactor and the more responsive behaviour when pressing and releasing B.

### DIFF
--- a/include/game/shop.h
+++ b/include/game/shop.h
@@ -17,11 +17,11 @@
 void game_shop_reset(void);
 
 /**
- * @brief Get a pointer to the Card we are currently showing the description of.
+ * @brief Get a pointer to the Item we are currently showing the description of.
  *
- * @return JokerObject*
+ * @return Item*
  */
-JokerObject* game_shop_get_description_card(void);
+Item* game_shop_get_description_item(void);
 
 /**
  * @brief Change to the shop background
--- a/include/item.h
+++ b/include/item.h
@@ -11,9 +11,12 @@
 #ifndef ITEM_H
 #define ITEM_H
 
+#include "graphic_utils.h"
 #include "mgba_logger.h"
 #include "random.h"
 #include "sprite.h"
+
+#include <stdint.h>
 
 // TODO: Merge these with GBAL_RETURN_IF_NULL macros?
 /**
@@ -98,9 +101,14 @@ typedef struct ItemFuncs
      */
     Item* (*roll_new)(enum RngSequence key);
     int (*get_buy_price)(Item* item);
+    int (*get_sell_price)(Item* item);
+    const char* (*get_name)(Item* item);
+    const char* (*get_subtype_str)(Item* item);
+    uint32_t (*get_subtype_colors)(Item* item);
     bool (*can_acquire)(Item* item);
     void (*acquire)(Item* item);
     void (*dispose)(Item** item);
+    int (*print_desc)(Item* item, Rect dest_rect);
     // TODO: void (*print_description)(Item* item); // or something of the form
 
     // Optional implementation functions will be added here
@@ -130,6 +138,50 @@ Item* item_roll_new(enum ItemType item_type, enum RngSequence key);
  * @return UNDEFINED in case of error, the item's buy price otherwise.
  */
 int item_get_buy_price(Item* item);
+
+/**
+ * @brief Returns the sell price of the item.
+ *
+ * Matches @ref ItemFuncs.get_sell_price()
+ *
+ * @param item The item whose price to return.
+ *
+ * @return UNDEFINED in case of error, the item's sell price otherwise.
+ */
+int item_get_sell_price(Item* item);
+
+/**
+ * @brief Returns the name of the Item
+ *
+ * Matches @ref ItemFuncs.get_name()
+ *
+ * @param item The item whose name to return.
+ *
+ * @return const char*
+ */
+const char* item_get_name(Item* item);
+
+/**
+ * @brief Returns the name of the Item's subtype (rarity for Jokers, just the type otherwise)
+ *
+ * Matches @ref ItemFuncs.get_subtype_str()
+ *
+ * @param item The item whose subtype's name to return.
+ *
+ * @return const char*
+ */
+const char* item_get_subtype_string(Item* item);
+
+/**
+ * @brief Returns the colors of the Item's subtype
+ *
+ * Matches @ref ItemFuncs.get_subtype_colors()
+ *
+ * @param item The item whose subtype's color to return.
+ *
+ * @return values of both main and shadow colors encoded into a single u32 (main first)
+ */
+uint32_t item_get_subtype_colors(Item* item);
 
 /**
  * @brief Acquires the item, adding to inventory if applicable.
@@ -171,5 +223,15 @@ void item_dispose(Item** item);
  * @param item The item to print under
  */
 void item_print_buy_price_under(Item* item);
+
+/**
+ * @brief Prints the item's description inside the given rectangle
+ *
+ * @param item The item to print the description of
+ * @param dest_rect the target rectangle the description needs to fit in
+ *
+ * @return the number of lines used by the description
+ */
+int item_print_description(Item* item, Rect dest_rect);
 
 #endif // ITEM_H
--- a/include/joker.h
+++ b/include/joker.h
@@ -171,7 +171,9 @@ u32 joker_get_score_effect(
     JokerEffect** joker_effect
 );
 
-const char* joker_get_rarity_string(u8 rarity);
+const char* joker_object_get_name(Item* joker_object);
+
+const char* joker_object_get_rarity_string(Item* joker_object);
 
 /**
  * @brief Get Joker rarity panel color.
@@ -186,13 +188,10 @@ const char* joker_get_rarity_string(u8 rarity);
  *  - 11,12 -> Spectral Card
  *  - 13,14 -> Voucher
  *
- * @param rarity Value of the rarity (Common, Rare...)
- * @param main_color Whether we want the main or shadow color
- * @return u16 value of the color, not a pointer
+ * @param joker_object pointer to a JokerObject we need the rarity's color of
+ * @return value of both main and shadow colors, encoded into a single u32 (main first)
  */
-u16 joker_get_rarity_color(u8 rarity, bool main_color);
-
-int joker_get_sell_value(const Joker* joker);
+u32 joker_object_get_rarity_colors(Item* joker_object);
 
 JokerObject* joker_object_new(Joker* joker);
 void joker_object_destroy(JokerObject** joker_object);
@@ -208,6 +207,15 @@ void joker_object_shake(JokerObject* joker_object, mm_word sound_id);
  */
 int joker_object_get_buy_price(Item* joker_object);
 
+/**
+ * @brief Returns the sell price of the joker object.
+ *
+ * @param joker_object the joker object whose price to return.
+ *
+ * @return UNDEFINED in case of error, the sell price of the joker otherwise.
+ */
+int joker_object_get_sell_price(Item* joker_object);
+
 // TODO: Move to an owned_jokers.c/.h file?
 /**
  * @brief Add a Joker to the list of owned Jokers and place it in the joker row.
@@ -222,6 +230,17 @@ void joker_object_add_to_owned(Item* joker_object);
  * @param joker_object Pointer to the JokerObject Item* to destroy; set to NULL.
  */
 void joker_object_dispose(Item** joker_object);
+
+int joker_object_print_description(Item* joker_object, Rect dest_rect);
+
+/**
+ * @brief Rolls a random Joker among the available ones of a certain rarity
+ *
+ * @param joker_rarity the rarity of the joker we want to roll
+ *
+ * @return rolled joker ID
+ */
+int joker_roll_id_with_rarity(int joker_rarity);
 
 /**
  * @brief Set whether a Joker is available to be rolled for the shop, packs, etc.
--- a/source/game/shop.c
+++ b/source/game/shop.c
@@ -31,12 +31,12 @@
 #define TM_END_GAME_SHOP_INTRO    12
 #define TM_CREATE_SHOP_ITEMS_WAIT 1
 #define TM_SHIFT_SHOP_ICON_WAIT   7
-#define TM_SHOW_CARD_DESC_WAIT    12
+#define TM_SHOW_ITEM_DESC_WAIT    12
 #define TM_HIDE_DECK_WAIT         5
 
 // Pixel sized
 #define ITEM_SHOP_Y               71
-#define OWNED_CARDS_HIDE_Y_OFFSET 50
+#define OWNED_ITEMS_HIDE_Y_OFFSET 50
 
 // Shop
 #define REROLL_BASE_COST     5 // Base cost for rerolling the shop items
@@ -65,26 +65,26 @@
 // Positions in tiles
 static const Rect     SHOP_ICON_FROM_RECT           = {  0, 26,  8, 26};
 static const BG_POINT SHOP_ICON_TO_POS              = {  0,  0};
-static const BG_POINT OWNED_CARDS_PANEL_3X3_SRC_POS = { 29, 21};
+static const BG_POINT OWNED_ITEMS_PANEL_3X3_SRC_POS = { 29, 21};
 static const Rect     OWNED_JOKERS_PANEL_RECT       = {  9,  1, 21,  5};
 static const Rect     OWNED_CONSUMABLES_PANEL_RECT  = { 23,  1, 28,  5};
-static const Rect     OWNED_CARDS_PANEL_RECT        = {  9,  1, 28,  5};
-static const Rect     OWNED_CARDS_PANEL_ANIM_CLEAR  = {  9,  0, 28,  1};
-static const Rect     CARD_DESC_9_PTCH_TO_RECT      = {  9,  6, 28, 18};
-static const NinePatchRect CARD_DESC_9_PTCH_SRC = {
+static const Rect     OWNED_ITEMS_PANEL_RECT        = {  9,  1, 28,  5};
+static const Rect     OWNED_ITEMS_PANEL_ANIM_CLEAR  = {  9,  0, 28,  1};
+static const Rect     ITEM_DESC_9_PTCH_TO_RECT      = {  9,  6, 28, 18};
+static const NinePatchRect ITEM_DESC_9_PTCH_SRC = {
                                         .patch_rect = { 27, 25, 31, 31},
                                         .margins    = {  2,  3,  2,  3}
 };
-static const int      CARD_DESC_MAX_TEXT_HEIGHT     = CARD_DESC_9_PTCH_TO_RECT.bottom -
-                                                      CARD_DESC_9_PTCH_TO_RECT.top + 1 -
-                                                      CARD_DESC_9_PTCH_SRC.margins.top -
-                                                      CARD_DESC_9_PTCH_SRC.margins.bottom;
-static const Rect     CARD_DESC_TEXT_RECT           = { 11,  9, 26, 18};
-static const Rect     CARD_NAME_TEXT_RECT           = { 10,  7, 27,  7};
+static const int      ITEM_DESC_MAX_TEXT_HEIGHT     = ITEM_DESC_9_PTCH_TO_RECT.bottom -
+                                                      ITEM_DESC_9_PTCH_TO_RECT.top + 1 -
+                                                      ITEM_DESC_9_PTCH_SRC.margins.top -
+                                                      ITEM_DESC_9_PTCH_SRC.margins.bottom;
+static const Rect     ITEM_DESC_TEXT_RECT           = { 11,  9, 26, 18};
+static const Rect     ITEM_NAME_TEXT_RECT           = { 10,  7, 27,  7};
 
 // Positions in pixels
 static const BG_POINT SHOP_JOKER_SPRITES_INIT_POS = {120, 160};
-static const BG_POINT CARD_DESCRIPTION_SPRITE_POS = {135,   9};
+static const BG_POINT ITEM_DESCRIPTION_SPRITE_POS = {135,   9};
 static const Rect     SHOP_PRICES_TEXT_RECT       = { 72,  56, 192, 160 };
 static const Rect     SHOP_REROLL_RECT            = { 88,  96, UNDEFINED, UNDEFINED };
 // clang-format on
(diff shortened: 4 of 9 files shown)


MathisMartin31:
@MeirGavish Sorry I didn't see the conflicts ><, everything should be good to go now :3

source/game/shop.c
```diff
@@ -586,182 +592,166 @@ static void shop_process_user_input(void)
...
 
         // Draw description panel
-        Rect actual_dest_rect = CARD_DESC_9_PTCH_TO_RECT;
+        Rect actual_dest_rect = ITEM_DESC_9_PTCH_TO_RECT;
         actual_dest_rect.bottom -= desc_bottom_offset;
-        main_bg_se_copy_expand_9_patch(actual_dest_rect, &CARD_DESC_9_PTCH_SRC);
+        main_bg_se_copy_expand_9_patch(actual_dest_rect, &ITEM_DESC_9_PTCH_SRC);
 
         // Print joker name
+        const char* item_name = item_get_name(s_description_item);
```
ricfehr3: This can be `NULL`. The derived function for `Item->get_name()` can return NULL here. 

Passing NULL to `printf` and `strlen` is undefined behavior. https://stackoverflow.com/questions/11589342/what-is-the-behavior-of-printing-null-with-printfs-s-specifier


I think we should at least log a warning here.

source/game/shop.c
```diff
@@ -586,182 +592,166 @@ static void shop_process_user_input(void)
...
-            CARD_DESC_MAX_TEXT_HEIGHT -
-            info->joker_print_desc(s_description_card->joker, CARD_DESC_TEXT_RECT);
+        int desc_bottom_offset = ITEM_DESC_MAX_TEXT_HEIGHT -
+                                 item_print_description(s_description_item, ITEM_DESC_TEXT_RECT);
 
-        // Print Rarity and change color or the panel
+        // Print Rarity/Type and change color or the panel
         // Do it before drawing the panel so the color is already set
-        const char* rarity_str = joker_get_rarity_string(info->rarity);
+        const char* subtype_str = item_get_subtype_string(s_description_item);
```
ricfehr3, line 694: Same as: https://github.com/GBALATRO/balatro-gba/pull/613/changes#r4052024700

source/game/shop.c
```diff
@@ -586,182 +592,166 @@ static void shop_process_user_input(void)
...
-    else if (s_timer == TM_SHOW_CARD_DESC_WAIT + 1)
+    else if (s_timer == TM_SHOW_ITEM_DESC_WAIT + 1)
     {
         // Compute needed space for the description
-        const JokerInfo* info = get_joker_registry_entry(s_description_card->joker->id);
-        int desc_bottom_offset =
-            CARD_DESC_MAX_TEXT_HEIGHT -
-            info->joker_print_desc(s_description_card->joker, CARD_DESC_TEXT_RECT);
+        int desc_bottom_offset = ITEM_DESC_MAX_TEXT_HEIGHT -
+                                 item_print_description(s_description_item, ITEM_DESC_TEXT_RECT);
```
ricfehr3: I think it's fine in this case for `item_print_description` to have the dual responsibility of printing and returning the number of lines printed. I think it would make more sense here if there was an intermediate variable assigned, just so it's clear when reading:

```suggestion
        int printed_lines = item_print_description(s_description_item, ITEM_DESC_TEXT_RECT);
        int desc_bottom_offset = ITEM_DESC_MAX_TEXT_HEIGHT - printed_lines;
                                 
```

include/joker.h
```diff
@@ -223,6 +231,17 @@ void joker_object_add_to_owned(Item* joker_object);
  */
 void joker_object_dispose(Item** joker_object);
 
+int joker_object_print_description(Item* joker_object, Rect dest_rect);
```
ricfehr3, line 257: Looks like this is missing a Doxygen comment.

source/joker.c
```diff
@@ -165,22 +165,42 @@ u32 joker_get_score_effect(
...
+    Joker* joker = ((JokerObject*)joker_object)->joker;
+    const JokerInfo* info = get_joker_registry_entry(joker->id);
+    GBAL_RETURN_IF_NULL_RET(info, 0);
+    u8 rarity = joker->rarity;
 
     // +1 to account for the transparency
     // odd indices are the main colors, even ones are the shadows
-    return card_rarity_pal_gfxPal[1 + 2 * rarity + (main_color ? 0 : 1)];
+    u32 colors =
+        card_rarity_pal_gfxPal[1 + 2 * rarity] | (card_rarity_pal_gfxPal[1 + 2 * rarity + 1] << 16);
```
ricfehr3:
```suggestion
    u32 pal_base = 1 + 2 * rarity;
    u32 low = card_rarity_pal_gfxPal[pal_base];
    u32 high = card_rarity_pal_gfxPal[pal_base + 1] << 16;
    u32 colors = low | high;
```

Nitpick. I think this one could be represented better by breaking out the variables.

ricfehr3 (changes requested):
Nice work @MathisMartin31 . I found what may be some potential bugs, but easily mitigated. This is definitely a really good step for the game 💯

MeirGavish:
> @MeirGavish Sorry I didn't see the conflicts ><, everything should be good to go now :3

It's no problem, don't worry about it. I'm still reviewing #541 because it's been waiting for a long time and still haven't looked at this PR, but maybe I will jump to it and review it too as it's smaller and also quite important.

source/game/shop.c
```diff
@@ -586,182 +592,184 @@ static void shop_process_user_input(void)
...
-            CARD_DESC_TEXT_RECT.left * TILE_SIZE,
-            (CARD_DESC_TEXT_RECT.bottom - desc_bottom_offset - 1) * TILE_SIZE,
-            (rect_width(&CARD_DESC_TEXT_RECT) - strlen(rarity_str)) / 2,
+            ITEM_DESC_TEXT_RECT.left * TILE_SIZE,
+            (ITEM_DESC_TEXT_RECT.bottom - desc_bottom_offset - 1) * TILE_SIZE,
+            (rect_width(&ITEM_DESC_TEXT_RECT) - strlen(subtype_str)) / 2,
             "",
-            rarity_str
+            subtype_str
         );
```
ricfehr3, line 711: The `subtype_str` is checked for `NULL`, but is still passed to `strlen`, which is undefined behavior.

I got curious about this actually. For my own fun I looked into it more, specifically for GBA. The [`newlib`](https://github.com/devkitPro/newlib/tree/master) used in devkitpro has the implementations of `strlen` and `vnprintf` (`tteprintf` eventually calls `vnprintf`).

For `vnprintf` the comment acknowledges passing `NULL` as undefined behavior, but chooses to hardcode `(null)` for safety: https://github.com/devkitPro/newlib/blob/b14e2e7d15a8f4ef4829d797692b42541c61e0ed/newlib/libc/stdio/vfprintf.c#L1182

I tested this by doing:

```c
// Need to turn of -Werror for this to work
char* void_null = NULL;
tte_printf("%s", void_null);
```
I was able to see "(null)" 💯 
Image

For `strlen` there is no protection ([source code](https://github.com/devkitPro/newlib/blob/master/newlib/libc/string/strlen.c#L35)). It looks like without defining a `PREFER_SIZE_OVER_SPEED` macro or `-Os` (optimize size), the function is pretty simple (adding the optimizations does even more dangerous NULL pointer manipulation):

```c
// I just removed the code that isn't compiled in
strlen (const char *str)
{
  const char *start = str; // Now start is NULL

  while (*str) // What is the data at address 0x00? if it's not 0, this is cooked
    str++; 
  return str - start; // Not immediately destructive, but using this variable could cause dangerous problems
}
```

Unfortunately `strlen(NULL)` just crashes, on both MGBA and real hardware. :/ I was hoping it would be more interesting and maybe read address 0x00, but it just crashes, haha. 

Anyway, I think if `subtype_str` is `NULL` it should be set to a known value like in newlib. Something like:

```suggestion
        const char undef_str[] = "NO DESC.";
        if (subtype_str == NULL)
        {
            MGBA_FUNC_WARN(
                "Could not retrieve subtype name string of Item of type %d and id %d",
                s_description_item->type,
                s_description_item->id
            );
            
            subtype_str = (char*)undef_str;
        }

        tte_printf(
            TTE_WHITE_TAG "#{P:%d,%d}%*s%s",
            CARD_DESC_TEXT_RECT.left * TILE_SIZE,
            (CARD_DESC_TEXT_RECT.bottom - desc_bottom_offset - 1) * TILE_SIZE,
            (rect_width(&CARD_DESC_TEXT_RECT) - strlen(rarity_str)) / 2,
            "",
            rarity_str
            subtype_str
        );
);
```

source/joker.c
```diff
@@ -165,36 +165,50 @@ u32 joker_get_score_effect(
...
+    GBAL_RETURN_IF_NULL_RET(info, 0);
+    u8 rarity = joker->rarity;
+
+    // +1 to account for the mandatory transparency in Asperite color palettes
+    u32 pal_base = 1 + 2 * rarity;
+    u32 low = card_rarity_pal_gfxPal[pal_base];            // even indices are the shadows
+    u32 high = card_rarity_pal_gfxPal[pal_base + 1] << 16; // odd ones are the main colors
+
+    return low | high;
 }
```
ricfehr3: Will we usually be returning two `u16`s? If so, I think we should use a struct with two `u16`s instead of a `u32` here and for all items that implement this.

include/joker.h
```diff
@@ -171,7 +171,9 @@ u32 joker_get_score_effect(
     JokerEffect** joker_effect
 );
 
-const char* joker_get_rarity_string(u8 rarity);
+const char* joker_object_get_name(Item* joker_object);
+
+const char* joker_object_get_rarity_string(Item* joker_object);
```
ricfehr3, line 188: These might be useful to have Doxygen comments for since it's returning a string it would be nice to know if it can return `NULL` or not, for example.

ricfehr3 (changes requested):
Gave it another round of review. I had fun investigating this one, haha. Let me know what you think.

MeirGavish:
I noticed I forgot to request a review from Copilot so requesting one now.

source/joker.c
MeirGavish: Yes, I agree, definitely should be named

source/game/shop.c
```diff
@@ -586,182 +592,184 @@ static void shop_process_user_input(void)
...
-        const char* rarity_str = joker_get_rarity_string(info->rarity);
+        const char* subtype_str = item_get_subtype_string(s_description_item);
+        if (subtype_str == NULL)
+        {
+            MGBA_FUNC_WARN(
+                "Could not retrieve subtype name string of Item of type %d and id %d",
+                s_description_item->type,
+                s_description_item->id
+            );
+        }
```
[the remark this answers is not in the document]
MeirGavish: This is true, the issue doesn't show in regular compilation because logging is conditionally compiled but if you turn on logging, the build fails here.
I don't think the item struct should have an ID field, it can be omitted from the error message. We can also add an `item_get_id()` function but I think it's a little excessive just for an error message.

source/game/shop.c
MeirGavish: Yeah, I think `NULL` is always an invalid address, these crashes are why we add NULL-checks. I think this is a good solution. I'm also okay with just not printing a description at all in this case since an error is logged.
Your suggestion has a few duplicated lines which I'm editing to remove.

source/game/shop.c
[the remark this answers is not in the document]
MathisMartin31: Yeah, I don't know why but I was absolutely certain that the id was readable from the Item ><'...

source/game/shop.c
MathisMartin31: Interesting finds! I'll do what you suggested.
These NULL checks are still going to haunt me aren't they, no matter how much I try I just never think about them as much as I should ^^'

MathisMartin31:
@MeirGavish @ricfehr3 thank you both for your very fast reviews ^^
Should be good for you to take another look

source/joker.c
```diff
@@ -181,27 +181,33 @@ const char* joker_object_get_rarity_string(Item* joker_object)
...
+
+    GBAL_RETURN_IF_NULL_RET(joker_object, colors);
     Joker* joker = ((JokerObject*)joker_object)->joker;
     const JokerInfo* info = get_joker_registry_entry(joker->id);
-    GBAL_RETURN_IF_NULL_RET(info, 0);
+    GBAL_RETURN_IF_NULL_RET(info, colors);
+
     u8 rarity = joker->rarity;
+    if (rarity >= MAX_RARITIES)
+        return colors;
```
MeirGavish: And log an error

include/item.h
```diff
@@ -98,9 +115,14 @@ typedef struct ItemFuncs
...
     int (*get_buy_price)(Item* item);
+    int (*get_sell_price)(Item* item);
+    const char* (*get_name)(Item* item);
+    const char* (*get_subtype_str)(Item* item);
+    ItemSubtypeColors (*get_subtype_colors)(Item* item);
     bool (*can_acquire)(Item* item);
     void (*acquire)(Item* item);
     void (*dispose)(Item** item);
+    int (*print_desc)(Item* item, Rect dest_rect);
     // TODO: void (*print_description)(Item* item); // or something of the form
```
MeirGavish:
```suggestion
    int (*print_description)(Item* item, Rect dest_rect);
```

I think it's better to not shorten this in the name, it's the important part of the name.
Also, it's time to remove the TODO.

include/item.h
```diff
@@ -98,9 +115,14 @@ typedef struct ItemFuncs
...
 
     // Optional implementation functions will be added here
```
MeirGavish:
```suggestion
     */
    Item* (*roll_new)(enum RngSequence key);
    int (*get_buy_price)(Item* item);
    const char* (*get_name)(Item* item);
    const char* (*get_subtype_str)(Item* item);
    ItemSubtypeColors (*get_subtype_colors)(Item* item);
    bool (*can_acquire)(Item* item);
    void (*acquire)(Item* item);
    void (*dispose)(Item** item);
    int (*print_desc)(Item* item, Rect dest_rect);
    // TODO: void (*print_description)(Item* item); // or something of the form

    // Functions whose implementation is optional - not all items have to implement.
    int (*get_sell_price)(Item* item);
```

include/item.h
```diff
@@ -172,4 +240,14 @@ void item_dispose(Item** item);
...
 void item_print_buy_price_under(Item* item);
 
-#endif // ITEM_H
+/**
+ * @brief Prints the item's description inside the given rectangle
+ *
+ * @param item The item to print the description of
+ * @param dest_rect the target rectangle the description needs to fit in
+ *
+ * @return the number of lines used by the description
```
MeirGavish:
```suggestion
 *
 * Matches @ref ItemFuncs.print_description()
 *
 * @param item The item to print the description of
 * @param dest_rect the target rectangle the description needs to fit in
 *
 * @return the number of lines used by the description
```

Just for consistency with the rest of the functions' documentation

source/game/shop.c
```diff
@@ -586,182 +592,187 @@ static void shop_process_user_input(void)
...
     // Anim end
-    else if (s_timer == TM_SHOW_CARD_DESC_WAIT + 1)
+    else if (s_timer == TM_SHOW_ITEM_DESC_WAIT + 1)
     {
-        // Compute needed space for the description
-        const JokerInfo* info = get_joker_registry_entry(s_description_card->joker->id);
-        int desc_bottom_offset =
-            CARD_DESC_MAX_TEXT_HEIGHT -
-            info->joker_print_desc(s_description_card->joker, CARD_DESC_TEXT_RECT);
+        static const char undef_str[] = "UNKNOWN";
```
MeirGavish:
```suggestion
        static const char undef_str[] = "UNDEFINED";
```

"Unkown" implies that there is a value and we just don't know it but in truth this is an error value for which I think "UNDEFINED" or even "NULL" is better.

include/item.h
```diff
@@ -98,9 +115,14 @@ typedef struct ItemFuncs
      */
     Item* (*roll_new)(enum RngSequence key);
     int (*get_buy_price)(Item* item);
+    int (*get_sell_price)(Item* item);
+    const char* (*get_name)(Item* item);
+    const char* (*get_subtype_str)(Item* item);
+    ItemSubtypeColors (*get_subtype_colors)(Item* item);
```
MeirGavish: I've been thinking about this. In general, the number of functions that end up being added is bigger than I expected but I think it's okay.
For this though I think it would make more sense to bundle the `get_subtype_str` and `get_subtype_colors` functions into a single function (`get_description_subtype`?) that would return a bigger struct.
The reasoning is that the subtype string and colors must be implemented together, it's weird if one of them is implemented and the other isn't.
Also the word "description" has to be in the name because these are specifically for the description, there isn't anywhere else where you use the type for all item types except jokers where you use rarity.

include/game/shop.h
```diff
@@ -17,11 +17,11 @@
 void shop_reset(void);
 
 /**
- * @brief Get a pointer to the Card we are currently showing the description of.
+ * @brief Get a pointer to the Item we are currently showing the description of.
  *
- * @return JokerObject*
+ * @return Item*
```
MeirGavish:
```suggestion
 * @return Item* - NULL if no description is currently being shown
```

Since we're already touching the documentation here, it will be good to state the return value can be NULL.

source/item_funcs.c
```diff
@@ -53,6 +65,15 @@ static Item* item_roll_new_unimplemented(enum RngSequence key)
     return NULL;
 }
 
+static int item_get_sell_price_unimplemented(Item* item)
+{
+    MGBA_FUNC_ERROR(
+        "Unimplemented get_sell_price function called for item type type %d",
+        (item)->type
+    );
```
MeirGavish:
```suggestion
    GBAL_RETURN_IF_NULL_RET(item, UNDEFINED);
    MGBA_FUNC_ERROR(
        "Unimplemented get_sell_price function called for item type type %d",
        (item)->type
    );
```

Need to check for NULL here since the item is being dereferenced for the error message.

source/item_funcs.c
```diff
@@ -62,4 +83,13 @@ static void item_acquire_unimplemented(Item* item)
...
 {
     return true;
+}
+
+static int item_print_desc_unimplemented(Item* item, Rect dest_rect)
+{
+    MGBA_FUNC_ERROR(
+        "Unimplemented print_description function called for item type type %d",
+        (item)->type
+    );
```
MeirGavish:
```suggestion
static int item_print_desc_unimplemented(Item* item, Rect dest_rect)
{
    GBAL_RETURN_IF_NULL_RET(item, 0);
    MGBA_FUNC_ERROR(
        "Unimplemented print_description function called for item type type %d",
        (item)->type
    );
```

Here as well.

source/game/shop.c
```diff
@@ -586,182 +592,187 @@ static void shop_process_user_input(void)
...
+
+        // Print the Item's name
+        const char* item_name = item_get_name(s_description_item);
+        if (item_name == NULL)
+        {
+            MGBA_FUNC_WARN(
+                "Could not retrieve name of Item of type %d",
+                s_description_item->type,
```
MeirGavish:
```suggestion
        if (item_name == NULL)
        {
            MGBA_FUNC_WARN(
                "Could not retrieve name of Item of type %d",
                s_description_item->type
```

Missed a comma there, this still doesn't compile with loggin enabled.

source/item_funcs.c
```diff
@@ -29,9 +36,14 @@ ItemFuncs item_func_table[] = {
     [ITEM_TYPE_PLAYING_CARD] = {
         .roll_new = item_roll_new_unimplemented,
         .get_buy_price = card_object_get_buy_price,
+        .get_sell_price = item_get_sell_price_unimplemented,
+        .get_name = NULL,
+        .get_subtype_str = NULL,
+        .get_subtype_colors = NULL,
```
MeirGavish: It's kind of inconsistent to have dedicated unimplemented functions for some items, and `NULL` for others... is there a specific reason these are `NULL` and the others aren't?
I think my original reasoning for making unimplemented function versions was to be on the safe side (and also to be more explicit that some unimplemented functions are intentional but it doesn't seem to really serve this purpose) but since the wrapper functions do check for NULL, I think it's okay to remove the unimplemented function versions and assign `NULL` instead, they do feel a little tedious.
The important thing is to use the custom error messages from the unimplemented functions in the wrappers because it would be a shame to lose them and make sure the return value is correct.

include/joker.h
```diff
@@ -186,13 +200,12 @@ const char* joker_get_rarity_string(u8 rarity);
  *  - 11,12 -> Spectral Card
  *  - 13,14 -> Voucher
  *
```
MeirGavish: Now it's better to move the color listing documentation to either the color function or the color struct.

include/item.h
```diff
@@ -131,6 +153,52 @@ Item* item_roll_new(enum ItemType item_type, enum RngSequence key);
...
+/**
+ * @brief Returns the name of the Item's subtype (rarity for Jokers, just the type otherwise)
+ *
+ * Matches @ref ItemFuncs.get_subtype_str()
+ *
+ * @param item The item whose subtype's name to return.
+ *
+ * @return const char*
+ */
+const char* item_get_subtype_string(Item* item);
```
MeirGavish:
```suggestion
/**
 * @brief Returns the name of the Item
 *
 * Matches @ref ItemFuncs.get_name()
 *
 * @param item The item whose name to return.
 *
 * @return const char* - can be NULL if an error occurred
 */
const char* item_get_name(Item* item);

/**
 * @brief Returns the name of the Item's subtype (rarity for Jokers, just the type otherwise)
 *
 * Matches @ref ItemFuncs.get_subtype_str()
 *
 * @param item The item whose subtype's name to return.
 *
 * @return const char* - can be NULL if an error occurred
 */
const char* item_get_subtype_string(Item* item);
```

The documentation should include that it can return NULL.
Though, come to think of it, it would be better if these functions are the ones to return the constant "UNDEFINED" string in case of an error and never return NULL, instead of having the caller handle it.

include/item.h
```diff
@@ -131,6 +153,52 @@ Item* item_roll_new(enum ItemType item_type, enum RngSequence key);
...
+
+/**
+ * @brief Returns the colors of the Item's subtype
+ *
+ * Matches @ref ItemFuncs.get_subtype_colors()
+ *
+ * @param item The item whose subtype's color to return.
+ *
+ * @return struct containing values of both main and shadow colors
```
MeirGavish:
```suggestion
 * @return struct containing values of both main and shadow colors, all colors are 0 in case of an error
```

Again, error case needs to be documented

source/joker.c
```diff
@@ -282,6 +302,16 @@ void joker_object_dispose(Item** joker_object_item)
     *joker_object_item = NULL;
 }
 
+int joker_object_print_description(Item* joker_object_item, Rect dest_rect)
+{
+    GBAL_RETURN_IF_NULL_RET(joker_object_item, 0);
+    Joker* joker = ((JokerObject*)joker_object_item)->joker;
```
```suggestion
    GBAL_RETURN_IF_NULL_RET(joker_object_item, 0);
    ITEM_RETURN_IF_UNEXPECTED_TYPE_RET(joker_object_item, ITEM_TYPE_JOKER, UNDEFINED)
    Joker* joker = ((JokerObject*)joker_object_item)->joker;
```

Check the item type before casting

source/game/shop.c
```diff
@@ -176,15 +182,15 @@ static int s_reroll_cost = REROLL_BASE_COST;
 // Variables relative to the Card we are showing the description of
 
 // TODO: Change this to item once it has description printing API.
```
MeirGavish:
```suggestion
```

Time to remove this TODO

source/joker.c
```diff
@@ -165,22 +165,49 @@ u32 joker_get_score_effect(
...
     return JOKER_RARITY_STRINGS_LUT[rarity];
 }
 
-u16 joker_get_rarity_color(u8 rarity, bool main_color)
+ItemSubtypeColors joker_object_get_rarity_colors(Item* joker_object)
 {
+    ItemSubtypeColors colors = {0};
+
+    GBAL_RETURN_IF_NULL_RET(joker_object, colors);
+    Joker* joker = ((JokerObject*)joker_object)->joker;
```
MeirGavish: Check the type before casting the `Item*` to `JokerObject*`

MeirGavish (changes requested):

MeirGavish:
Woops, posted before writing a general review comment.
I gave this a deeper looks, looks good in general, I have some comments.
I haven't looked at everything but I have looked at most things, I think... I'll give it another pass later but I think the main things are in this review already.

PR #614, opened 2026-08-12
[Refactor] Replaced rogue call to obj_hide with sprite_object_hide

MeirGavish:
I noticed this call to `obj_hide()` that I seem to have missed in #563, so I properly replaced it with the correct function call.


MathisMartin31 (approved):
There doesn't seem to be any other calls to that function on SpriteObjects, nice ^^
Approving this one :3

MeirGavish:
> There doesn't seem to be any other calls to that function on SpriteObjects, nice ^^ Approving this one :3

Thanks! This is one very small change, I'll merge it.

